PR #32, opened 2026-06-04
feat(organizations): add guarded tenant-provisioning service

hugobessa:
## Summary
Phase 1 of the **Auth–Tenant Provisioning** plan ([ai-plans/2026-06-04-AUTH_TENANT_PROVISIONING_IMPLEMENTATION_PLAN.md](../ai-plans/2026-06-04-AUTH_TENANT_PROVISIONING_IMPLEMENTATION_PLAN.md)).

Adds the single guarded entry point that turns a membership-less user into an organization member, so every signup path (email-confirmation hook, social adapter, explicit invite-accept) routes the same decision through one place and can never silently create a second membership. No user-visible behavior on its own — this is the scaffolding Phases 3–7 consume.

## Changes
- New `UserAlreadyHasMembershipError` (DRF `ValidationError` → renders as 400) in `organizations/exceptions.py`.
- New `OrganizationService.provision_tenant_for_user(user, organization_name=None)` with ordered branches: already-member → raise; non-expired/unaccepted invitation matching `user.email` → MEMBER join + mark invitation accepted; else `organization_name` → `create_organization` (ADMIN); else `None` (gated onboarding).
- Hardened `accept_invitation`: raises the typed error before the DB write when the user already has a membership, and wraps the membership insert in a savepoint so a concurrent unique-violation rolls back cleanly under `ATOMIC_REQUESTS` instead of poisoning the request transaction.
- Switched the `IntegrityError` import to Django's (`from django.db import IntegrityError`) so the concurrency backstop catches what the ORM actually raises.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 1 — "Central tenant-provisioning service".
- Spec use-case: shared scaffolding (no use-case yet); underpins Use-cases 1, 2, 4, 5.

## Test plan
- `uv run pytest -n auto` (full suite: 1170 passed)

Covers all four provisioning branches, the expired-invite fall-through, the hardened `accept_invitation`, and three backstop tests that mock `OrganizationMembership.objects.create` to raise `django.db.IntegrityError` and assert the typed error surfaces.

### DIFF
--- a/organizations/exceptions.py
+++ b/organizations/exceptions.py
@@ -14,3 +14,8 @@ class InvalidInvitationTokenError(ValidationError):
 class InvitationNotFoundError(ValidationError):
     default_detail = "Invitation does not exist"
     default_code = "invitation_not_found"
+
+
+class UserAlreadyHasMembershipError(ValidationError):
+    default_detail = "User already belongs to an organization."
+    default_code = "user_already_has_membership"
--- a/organizations/services.py
+++ b/organizations/services.py
@@ -1,12 +1,11 @@
 import datetime
 from typing import Annotated
 
-from django.db import transaction
+from django.db import IntegrityError, transaction
 from django.urls import reverse
 
 from allauth.utils import build_absolute_uri
 from dependency_injector.wiring import Provide, inject
-from psycopg import IntegrityError
 from vintasend.services.notification_service import (
     NotificationContextDict,
     NotificationService,
@@ -23,6 +22,7 @@ from organizations.exceptions import (
     DuplicateInvitationError,
     InvalidInvitationTokenError,
     InvitationNotFoundError,
+    UserAlreadyHasMembershipError,
 )
 from organizations.models import (
     Organization,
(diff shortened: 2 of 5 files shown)


organizations/services.py
```diff
@@ -1,12 +1,11 @@
 import datetime
 from typing import Annotated
 
-from django.db import transaction
+from django.db import IntegrityError, transaction
```
hugobessa, line 4: Load-bearing import: this must be Django's `IntegrityError`, not psycopg's. The ORM re-wraps the driver's exception as `django.db.utils.IntegrityError`, so the `except IntegrityError` concurrency backstops below (and in `invite_user_to_organization`) only catch a unique-constraint violation when this points at the Django class. With psycopg's class the backstop silently caught nothing — that was a review blocker, fixed here.

organizations/services.py
```diff
@@ -147,24 +147,99 @@ def invite_user_to_organization(
...
+
+        Raises UserAlreadyHasMembershipError if the user is already a member of any
+        organization before the membership create is attempted, preventing an IntegrityError
+        from the OneToOne constraint on OrganizationMembership.user.
+
         :param token: Invitation token.
         :param user: User who is accepting the invitation.
         :return: Created OrganizationMembership instance.
         """
+        if hasattr(user, "organization_membership"):
```
hugobessa, line 159: `hasattr(user, "organization_membership")` returns False when the reverse OneToOne has no related row (the descriptor raises `RelatedObjectDoesNotExist`, a subclass of `AttributeError`). This is the fast single-threaded guard; the savepoint + `except IntegrityError` just below is the true-race backstop for two requests that both pass this check.

organizations/services.py
```diff
@@ -147,24 +147,99 @@ def invite_user_to_organization(
...
+
         now = datetime.datetime.now(tz=datetime.UTC)
         invitations = OrganizationInvitation.objects.filter(email=user.email, expires_at__gt=now)
         for invitation in invitations:
             if verify_long_lived_token(token, invitation.token_hash):
-                membership = OrganizationMembership.objects.create(
-                    user=user, organization=invitation.organization
-                )
+                try:
+                    with transaction.atomic():
```
hugobessa, line 167: Nested `with transaction.atomic()` around the insert is deliberate. Under `ATOMIC_REQUESTS=True` (production), an `IntegrityError` aborts the whole request transaction; re-raising a DRF `ValidationError` returns 400 without signalling the tx middleware to roll back, so the request would try to commit a poisoned transaction. The savepoint contains the failure and restores a usable connection.

organizations/services.py
```diff
@@ -147,24 +147,99 @@ def invite_user_to_organization(
...
+
+        now = datetime.datetime.now(tz=datetime.UTC)
+        pending_invitation = OrganizationInvitation.objects.filter(
+            email=user.email,
+            expires_at__gt=now,
+            accepted_at__isnull=True,
+            membership__isnull=True,
+        ).first()
+
+        if pending_invitation is not None:
```
hugobessa, line 220: Invite-first ordering is a Guiding Decision: a pending invitation always wins over creating a new org, so an invited user joins the inviting tenant (MEMBER) rather than spawning a stray org. Detection is by email and `OrganizationInvitation.email` is globally unique, so there is at most one match — no tie-break needed. Phases 3/4/6 call this method; they never re-implement the branch logic.

organizations/services.py
```diff
@@ -147,24 +147,99 @@ def invite_user_to_organization(
...
+                    role=OrganizationRole.MEMBER,
+                )
+            except IntegrityError as e:
+                raise UserAlreadyHasMembershipError() from e
+            pending_invitation.accepted_at = now
+            pending_invitation.membership = membership
+            pending_invitation.save()
+            return membership
+
+        if organization_name:
```
hugobessa, line 234: The `organization_name` branch delegates to `create_organization` (which makes the creator ADMIN) and is wrapped in its own IntegrityError guard so a concurrent race on this path also surfaces `UserAlreadyHasMembershipError` rather than a raw DB error. `create_organization` itself is left unchanged so its other callers keep their current behavior.

organizations/exceptions.py
```diff
@@ -14,3 +14,8 @@ class InvalidInvitationTokenError(ValidationError):
 class InvitationNotFoundError(ValidationError):
     default_detail = "Invitation does not exist"
     default_code = "invitation_not_found"
+
+
+class UserAlreadyHasMembershipError(ValidationError):
```
hugobessa, line 19: Extends DRF `ValidationError` so it renders as a 400 automatically. Phase 7 surfaces this at the accept-invitation endpoint; defining it here (raised by both `accept_invitation` and `provision_tenant_for_user`) keeps the one-org-per-user rejection typed across every path.

organizations/tests/test_services.py
```diff
@@ -492,7 +504,251 @@ def test_accept_invitation_already_accepted(self, organization_service, organiza
...
+
+    # -----------------------------------------------------------------------
+    # Backstop tests: IntegrityError → UserAlreadyHasMembershipError
+    # These tests genuinely exercise the except IntegrityError path by mocking
+    # OrganizationMembership.objects.create to raise django.db.IntegrityError.
+    # They would FAIL if the import were psycopg.IntegrityError instead of the
+    # Django one, proving that the correct class is caught.
+    # -----------------------------------------------------------------------
+
+    def test_provision_tenant_for_user_invite_branch_integrity_error_backstop(
```
hugobessa, line 681: These backstop tests mock `OrganizationMembership.objects.create` to raise `django.db.IntegrityError`, which is the only way to exercise the `except` path — the sequential `_already_member_guard` test hits the `hasattr` check before any write and never reaches it. Added after review found the original concurrency test proved nothing about the backstop.

PR #33, opened 2026-06-04
feat(accounts): capture intended organization name at signup

hugobessa:
## Summary
Phase 2 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 1 (PR #32).

Email/password signup now persists the intended organization name so the org can be created later, at email-confirmation time (Phase 3). Because the org is deferred to verification, the name has to be stashed at signup-form time and consumed on confirmation. Invited signups leave it blank — they auto-join the inviting org rather than create one.

## Changes
- `Profile.pending_organization_name` — new `CharField(max_length=255, blank=True, default="")` + additive, reversible migration (`users/0002`). Empty-string blank state, consistent with `first_name`/`last_name`; metadata-only `ADD COLUMN` on Postgres, no lock risk.
- `BaseVintaScheduleSignupForm` — optional `organization_name` field; `signup()` stashes it on the profile **only when no pending invitation matches the signup email**, else leaves it blank.
- **Case-insensitive invite matching, consistently across all three sites** (`_has_pending_invitation`, `provision_tenant_for_user`, `accept_invitation`) — `email__iexact`. See the inline comment for why this had to span Phase 1 code.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 2 — "Capture intended org name at signup".
- Spec use-case: shared scaffolding for the email/password use-cases (UC1/UC2).

## Test plan
- `uv run python manage.py makemigrations --check`
- `uv run pytest -n auto` (full suite: 1180 passed)

Covers uninvited→captured, invited(pending)→blank, expired-invite→captured, case-differing-invite→detected, and signup backward-compat (names preserved, create-or-update both work).


users/models.py
```diff
@@ -68,6 +68,16 @@ class Profile(BaseModel):
     first_name = models.CharField(max_length=255, blank=True)
     last_name = models.CharField(max_length=255, blank=True)
     profile_picture = S3DirectImageField(dest="profile_pictures", blank=True, null=True)
+    pending_organization_name = models.CharField(
```
hugobessa, line 71: Stored as empty-string blank (not null) to match `first_name`/`last_name`. This holds the intended org name only between signup and email confirmation; Phase 3's `email_confirmed` hook consumes it (passes it to `provision_tenant_for_user`) and clears it. It is deliberately NOT acted on in this phase — no org is created at signup-form time, because the Guiding Decision defers org creation to after email verification.

accounts/base_forms.py
```diff
@@ -1,36 +1,71 @@
...
         """
         user.save()
 
         first_name = self.cleaned_data.get("first_name", "")
         last_name = self.cleaned_data.get("last_name", "")
+        organization_name = self.cleaned_data.get("organization_name", "")
+
+        # If a pending invitation exists for this email, leave the org name blank
+        # so Phase 3's provisioning hook falls through to the invite-auto-join path.
+        if self._has_pending_invitation(user.email):
```
hugobessa, line 55: Invited users must NOT capture an org name: if a pending invitation matches the signup email, leave `pending_organization_name` blank so Phase 3 falls through to the invite-auto-join branch rather than creating a stray org. The pending-invite predicate here (email + not-expired + unaccepted + unlinked) is exactly the one `provision_tenant_for_user` uses — they must stay identical or capture and provisioning could disagree.

accounts/base_forms.py
```diff
@@ -1,36 +1,71 @@
...
+    organization_name = forms.CharField(
+        max_length=255,
+        required=False,
+        label="Organization Name",
+    )
+
+    def _has_pending_invitation(self, email: str) -> bool:
+        """Return True if a non-expired, unaccepted invitation exists for *email*."""
+        return OrganizationInvitation.objects.filter(
+            email__iexact=email,
```
hugobessa, line 32: `email__iexact`, not plain equality. Invitation emails are stored as-supplied (un-normalized) while `User.email` normalizes the domain, so a case-differing local part would otherwise miss the match. Critically, this phase also switched `provision_tenant_for_user` and `accept_invitation` (Phase 1 code) to `iexact` in the same change — fixing only the form would introduce an asymmetry where signup skips org capture (sees the invite) but confirmation fails to auto-join (misses it), orphaning the user. All three match sites are now consistent.

PR #34, opened 2026-06-04
feat(accounts): provision tenant on email confirmation

hugobessa:
## Summary
Phase 3 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 2 (PR #33). Implements **Use-case 1** (self-service email/password signup, no invite).

On email confirmation, a handler provisions the user's tenant: it resolves `OrganizationService` from the DI container and calls `provision_tenant_for_user(user, organization_name=profile.pending_organization_name)`. For an uninvited user this creates a new organization with the user as ADMIN and clears the stashed name. The handler is idempotent and also transparently covers the invited path (Phase 1's service is invite-first), though invited-path tests land in Phase 4.

## Open Question 1 — resolved
The plan flagged uncertainty over whether allauth's `email_confirmed` signal fires for **code-based headless** verification. Verified against allauth 65.18.0 source: the by-code headless flow funnels through `verify_email()`, which always emits `email_confirmed` — so the signal handler is the correct, reliably-firing hook (no adapter override needed). Chain: headless verify endpoint → `EmailVerificationProcess.finish()` → `mark_email_address_as_verified()` → `adapter.confirm_email()` → `verify_email()` → `email_confirmed.send(...)`.

## Changes
- `accounts/signals.py` — `email_confirmed` handler: DI-resolve `OrganizationService`, provision, clear `pending_organization_name` on success, swallow `UserAlreadyHasMembershipError` for idempotency, guard missing user/profile.
- `accounts/apps.py` — `AccountsConfig.ready()` connects the handler (with `dispatch_uid`).

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 3 — "Create own org on email verification (no invite)".
- Spec use-case: Use-case 1.

## Test plan
- `uv run pytest -n auto` (full suite: 1185 passed)

Tests drive the **real** allauth `verify_email` flow (not a manual signal send): uninvited→org+ADMIN+name cleared; re-confirmation→no-op; blank name + no invite→no org (gated); invited→MEMBER; missing-profile guard.


accounts/signals.py
```diff
@@ -0,0 +1,115 @@
...
+from allauth.account.signals import email_confirmed
+
+from organizations.exceptions import UserAlreadyHasMembershipError
+from users.models import Profile
+
+
+logger = logging.getLogger(__name__)
+
+
+def _provision_on_email_confirmed(
```
hugobessa: Hook choice (Open Question 1 in the plan): `email_confirmed` is the right signal because the code-based headless verify flow routes through allauth's `verify_email()`, which always emits it — confirmed in the 65.18.0 source. An adapter override was considered and rejected as more coupled and no more reliable. Org creation is deliberately here (confirmation), not at signup, per the Guiding Decision to defer org creation until the email is verified.

accounts/signals.py
```diff
@@ -0,0 +1,115 @@
...
+        # No pending invite, no org name → user stays gated (onboarding later).
+        logger.debug(
+            "No pending invite and no org name for user %s; user remains membership-less (gated).",
+            user.pk,
+        )
+
+
+def connect() -> None:
+    """Connect all signal handlers for the accounts app."""
+    email_confirmed.connect(
```
hugobessa: `dispatch_uid` makes the connect idempotent — some test configs call `ready()` more than once, and without it the handler would fire multiple times per confirmation. Combined with the handler swallowing `UserAlreadyHasMembershipError`, a re-fired or re-confirmed event creates nothing extra (a real second `verify_email` re-emits the signal — allauth does NOT early-return for an already-verified address).

accounts/signals.py
```diff
@@ -0,0 +1,115 @@
...
+            user.pk,
+        )
+        return
+
+    if membership is not None:
+        # Clear the stashed org name now that provisioning succeeded.
+        # Only write if there is actually something to clear (avoids a
+        # redundant "" -> "" save on the invite path where Phase 2 already
+        # left the field blank).
+        if profile.pending_organization_name:
```
hugobessa: Only clear `pending_organization_name` when it actually holds something. On the invite path Phase 2 already left it blank, so this avoids a redundant `"" -> ""` write. The clear + the org/membership insert both sit inside the request transaction (the service is `@transaction.atomic()`), so a failure rolls back together and the idempotency guard absorbs the retry.

accounts/account_adapters.py
```diff
@@ -217,6 +219,83 @@ def send_confirmation_mail(self, request, emailconfirmation, signup):
             context_kwargs=NotificationContextDict(ctx),
         )
 
+    def confirm_email(self, request, email_address: EmailAddress) -> bool:
```
hugobessa, line 225: Amended hook (was an `email_confirmed` signal handler). Overriding `AccountAdapter.confirm_email` puts provisioning on the real production confirmation path — verified in allauth 65.18.0: `mark_email_address_as_verified → get_adapter().confirm_email`. The override sits ABOVE `verify_email` (which still emits the now-unused signal), so it runs for both code-based and link-based headless verification. Chosen over the signal for a discoverable stack trace and step-through debugging at a single explicit call site.

accounts/account_adapters.py
```diff
@@ -217,6 +219,83 @@ def send_confirmation_mail(self, request, emailconfirmation, signup):
...
+        the tenant imperatively in the adapter — at the moment the email is verified —
+        rather than via a decoupled email_confirmed signal handler, so the email/password
+        path's provisioning lives at an explicit, step-debuggable call site (mirroring the
+        intent of provisioning inside the adapters rather than through signals).
+
+        Idempotent: swallows UserAlreadyHasMembershipError so re-confirmation is a no-op.
+        """
+        from di_core.containers import container as di_container
+
+        confirmed = super().confirm_email(request, email_address)
```
hugobessa, line 242: Returns the `confirmed` bool from `super().confirm_email` on EVERY branch (early-return, missing-profile guard, DI-None guard, already-member swallow, no-op) so allauth's contract is preserved. Provisioning only runs when `confirmed` is truthy. Passing `pending_organization_name or None` keeps the email adapter's call shape parallel to the social adapter's (which passes no name).

accounts/account_adapters.py
```diff
@@ -217,6 +219,83 @@ def send_confirmation_mail(self, request, emailconfirmation, signup):
...
+            return confirmed
+
+        organization_service = di_container.organization_service()
+
+        try:
+            membership = organization_service.provision_tenant_for_user(
+                user=user,
+                organization_name=organization_name,
+            )
+        except UserAlreadyHasMembershipError:
```
hugobessa, line 264: Idempotency: a real re-confirmation re-enters `confirm_email`; the already-member case is swallowed so no second org/membership is created. `provision_tenant_for_user` is `@transaction.atomic` — under `ATOMIC_REQUESTS` it nests as a savepoint, so a race-induced rollback doesn't poison the request transaction.

accounts/tests/test_email_confirmation_provisioning.py
```diff
@@ -0,0 +1,190 @@
...
+    """Create (and persist) an allauth EmailAddress for *user*."""
+    return EmailAddress.objects.create(
+        user=user,
+        email=user.email,
+        verified=verified,
+        primary=True,
+    )
+
+
+def _confirm_email(rf, email_address: EmailAddress) -> bool:
```
hugobessa, line 37: Helper now drives `get_adapter(request).confirm_email(...)` instead of `verify_email(...)`. This is load-bearing: `verify_email` sits BELOW `confirm_email`, so calling it directly would bypass the override and the test would pass for the wrong reason (or fail to exercise provisioning). `get_adapter` resolves the configured `AccountAdapter`, identical to production.

PR #35, opened 2026-06-04
test(accounts): cover invited-email auto-join on confirmation

hugobessa:
## Summary
Phase 4 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 3 (PR #34). Implements **Use-case 2** (email/password signup by an invited address).

**Tests-only — no production code changed.** Phase 1's invite-first `provision_tenant_for_user`, Phase 2's signup capture-skip, and Phase 3's `email_confirmed` handler already implement invited-email auto-join. This phase proves Use-case 2 end-to-end and locks it with integration tests, per the plan's one-phase-per-use-case rule.

## Changes
- New `accounts/tests/test_email_invite_autojoin.py` (4 integration tests) driving the **real** signup form (`BaseVintaScheduleSignupForm.signup()`) and the **real** allauth `verify_email` path:
  1. Full end-to-end: invited email → signup (capture-skip leaves name blank) → confirm → MEMBER in inviting org, invitation accepted+linked, zero new orgs.
  2. Invite wins over a non-blank `pending_organization_name` → no stray org.
  3. Case-insensitive invite (case-differing local part) → still auto-joins via `email__iexact`.
  4. Invitation marked accepted (`accepted_at` set, `membership` linked) after auto-join.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 4 — "Auto-join invited org on email verification".
- Spec use-case: Use-case 2.

## Test plan
- `uv run pytest -n auto` (full suite: 1189 passed)

### DIFF
--- a/accounts/tests/test_email_invite_autojoin.py
+++ b/accounts/tests/test_email_invite_autojoin.py
@@ -1,0 +1,260 @@
+"""
+Phase 4 — Integration tests: Auto-join invited org on email verification.
+
+These tests prove Use-case 2 end-to-end: a user who signed up with an invited email
+address automatically joins the inviting organisation as MEMBER the moment their email
+is confirmed, with no separate accept step and no stray organisations created.
+
+Four scenarios are covered:
+
+1. Full end-to-end flow — pending invitation + real signup form + real verify_email →
+   MEMBER membership, invitation accepted and linked, zero new organisations created.
+2. Invite wins over org name — even if pending_organization_name is non-blank at
+   confirmation time, the invite-first branch wins and no stray org is created.
+3. Case-insensitive invite — invitation stored with a case-differing local part still
+   matches at confirmation (guards the email__iexact consistency).
+4. Invitation marked accepted — accepted_at and membership FK are both set after
+   auto-join.
+
+All tests drive the REAL allauth verify_email path so the email_confirmed signal fires
+exactly as it does at runtime (same pattern as test_email_confirmation_provisioning.py).
+"""
+
+import datetime
+
+import pytest
+from allauth.account.internal.flows.email_verification import verify_email
+from allauth.account.models import EmailAddress
+from model_bakery import baker
+
+from organizations.models import Organization, OrganizationInvitation, OrganizationMembership
+from users.factories import UserFactory
+
+
+# ---------------------------------------------------------------------------
+# Helpers (mirror the pattern from test_email_confirmation_provisioning.py)
+# ---------------------------------------------------------------------------
+
+
+def _create_email_address(user, verified: bool = False) -> EmailAddress:
+    """Create (and persist) an allauth EmailAddress for *user*."""
+    return EmailAddress.objects.create(
+        user=user,
+        email=user.email,
+        verified=verified,
+        primary=True,
+    )
+
+
+def _confirm_email(rf, email_address: EmailAddress) -> bool:
+    """Run the real allauth verify_email flow for *email_address*.
+
+    Uses a minimal GET request so add_message() has a request object and the
+    message storage backend doesn't raise.  CookieStorage is used because the
+    RequestFactory doesn't set up session middleware.  The signal fires inside
+    this call, exercising the same code path as the headless verify-email endpoint.
+    """
+    from django.contrib.messages.storage.cookie import CookieStorage
+
+    request = rf.get("/")
+    request._messages = CookieStorage(request)
+    return verify_email(request, email_address)
+
+
+def _pending_invitation(
+    org: Organization,
+    inviter,
+    email: str,
+) -> OrganizationInvitation:
+    """Create a non-expired, unaccepted OrganizationInvitation."""
+    return baker.make(
+        OrganizationInvitation,
+        email=email,
+        organization=org,
+        invited_by=inviter,
+        expires_at=datetime.datetime.now(tz=datetime.UTC) + datetime.timedelta(days=7),
+        accepted_at=None,
+        membership=None,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Tests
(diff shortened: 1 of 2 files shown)


accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,260 @@
...
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+class TestInvitedEmailAutoJoin:
+    """Integration: invited-email auto-join on confirmation (Use-case 2)."""
+
+    def test_full_end_to_end_invited_signup_autojoin(self, rf):
```
hugobessa, line 93: This is the true Use-case 2 end-to-end test: it drives the REAL `BaseVintaScheduleSignupForm.signup()` (so Phase 2's capture-skip runs and asserts `pending_organization_name` ends blank for the invited email) and then the REAL allauth `verify_email` (so Phase 3's signal handler fires). Nothing is mocked — it proves the three phases compose into the no-separate-accept-step auto-join the spec requires. Phase 4 added no production code; the invite-first service already did the work.

accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,260 @@
...
+        # Zero new organisations were created — only the pre-existing one exists.
+        assert Organization.objects.count() == 1
+
+        # Invitation is marked accepted and linked to the membership.
+        invitation.refresh_from_db()
+        assert invitation.accepted_at is not None, "Invitation accepted_at must be set"
+        assert invitation.membership is not None, "Invitation must be linked to the membership"
+        assert invitation.membership_id == membership.pk
+
+    def test_invite_wins_over_pending_organization_name(self, rf):
```
hugobessa, line 155: Guards the invite-first invariant directly: even with a non-blank `pending_organization_name` at confirmation, the user joins the inviting org as MEMBER and no stray org is created. The form normally prevents this state (Phase 2), but the test sets the name directly to lock the service's guarantee independently of the form — defense in depth against a future form regression.

accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,260 @@
...
+        # User is MEMBER of the inviting org, not ADMIN of a new one.
+        assert OrganizationMembership.objects.filter(user=user).count() == 1
+        membership = OrganizationMembership.objects.get(user=user)
+        assert membership.organization == org
+        assert membership.role == "member"
+
+        # No stray org was created.
+        assert Organization.objects.count() == 1
+
+    def test_case_insensitive_invite_autojoin(self, rf):
```
hugobessa, line 191: Regression guard for the `email__iexact` consistency fix from Phase 2: invitation stored with a case-differing local part still matches at confirmation. If any of the three match sites drifted back to case-sensitive equality, this auto-join would silently fail and the user would be orphaned — exactly the failure mode the iexact change prevents.

accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,263 @@
...
+    message storage backend doesn't raise.  CookieStorage is used because the
+    RequestFactory doesn't set up session middleware.  Provisioning fires inside
+    this call via the adapter override, exercising the same hook as the headless
+    verify-email endpoint.
+    """
+    from django.contrib.messages.storage.cookie import CookieStorage
+
+    request = rf.get("/")
+    request._messages = CookieStorage(request)
+    return get_adapter(request).confirm_email(request, email_address)
```
hugobessa, line 64: Amended for the Phase 3 rework: this helper now drives `get_adapter(request).confirm_email(...)`, not `verify_email(...)`. After Phase 3 replaced the `email_confirmed` signal with an `AccountAdapter.confirm_email` override, calling `verify_email` directly bypasses provisioning — these four tests failed on the rebase until the helper was switched. Mirrors the helper in `test_email_confirmation_provisioning.py`.

accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,263 @@
...
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+class TestInvitedEmailAutoJoin:
+    """Integration: invited-email auto-join on confirmation (Use-case 2)."""
+
+    def test_full_end_to_end_invited_signup_autojoin(self, rf):
```
hugobessa, line 93: True Use-case 2 end-to-end: drives the REAL `BaseVintaScheduleSignupForm.signup()` (Phase 2 capture-skip leaves `pending_organization_name` blank for the invited email) then the REAL `AccountAdapter.confirm_email` (Phase 3 override) — proving the three phases compose into the no-separate-accept-step auto-join. Nothing mocked.

accounts/tests/test_email_invite_autojoin.py
```diff
@@ -0,0 +1,263 @@
...
+        # User is MEMBER of the inviting org, not ADMIN of a new one.
+        assert OrganizationMembership.objects.filter(user=user).count() == 1
+        membership = OrganizationMembership.objects.get(user=user)
+        assert membership.organization == org
+        assert membership.role == "member"
+
+        # No stray org was created.
+        assert Organization.objects.count() == 1
+
+    def test_case_insensitive_invite_autojoin(self, rf):
```
hugobessa, line 191: Regression guard for the `email__iexact` consistency fix (Phase 2): an invitation stored with a case-differing local part still matches at confirmation. If any match site drifted back to case-sensitive equality, this auto-join would silently fail and orphan the user.

PR #36, opened 2026-06-04
test(accounts): gated onboarding for uninvited social signup

hugobessa:
## Summary
Phase 5 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 4 (PR #35). Implements **Use-case 3** (social signup, no invite).

**Tests-only — no production code changed.** Social signup has no email-verification step, so the email-confirmation hook (Phase 3) never fires for it. An uninvited social user therefore lands authenticated but membership-less (gated) and creates their org through the existing `OrganizationViewSet` create endpoint, becoming ADMIN. The gate already exists: `OrganizationManagementPermission` lets membership-less users create and blocks members from creating a second org; `OrganizationInvitationPermission` requires a membership. `SocialAccountAdapter.save_user` only persists the user + ensures a Profile — it creates no org/membership. This phase locks that behavior with tests.

## Changes
- New `accounts/tests/test_social_gated_onboarding.py` (6 tests):
  1. `save_user` for an uninvited social user → no membership, no org.
  2. `user.organization_membership` raises when absent (gating invariant).
  3. Gated user POSTs org-create → 201, becomes ADMIN.
  4. Second create → 403, no second org/membership.
  5. Membership-less user → invitation list 403.
  6. Membership-less user → invitation create 403.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 5 — "Gated onboarding for uninvited social signup".
- Spec use-case: Use-case 3 (and Acceptance scenarios 2 and 6).

## Test plan
- `uv run pytest organizations/tests/ accounts/tests/ -n auto`
- `uv run pytest -n auto` (full suite: 1195 passed)

### DIFF
--- a/accounts/tests/test_social_gated_onboarding.py
+++ b/accounts/tests/test_social_gated_onboarding.py
@@ -1,0 +1,234 @@
+"""
+Phase 5 — Integration tests: Gated onboarding for uninvited social signup.
+
+Use-case 3: A social user who signed up without a pending invitation lands
+authenticated but membership-less (gated).  They then create their own
+organisation via the existing OrganizationViewSet create endpoint and become
+ADMIN.  A second create attempt is rejected by OrganizationManagementPermission.
+
+Four scenarios are covered:
+
+1. Uninvited social signup (save_user) leaves the user membership-less — no
+   Organisation is created, no OrganizationMembership row exists.
+2. Gated → create org → ADMIN: a membership-less authenticated user POSTs the
+   org-create endpoint, succeeds, and is recorded as ADMIN of the new org.
+3. Second create rejected: the now-member user POSTs org-create again and is
+   refused with 403; no second org or membership is created.
+4. Membership-less user blocked from a member-only tenant endpoint: a
+   membership-less user hitting OrganizationInvitationViewSet gets 403 (no
+   membership → OrganizationInvitationPermission denies access).
+
+The social save_user path is simulated by:
+  - Calling SocialAccountAdapter.save_user() directly (with the real allauth
+    super().save_user() replaced by a minimal stub that saves the user, matching
+    what the existing test_account_adapters.py tests do).
+  - This replicates the production codepath without spinning up a full OAuth
+    round-trip, and is consistent with the existing adapter test pattern.
+"""
+
+from unittest.mock import MagicMock, patch
+
+from django.urls import reverse
+
+import pytest
+from allauth.socialaccount.models import SocialLogin
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from accounts.account_adapters import SocialAccountAdapter
+from organizations.models import Organization, OrganizationMembership, OrganizationRole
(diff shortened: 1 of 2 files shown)


accounts/tests/test_social_gated_onboarding.py
```diff
@@ -0,0 +1,234 @@
...
+
+@pytest.mark.django_db
+class TestSocialGatedOnboarding:
+    """Phase 5 / Use-case 3: uninvited social signup → gated → create org → ADMIN."""
+
+    # ------------------------------------------------------------------
+    # Scenario 1 — save_user leaves the user membership-less
+    # ------------------------------------------------------------------
+
+    def test_uninvited_social_signup_leaves_user_membership_less(self):
```
hugobessa, line 93: The key invariant for the social path: `save_user` must leave an uninvited user membership-less. Social signup skips email verification (`SOCIALACCOUNT_EMAIL_VERIFICATION=None`), so the Phase 3 `email_confirmed` hook never fires for it — there is no automatic provisioning moment. Auto-join for INVITED social users is Phase 6; this phase deliberately covers only the uninvited→gated path. The adapter's parent `save_user` is stubbed (same pattern as `test_account_adapters.py`) so the real adapter code runs without an OAuth round-trip.

accounts/tests/test_social_gated_onboarding.py
```diff
@@ -0,0 +1,234 @@
...
+        user = _social_save_user("gated@social.example.com")
+
+        with pytest.raises(ObjectDoesNotExist):
+            _ = user.organization_membership
+
+    # ------------------------------------------------------------------
+    # Scenario 2 — gated user creates org and becomes ADMIN
+    # ------------------------------------------------------------------
+
+    def test_gated_social_user_can_create_org_and_becomes_admin(self):
```
hugobessa, line 132: No new gate was built for this phase — `OrganizationManagementPermission` already allows a membership-less user to create (and blocks a member from creating a second). This test locks that existing behavior as the social onboarding contract: gated user → create → ADMIN. Phase 8 audits the gate across all tenant-scoped endpoints more broadly.

accounts/tests/test_social_gated_onboarding.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # User is now ADMIN
+        membership = OrganizationMembership.objects.get(user=user)
+        assert membership.role == OrganizationRole.ADMIN
+
+    # ------------------------------------------------------------------
+    # Scenario 3 — second create attempt is rejected
+    # ------------------------------------------------------------------
+
+    def test_second_org_create_rejected_after_first_succeeds(self):
```
hugobessa, line 162: One-org-per-user enforced at the permission layer: the second create returns 403 (not a DB error), and the test asserts no second org/membership leaked. Combined with the service-level `UserAlreadyHasMembershipError` guard from Phase 1, the invariant holds at both the API and service layers.

PR #37, opened 2026-06-04
feat(accounts): auto-join invited org on social signup

hugobessa:
## Summary
Phase 6 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 5 (PR #36). Implements **Use-case 4** (social signup by an invited address).

Social signup has no email-verification step to hook on, so provisioning happens at `save_user` time. `SocialAccountAdapter.save_user` now calls `provision_tenant_for_user(user)` (no org name) after the user + Profile are persisted: a matching pending invitation → MEMBER join in the inviting org (skips the create-org prompt); no invite → returns `None`, user stays gated (preserves Phase 5). `UserAlreadyHasMembershipError` is swallowed for idempotency.

## Changes
- `accounts/account_adapters.py` — `SocialAccountAdapter._provision_org_membership(user)`, called at the end of `save_user`. DI-resolves `OrganizationService` (same container pattern as `accounts/signals.py`). Guards blank email and missing container.
- `organizations/services.py` — savepoint hygiene: wrapped both `create()` sites in `provision_tenant_for_user` in inner `with transaction.atomic():` (mirrors what Phase 1 already did for `accept_invitation`), so a race-induced IntegrityError rolls back cleanly instead of poisoning the request transaction.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 6 — "Auto-join invited org on social signup".
- Spec use-case: Use-case 4.

## Test plan
- `uv run pytest -n auto` (full suite: 1202 passed)

Tests drive the real `SocialAccountAdapter.save_user` (parent stubbed): invited→MEMBER+invitation accepted+linked, no new org; uninvited→stays gated (Phase 5 regression); case-insensitive invite; already-member guard genuinely fires on a second provisioning attempt; plus two transaction-not-poisoned regression tests for the savepoint fix.

### DIFF
--- a/accounts/account_adapters.py
+++ b/accounts/account_adapters.py
@@ -77,7 +77,67 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
         )
         user.profile = profile
         self._enqueue_profile_picture_download(profile, sociallogin)
+        # Phase 6 — Auto-join invited org on social signup.
+        # At this point the user is persisted (has a pk) and the Profile
+        # invariant is satisfied.  Attempt to provision a tenant: if a
+        # pending invitation matches the social email the user becomes a MEMBER
+        # of the inviting org immediately, skipping the create-org prompt.
+        # When there is no matching invitation, provision_tenant_for_user
+        # returns None and the user stays membership-less (gated), preserving
+        # the Phase 5 uninvited-stays-gated behavior.
+        # UserAlreadyHasMembershipError is treated as a no-op so that a social
+        # re-login for a user who is already a member does not raise.
+        self._provision_org_membership(user)
         return user
+
+    @staticmethod
+    def _provision_org_membership(user: User) -> None:
+        """Attempt to auto-join the user to an inviting organisation.
+
+        Resolves OrganizationService from the DI container (same pattern as
+        accounts/signals.py) and calls provision_tenant_for_user with no
+        organisation_name — social signups never carry one.  No-ops silently on
+        UserAlreadyHasMembershipError (re-login or race).
+        """
+        if not user.email:
+            logger.warning(
+                "Social user %s has no email; skipping org provisioning.",
+                user.pk,
+            )
+            return
+
+        from di_core.containers import container as di_container
+
+        if di_container is None:
+            logger.error(
+                "DI container is not initialised; cannot provision tenant for social user %s.",
+                user.pk,
+            )
+            return
+
+        organization_service = di_container.organization_service()
+        try:
+            membership = organization_service.provision_tenant_for_user(user=user)
+        except UserAlreadyHasMembershipError:
+            # Re-login or race: user already has a membership — no-op.
+            logger.debug(
+                "Social user %s already has a membership; skipping re-provisioning.",
+                user.pk,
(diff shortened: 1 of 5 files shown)


accounts/account_adapters.py
```diff
@@ -74,8 +75,68 @@ def save_user(self, request, sociallogin, form=None):
...
+        # At this point the user is persisted (has a pk) and the Profile
+        # invariant is satisfied.  Attempt to provision a tenant: if a
+        # pending invitation matches the social email the user becomes a MEMBER
+        # of the inviting org immediately, skipping the create-org prompt.
+        # When there is no matching invitation, provision_tenant_for_user
+        # returns None and the user stays membership-less (gated), preserving
+        # the Phase 5 uninvited-stays-gated behavior.
+        # UserAlreadyHasMembershipError is treated as a no-op so that a social
+        # re-login for a user who is already a member does not raise.
+        self._provision_org_membership(user)
```
hugobessa, line 100: Hook placed at the END of `save_user`, after `super().save_user()` commits the user (pk set) and `Profile.get_or_create` satisfies the profile invariant. Verified against allauth's social flow: `save_user` runs only on signup (auto-signup and form paths both persist the user before this point) and is NOT called on re-login — so this never acts on an unsaved user, and the idempotency swallow is defense-in-depth rather than load-bearing.

accounts/account_adapters.py
```diff
@@ -74,8 +75,68 @@ def save_user(self, request, sociallogin, form=None):
...
+    @staticmethod
+    def _provision_org_membership(user: User) -> None:
+        """Attempt to auto-join the user to an inviting organisation.
+
+        Resolves OrganizationService from the DI container (same pattern as
+        accounts/signals.py) and calls provision_tenant_for_user with no
+        organisation_name — social signups never carry one.  No-ops silently on
+        UserAlreadyHasMembershipError (re-login or race).
+        """
+        if not user.email:
```
hugobessa, line 111: Blank-email guard: `provision_tenant_for_user` matches invitations by `email__iexact`, so a blank `user.email` could match an invitation row with email "". `EMAIL_REQUIRED` makes this unreachable in the normal flow, but the guard mirrors the `user is None` defensiveness in `accounts/signals.py` and protects against custom/edge providers.

organizations/services.py
```diff
@@ -221,11 +221,12 @@ def provision_tenant_for_user(
 
         if pending_invitation is not None:
             try:
-                membership = OrganizationMembership.objects.create(
-                    user=user,
-                    organization=pending_invitation.organization,
-                    role=OrganizationRole.MEMBER,
-                )
+                with transaction.atomic():
```
hugobessa, line 224: Savepoint hygiene fix surfaced by this phase's review: both `create()` sites in `provision_tenant_for_user` now run inside an inner `with transaction.atomic()` so a concurrent unique-violation rolls back just the savepoint and the typed `UserAlreadyHasMembershipError` propagates over a usable transaction. Phase 1 already did this for `accept_invitation`; this closes the same gap on the service that the email-confirm hook AND the social adapter both call.

PR #38, opened 2026-06-04
feat(organizations): reject already-member invite acceptance with a clear 400

hugobessa:
## Summary
Phase 7 of the **Auth–Tenant Provisioning** plan. Stacked on Phase 6 (PR #37). Implements **Use-case 5** (invited address that already has an account and a membership).

`accept_invitation` already raises the typed `UserAlreadyHasMembershipError` before any DB write (Phase 1). This phase surfaces it at `AcceptInvitationView` as a clean, stable 400 — matching the endpoint's existing `{"error": ...}` body shape — and locks the no-state-change guarantee: the user's existing membership/org/role is untouched and the target invitation stays pending (revocable by an admin).

## Changes
- `organizations/views.py` — explicit `except UserAlreadyHasMembershipError` in `AcceptInvitationView.create()` returning `{"error": "User already belongs to an organization."}` with HTTP 400. Without it, the DRF default handler would have rendered a list-shaped `ValidationError` body inconsistent with the sibling handlers.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 7 — "Reject already-member invite acceptance at the API".
- Spec use-case: Use-case 5 (Acceptance scenario 5).

## Test plan
- `uv run pytest -n auto` (full suite: 1204 passed)

New tests: already-member user POSTs a valid token for a DIFFERENT org → 400 `{"error": ...}`; original membership (org+role) unchanged; target invitation still pending (`accepted_at is None`, `membership is None`); no second membership; plus a body-shape contract test.

### DIFF
--- a/organizations/views.py
+++ b/organizations/views.py
@@ -13,6 +13,7 @@ from organizations.exceptions import (
     DuplicateInvitationError,
     InvalidInvitationTokenError,
     InvitationNotFoundError,
+    UserAlreadyHasMembershipError,
 )
 from organizations.filtersets import OrganizationInvitationFilterSet
 from organizations.models import Organization, OrganizationInvitation
@@ -102,6 +103,11 @@ class AcceptInvitationView(generics.CreateAPIView):
 
         try:
             membership = serializer.create(serializer.validated_data)
+        except UserAlreadyHasMembershipError:
+            return Response(
+                {"error": UserAlreadyHasMembershipError.default_detail},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
         except InvalidInvitationTokenError as e:
             return Response(
                 {"error": str(e)},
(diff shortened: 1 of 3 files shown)


organizations/views.py
```diff
@@ -102,6 +103,11 @@ def create(self, request, *args, **kwargs):
 
         try:
             membership = serializer.create(serializer.validated_data)
+        except UserAlreadyHasMembershipError:
```
hugobessa, line 106: `UserAlreadyHasMembershipError` subclasses DRF `ValidationError`, so without this explicit catch it would reach DRF's default handler and render as a list-shaped body (`["User already belongs to an organization."]`) — inconsistent with the sibling handlers in this same method (`InvalidInvitationTokenError` etc.) which all return `{"error": ...}`. The handler uses `default_detail` rather than `str(e)` because `str()` on a DRF ValidationError yields the `[ErrorDetail(...)]` repr, not the clean message. Net: one consistent `{"error": ...}` contract across every 4xx this endpoint returns.

PR #39, opened 2026-06-04
fix(organizations): close hard gate — refuse membership-less users at all tenant-scoped endpoints

hugobessa:
## Summary
Phase 8 (FINAL) of the **Auth–Tenant Provisioning** plan. Stacked on Phase 7 (PR #38). Closes the spec's hard-gate objective: an authenticated user with no membership is uniformly refused at every tenant-scoped endpoint (empty queryset / permission denial / clean response) instead of getting a 500 from `RelatedObjectDoesNotExist`. Only onboarding (`POST /organizations/`, `POST /invitations/accept`) and account-level/self endpoints remain reachable for gated users.

## Audit + patches
Audited every `user.organization_membership` read site across all apps. **12 unguarded gaps patched** in `calendar_integration` (filtersets, serializers `__init__`/`create`/`save`, viewset actions) and `webhooks` (serializer create). Each gap now guards with `getattr(user, "organization_membership", None)` and early-returns an empty result for gated users — the membership-present path is byte-for-byte unchanged.

Secondary fix: empty querysets on `OrganizationModel` subclasses now use `original_manager.none()` instead of `objects.none()`. The tenant-safe manager raises `ImproperlyConfigured` when an unfiltered queryset is iterated, so the pre-existing `objects.none()` calls (in the unauthenticated branches) were latent bugs; `original_manager.none()` returns a plain empty queryset cleanly.

Also: guarded `CalendarGroupPermission.has_object_permission` (was an unguarded deref — latent 500), and added the hard-gate invariant docstring to `OrganizationMembership`.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 8 — "Audit + close the hard gate".
- Spec: Objective 4 (hard gate) + Acceptance scenario 6. Open Question 2 default honored — gated users keep account self-management + sign-out.

## Test plan
- `uv run pytest calendar_integration/tests/ accounts/tests/ organizations/tests/ -n auto`
- `uv run pytest -n auto` (full suite: 1226 passed)

New `accounts/tests/test_hard_gate_phase8.py` — membership-less refusal across REST tenant surfaces (list + create/bulk writes → clean 4xx/empty, never 500), onboarding endpoints stay reachable, member path still works.


organizations/models.py
```diff
@@ -96,6 +96,21 @@ class OrganizationMembership(BaseModel):
     """
     Represents a membership of a user in a calendar organization.
     This is used to link users to their respective calendar organizations.
+
```
hugobessa, line 99: The invariant this whole feature establishes, documented at its anchor: an authenticated user is EITHER gated (no membership — only onboarding responds) OR has exactly one membership. Any code reading `user.organization_membership` must guard with `getattr(..., None)` (or the equivalent `hasattr` / `DoesNotExist` try-except) so a gated user gets a clean refusal, never a 500.

calendar_integration/views.py
```diff
@@ -282,13 +282,15 @@ class CalendarEventViewSet(VintaScheduleModelViewSet):
...
+        raising Http404, so the response is a clean empty list / 404-on-object
+        rather than a 500.
         """
-        try:
-            organization_id = self.request.user.organization_membership.organization_id
-        except OrganizationMembership.DoesNotExist as e:
-            raise Http404("Calendar organization not found for the user.") from e
-        queryset = super().get_queryset().filter_by_organization(organization_id)
-        return queryset
+        membership = getattr(self.request.user, "organization_membership", None)
```
hugobessa, line 290: Representative gap closure: this previously raised `Http404` (and the bare `user.organization_membership.organization_id` deref would 500 for a gated user). Now it early-returns an empty queryset. The membership-PRESENT branch is unchanged — the guard only adds the gated-user short-circuit, so existing users see identical results.

calendar_integration/views.py
```diff
@@ -74,14 +74,14 @@ def get_queryset(self):
...
-            return Calendar.objects.none()
+            return Calendar.original_manager.none()
 
         try:
             organization_id = user.organization_membership.organization_id
             return super().get_queryset().filter_by_organization(organization_id)
         except OrganizationMembership.DoesNotExist:
             # If user has no calendar organization membership, return empty queryset
-            return Calendar.objects.none()
+            return Calendar.original_manager.none()
```
hugobessa, line 84: `original_manager.none()`, not `objects.none()`. `Calendar` (an `OrganizationModel`) uses a tenant-safe manager whose queryset raises `ImproperlyConfigured` when iterated without an organization filter — so `objects.none()` here was a latent 500 (DRF iterates the queryset to paginate). `original_manager` is the plain Django manager on `OrganizationModel`; `.none()` on it yields a genuinely empty, safely-iterable queryset. Verified against the running container.

calendar_integration/permissions.py
```diff
@@ -67,7 +67,8 @@ def has_permission(self, request, view):
         return getattr(user, "organization_membership", None) is not None
 
     def has_object_permission(self, request, view, obj):
```
hugobessa, line 69: Guarded `has_object_permission` so a membership-less user can never trip an unguarded `request.user.organization_membership` deref (a latent 500). `has_permission` already blocks gated users at the list level, so this is defense-in-depth for object-level checks; `membership.organization_id` resolves identically to the old expression for membership-present users.

PR #40, opened 2026-06-04
build(accounts): add export_allauth_schema --check pre-commit drift guard

hugobessa:
## Summary
`schema-auth.yml` is generated by `python manage.py export_allauth_schema` (it introspects the allauth config + the signup form), but — unlike `schema.yml` — nothing regenerated or verified it on commit. It drifted silently: the stale `username` field lingered after `ACCOUNT_USER_MODEL_USERNAME_FIELD = None`, and the Auth–Tenant Provisioning work added `organization_name` to the signup form without re-exporting. This adds a guard so that can't happen again.

## Changes
- `export_allauth_schema` gains a **`--check`** mode: regenerate the spec in memory, diff against the file on disk, exit non-zero on drift (no write). Mirrors `makemigrations --check`.
- New **`backend-auth-schema-local`** pre-commit hook runs `export_allauth_schema --check`, scoped to auth-relevant inputs (`settings/`, `base_forms.py`, `account_adapters.py`, `pyproject.toml`, `schema-auth.yml`) so it doesn't run on every commit.
- Regenerated `schema-auth.yml` to drop the stale `username` field, so the guard passes on `main`.

## Note on the feature stack
The Auth–Tenant Provisioning feature (PRs #32–#39) carries its own regenerated `schema-auth.yml` (adds `organization_name`, also drops `username`). When both this guard and that feature land, `schema-auth.yml` will be current either merge order — both regenerations agree on the `username` removal; the feature additionally adds `organization_name`. A `schema-auth.yml` merge conflict, if any, resolves to "drop username + add organization_name".

## Test plan
- `uv run python manage.py export_allauth_schema --check` → exits 0 ("up to date") on this branch.
- Dirty the file → `--check` exits 1. Verified.
- `uv run pytest -n auto` (unaffected; command-only change).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,3 +71,5 @@ repos:
         # Allauth headless spec depends on auth settings + the signup form/adapters.
         files: (.*/?settings/.+|.+base_forms\.py|.+account_adapters\.py|.+pyproject\.toml|^schema-auth\.yml$)
         pass_filenames: false
+        env:
+          DJANGO_SETTINGS_MODULE: config.settings.local
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,5 +71,3 @@ repos:
         # Allauth headless spec depends on auth settings + the signup form/adapters.
         files: (.*/?settings/.+|.+base_forms\.py|.+account_adapters\.py|.+pyproject\.toml|^schema-auth\.yml$)
         pass_filenames: false
-        env:
-          DJANGO_SETTINGS_MODULE: config.settings.local
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,3 +64,10 @@ repos:
         language: system
         files: ^
         pass_filenames: false
+      - id: backend-auth-schema
+        name: backend-auth-schema-local
+        entry: uv run python manage.py export_allauth_schema --check
+        language: system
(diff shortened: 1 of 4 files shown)


accounts/management/commands/export_allauth_schema.py
```diff
@@ -18,6 +18,14 @@ def add_arguments(self, parser):
             default="schema-auth.yml",
             help="Output path (.json, .yaml or .yml). Defaults to ./schema-auth.yml",
         )
+        parser.add_argument(
+            "--check",
```
hugobessa, line 22: `--check` regenerates the spec in memory and compares to the file on disk, raising `CommandError` (non-zero exit) on drift without writing — the CI/pre-commit-friendly counterpart to the existing write mode, mirroring `makemigrations --check`. Rendering is factored into `_render` so write and check paths serialize identically (sorted keys → deterministic diff).

.pre-commit-config.yaml
```diff
@@ -64,3 +64,10 @@ repos:
         language: system
         files: ^
         pass_filenames: false
+      - id: backend-auth-schema
```
hugobessa, line 67: Scoped `files:` to auth-schema inputs (settings, signup form, adapters, pyproject, and schema-auth.yml itself) rather than the always-run `files: ^` that `backend-schema` uses — the allauth spec only changes when those inputs change, so this keeps the common commit fast while still catching drift at its source.

PR #41, opened 2026-06-04
feat(organizations): add GET /organizations/current/ membership read endpoint

hugobessa:
## Summary
Phase 9 (appended 2026-06-04) of the **Auth–Tenant Provisioning** plan. Stacked on Phase 8 (PR #39).

Phases 1–8 *enforce* the hard gate but ship no way for the frontend to *read* it — there was no endpoint exposing the caller's organization/role (`OrganizationViewSet` is NoList; `/profile/me/` has no membership; the allauth session `User` has none). Without it the frontend can't distinguish "authenticated + no membership" (gated → route to onboarding) from "onboarded". This adds the read signal.

## Endpoint
`GET /organizations/current/`
- **200** — onboarded: `{ "role": "admin"|"member", "organization": { id, name, should_sync_rooms, created, modified } }`
- **404** — gated (no membership): `{ "detail": "No organization membership." }`
- **401** — unauthenticated

The action overrides `permission_classes=[IsAuthenticated]` (NOT the viewset's `OrganizationManagementPermission`, which returns False for members) so both gated and onboarded users can call it.

## Changes
- `organizations/views.py` — `current` action on `OrganizationViewSet` (`@action(detail=False, url_path="current", permission_classes=[IsAuthenticated])`), `getattr` guard → 404 when gated, `@extend_schema` for 200/404.
- `organizations/serializers.py` — `CurrentMembershipSerializer` (read-only: `role` + nested `organization`).
- `schema.yml` — regenerated; documents `/organizations/current/`.

## Plan reference
- Plan: `AUTH_TENANT_PROVISIONING`, Phase 9. Read counterpart to Phase 8's hard-gate enforcement.

## Test plan
- `uv run python manage.py spectacular --file schema.yml` (path present)
- `uv run pytest -n auto` (full suite: 1231 passed)

Covers admin/member 200, gated 404, unauthenticated 401, and an explicit member-not-blocked regression.


organizations/views.py
```diff
@@ -44,6 +48,26 @@ def get_queryset(self):
...
 
+    @extend_schema(
+        summary="Current organization + role for the authenticated user",
+        responses={
+            200: CurrentMembershipSerializer,
+            404: OpenApiResponse(description="No organization membership (gated user)"),
+        },
+    )
+    @action(detail=False, methods=["get"], url_path="current", permission_classes=[IsAuthenticated])
+    def current(self, request):
```
hugobessa, line 59: The `permission_classes=[IsAuthenticated]` override is load-bearing: the viewset's default `OrganizationManagementPermission.has_permission` returns False for users who ALREADY have a membership, so without the override an onboarded member would get 403 on their own current-org probe. This action must serve both gated AND onboarded users — it is the onboarding-state signal, deliberately NOT behind the hard gate. The `test_current_not_blocked_by_organization_management_permission` test locks this.

organizations/views.py
```diff
@@ -44,6 +48,26 @@ def get_queryset(self):
...
+        """Return the caller's organization and role.
+
+        HTTP 200 — the user is onboarded (has a membership).
+        HTTP 404 — the user is gated (no membership yet).
+        """
```
hugobessa, line 64: 404 (not an empty 200) is the gated signal — unambiguous for the frontend: 200 ⇒ onboarded (route to the app), 404 ⇒ gated (route to create-org / accept-invite). `getattr(..., None)` keeps a membership-less user from tripping `RelatedObjectDoesNotExist` (a 500), consistent with the Phase 8 guard pattern.

organizations/serializers.py
```diff
@@ -113,6 +113,25 @@ def create(self, validated_data: dict) -> OrganizationInvitation:
         return invitation
 
 
+class CurrentMembershipSerializer(serializers.ModelSerializer):
```
hugobessa, line 116: `organization` is a `SerializerMethodField` rather than a nested serializer field because `OrganizationSerializer.__init__` is `@inject`-decorated (dependency-injector), so instantiating it at class-definition time fails. Building it inside `get_organization` defers construction to request time when the DI container is wired — matching how the rest of the codebase consumes DI-injected serializers.

PR #42, opened 2026-06-05
feat(organizations): add reusable IsOrganizationAdmin permission

hugobessa:
## Summary
Phase 0 of the **REST API Frontend Gaps** plan. Adds a single reusable `IsOrganizationAdmin` DRF permission class so every admin-only endpoint added in later phases (members list/disable, room sync, event transfer, bundle update, public-API token management) shares one admin-gating implementation.

The permission:
- `has_permission` — authenticated user with an organization membership (safe `getattr`, preserving the hard-gate invariant: membership-less users get a clean denial, never a 500).
- `has_object_permission` — the object's organization must match the caller's membership organization **and** the caller must be an admin, delegated to the existing `User.is_organization_admin(organization_id)` (single source of truth). Handles both `Organization` instances and `OrganizationModel` subclasses.

No behavior change on its own — nothing consumes the permission yet. Pure additive scaffolding.

## Plan reference
- Plan: `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 0.
- Guiding Decision **Admin gating**: one `IsOrganizationAdmin` delegating to `User.is_organization_admin`.

## Test plan
- `uv run pytest organizations/tests/test_permissions.py -vs` — 11 cases (admin allowed, member denied, membership-less denied, cross-org denied, Organization + OrganizationModel object paths).
- Outer gate green: `ruff check`, `ruff format --check`, `makemigrations --check`, `manage.py check --deploy`, `pytest -n auto` (1243 passed).

### DIFF
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -1,8 +1,14 @@
+from typing import TYPE_CHECKING
+
 from django.core.exceptions import ObjectDoesNotExist
 
 from rest_framework.permissions import BasePermission
 
 from organizations.models import Organization, OrganizationInvitation, OrganizationModel
+
+
+if TYPE_CHECKING:
+    from users.models import User
 
 
 class OrganizationManagementPermission(BasePermission):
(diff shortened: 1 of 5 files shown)


organizations/permissions.py
```diff
@@ -61,3 +67,41 @@ def has_object_permission(self, request, view, obj):
...
+    - `has_object_permission`: delegates the "is this user an admin of this object's org"
+      decision to `User.is_organization_admin(organization_id)` so the rule has a single
+      implementation. Handles both Organization instances and OrganizationModel subclasses.
+    """
+
+    def has_permission(self, request, view) -> bool:
+        user: User = request.user
+        if not user or not user.is_authenticated:
+            return False
+        return getattr(user, "organization_membership", None) is not None
```
hugobessa, line 87: `has_permission` intentionally checks only membership *presence* here, not activity. `OrganizationMembership.is_active` does not exist yet — it lands in Phase 1, which will tighten this gate to also reject inactive memberships (the tenant-scoped "disable a user" path). Keeping Phase 0 free of an is_active reference keeps the field addition + gate change atomic in Phase 1.

organizations/permissions.py
```diff
@@ -61,3 +67,41 @@ def has_object_permission(self, request, view, obj):
...
+        elif isinstance(obj, OrganizationModel):
+            obj_organization_id = obj.organization_id
+        else:
+            return False
+
+        # Membership org must match object org; user must be an admin
+        if membership.organization_id != obj_organization_id:
+            return False
+
+        return user.is_organization_admin(membership.organization_id)
```
hugobessa, line 107: Admin check is delegated to `User.is_organization_admin(organization_id)` (already used by `CalendarPermissionService.can_manage_calendar_group`) so the "is this user an org admin" rule has exactly one implementation. Every admin endpoint in later phases reuses this permission rather than re-deriving the role check — per the plan's Guiding Decisions (Admin gating).

organizations/permissions.py
```diff
@@ -61,3 +67,41 @@ def has_object_permission(self, request, view, obj):
...
+        if membership is None:
+            return False
+
+        # Determine the object's organization_id
+        if isinstance(obj, Organization):
+            obj_organization_id = obj.id
+        elif isinstance(obj, OrganizationModel):
+            obj_organization_id = obj.organization_id
+        else:
+            return False
```
hugobessa, line 101: The Organization-vs-OrganizationModel branch mirrors the existing `OrganizationManagementPermission` so this class works whether the viewset's object is the Organization itself (uses `.id`) or any tenant-scoped model (uses `.organization_id`, e.g. Calendar, CalendarEvent, SystemUser). Objects of neither type are denied rather than silently allowed.

PR #43, opened 2026-06-05
feat(organizations): add is_active to membership + gate inactive members everywhere

hugobessa:
## Summary
Phase 1 of the **REST API Frontend Gaps** plan. Adds `OrganizationMembership.is_active` (the storage behind the upcoming tenant-scoped "disable a user" action) and — critically — closes the tenant-access gate against inactive memberships across the whole codebase.

The load-bearing insight: a *deactivated* member still has a membership row, so every place that resolved `user.organization_membership` to grant access would keep granting it. This PR introduces one canonical helper, `get_active_organization_membership(user)` (returns `None` for missing **or** inactive membership), and routes every tenant-access gate through it — read **and** write paths, REST permissions, GraphQL, and the service-layer admin check.

## What changed
- New `is_active` field (`default=True`, `db_default=True`, `db_index=True`) — additive, safe backfill, default-active so existing reads are unchanged.
- `get_active_organization_membership` helper in `organizations/models.py`.
- Gate closed at: all viewset `get_queryset` methods (calendar, events, blocked/available times, calendar-groups, organizations, invitations, webhooks); `IsOrganizationAdmin`, `CalendarGroupPermission`, `OrganizationInvitationPermission`, `OrganizationManagementPermission`; serializer create/save write paths (calendars, bundles, blocked/available times + bulk, webhook config); `OrganizationViewSet.current`; the public GraphQL `users` query; and `User.is_organization_admin`.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 1 + Guiding Decision "Disable user = deactivate membership".

## Review history
Adversarial review caught two security gaps in the first pass (inactive member could still **write** via serializer `create`, and retrieve/update/destroy their Organization via `OrganizationManagementPermission`) plus a public-API user leak — all fixed in the second commit, with regression tests added.

## Test plan
- `uv run pytest organizations/tests/ calendar_integration/tests/ webhooks/tests/ public_api/tests/ -n auto`
- New tests assert: inactive member → 404 on `/organizations/current/`, denied GET/PATCH/DELETE on their org, blocked from webhook create (row not created), excluded from the GraphQL users query; active members unaffected.
- Outer gate green: ruff, ruff format --check, mypy (no new errors), makemigrations --check, check --deploy, `pytest -n auto` (1262 passed).


organizations/models.py
```diff
@@ -7,6 +11,33 @@
...
+    membership — no tenant access, no 500. An active membership returns normally.
+    """
+    membership = getattr(user, "organization_membership", None)
+    if membership is None:
+        return None
+    if not membership.is_active:
+        return None
+    return membership
+
+
```
hugobessa, line 40: This helper is the single source of truth for tenant-access gating from here on. It collapses two states into one: a membership-less user (reverse OneToOne raises) and a deactivated member (row exists, is_active=False) both return None. Every gate — get_queryset, permissions, serializer writes, GraphQL — calls this instead of touching `user.organization_membership` directly, so "disabled user loses access" is enforced in exactly one place.

calendar_integration/serializers.py
```diff
@@ -121,7 +121,7 @@ def __init__(
 
     def create(self, validated_data):
         user = self.context["request"].user
-        membership = getattr(user, "organization_membership", None)
+        membership = get_active_organization_membership(user)
```
hugobessa, line 124: Write-path gating (not just reads). The view-level permission on these endpoints only checks `is_authenticated`, so the membership/active check has to happen here in `create`/`save` — otherwise a deactivated member could still POST calendars / blocked-times / available-times / webhook configs against their old org. This (and the matching create/save methods) was a BLOCKER surfaced in review; the raw `getattr` it replaced returned the inactive row and made the `if not membership` guard a no-op.

organizations/permissions.py
```diff
@@ -13,29 +16,42 @@
...
         except AttributeError:
             return True
 
+        membership = get_active_organization_membership(user)
+        if membership is None:
+            # Membership-less OR inactive members never have object-level
+            # access (they can only CREATE an org — handled in has_permission).
+            return False
+
+        return view.action != "create" and (
```
hugobessa, line 47: `OrganizationManagementPermission.has_object_permission` now denies any user without an *active* membership. Previously it relied on `except ObjectDoesNotExist` to detect membership-less users — but an inactive member's row exists and never raises, so the old code returned True and let a disabled user retrieve/update/destroy their Organization. Note the intentional split: `has_permission` lets inactive/membership-less users reach POST /organizations/ (onboarding), but the service layer refuses to create a second membership for someone who already has a row (see services.py).

organizations/services.py
```diff
@@ -156,6 +156,10 @@ def accept_invitation(self, token: str, user: User) -> OrganizationMembership:
         :param user: User who is accepting the invitation.
         :return: Created OrganizationMembership instance.
         """
+        # DESIGN: hasattr check intentionally matches BOTH active and inactive memberships.
+        # An inactive membership still has a DB row, so accept_invitation must refuse
+        # re-provisioning here.  The supported un-disable path is admin REACTIVATION
+        # (setting is_active=True); not re-accepting an invitation.
         if hasattr(user, "organization_membership"):
```
hugobessa, line 163: Deliberate design note, not a bug: the duplicate-membership guard matches both active and inactive rows, so a deactivated user cannot re-onboard by accepting a new invitation. The supported un-disable path is admin REACTIVATION (Phase 3 flips is_active back to True), not re-provisioning. Flagged so a future reader doesn't "fix" this into allowing duplicate rows.

organizations/migrations/0005_organizationmembership_is_active.py
```diff
@@ -0,0 +1,23 @@
...
+    dependencies = [
+        ("organizations", "0004_organizationmembership_role"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="organizationmembership",
+            name="is_active",
+            field=models.BooleanField(
+                db_default=True,
```
hugobessa, line 17: `db_default=True` (alongside `default=True`) makes this a safe, single-step additive column on an existing table: Postgres backfills existing rows at the DB level with no separate data migration and no table rewrite. Default-active is what keeps every pre-feature read byte-for-byte unchanged until a member is explicitly deactivated.

PR #44, opened 2026-06-05
feat(organizations): add admin organization members list endpoint

hugobessa:
## Summary
Phase 2 of the **REST API Frontend Gaps** plan. Adds `GET /organization-members/` (list) and `GET /organization-members/{id}/` (retrieve), admin-only, for the team datatable. Lists **both active and inactive** members so admins can see (and later re-enable) disabled users.

Also hardens the shared `IsOrganizationAdmin` permission: its `has_permission` now requires an active **admin** membership (previously only required membership), so the permission correctly gates collection actions (list/create) — not just object actions. This makes it genuinely reusable by every later admin endpoint without per-view workarounds.

## What changed
- `OrganizationMembershipViewSet` (`ReadOnlyVintaScheduleModelViewSet`, `IsOrganizationAdmin`), org-scoped queryset.
- `OrganizationMembershipSerializer` — read-only, exposes id/role/is_active + flattened user email + profile name.
- `IsOrganizationAdmin.has_permission` tightened to require `membership.is_admin`.
- Route `organization-members`; `schema.yml` regenerated.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 2 + API Design 4.1.

## Review history
Layer-3 review confirmed the permission flaw the implementer flagged: a list endpoint only triggers `has_permission`, so `IsOrganizationAdmin` was letting any active member list members. Fixed in the second commit (admin check moved into `has_permission`; per-view `check_permissions` override removed); regression test added asserting a non-admin member gets 403 on the list action.

## Test plan
- Asserts: admin list 200 with user fields (active+inactive); non-admin 403 on list; membership-less 403; inactive 403; cross-org members excluded; retrieve cross-org id 404.
- Outer gate green: ruff, format --check, mypy (no new errors), makemigrations --check, check --deploy, `pytest -n auto` (1272 passed).

### DIFF
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -5,6 +5,7 @@ from rest_framework.permissions import BasePermission
 from organizations.models import (
     Organization,
     OrganizationInvitation,
+    OrganizationMembership,
     OrganizationModel,
     get_active_organization_membership,
 )
@@ -105,6 +106,8 @@ class IsOrganizationAdmin(BasePermission):
         # Determine the object's organization_id
         if isinstance(obj, Organization):
             obj_organization_id = obj.id
+        elif isinstance(obj, OrganizationMembership):
+            obj_organization_id = obj.organization_id
         elif isinstance(obj, OrganizationModel):
             obj_organization_id = obj.organization_id
         else:
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -84,9 +84,10 @@ class IsOrganizationAdmin(BasePermission):
     """
     Permission for admin-only endpoints within an organization.
 
-    - `has_permission`: requires an authenticated user with an active
-      organization membership (using safe `getattr` to handle membership-less users).
-    - `has_object_permission`: delegates the "is this user an admin of this object's org"
+    - `has_permission`: requires an authenticated user with an active ADMIN organization
+      membership. This gate enforces the admin role at the collection level (list, create).
+    - `has_object_permission`: additionally enforces that the object's organization matches
+      the membership organization and delegates the "is this user an admin of this object's org"
       decision to `User.is_organization_admin(organization_id)` so the rule has a single
       implementation. Handles both Organization instances and OrganizationModel subclasses.
     """
@@ -95,7 +96,8 @@ class IsOrganizationAdmin(BasePermission):
         user: User = request.user
         if not user or not user.is_authenticated:
             return False
-        return get_active_organization_membership(user) is not None
+        membership = get_active_organization_membership(user)
+        return membership is not None and membership.is_admin
 
     def has_object_permission(self, request, view, obj) -> bool:
         user: User = request.user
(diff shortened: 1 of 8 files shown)


organizations/permissions.py
```diff
@@ -94,7 +96,8 @@ def has_permission(self, request, view) -> bool:
         user: User = request.user
         if not user or not user.is_authenticated:
             return False
-        return get_active_organization_membership(user) is not None
+        membership = get_active_organization_membership(user)
+        return membership is not None and membership.is_admin
```
hugobessa, line 100: This is the load-bearing change for every future admin endpoint. DRF only calls `has_object_permission` on detail actions; for list/create it calls `has_permission` alone. So the admin check has to live HERE, not only in has_object_permission — otherwise an admin-only collection endpoint silently admits any active member. Phases 3/6/7/8/12-15 reuse this permission as-is; none of them need a per-view admin check now.

organizations/serializers.py
```diff
@@ -132,6 +132,31 @@ def get_organization(self, obj: OrganizationMembership) -> dict:
...
+        model = OrganizationMembership
+        fields = (
+            "id",
+            "role",
+            "is_active",
+            "user_email",
+            "user_first_name",
+            "user_last_name",
+        )
+        read_only_fields = fields
```
hugobessa, line 157: Deliberately read-only and field-explicit: only id/role/is_active plus the user's email and profile first/last name are exposed — no password hash, phone, tokens, or other PII reach the datatable. `read_only_fields = fields` makes the whole serializer non-writable (writes happen via Phase 3's deactivate/reactivate actions, not PUT/PATCH here).

organizations/views.py
```diff
@@ -117,6 +121,31 @@ def perform_destroy(self, instance):
...
 
+class OrganizationMembershipViewSet(ReadOnlyVintaScheduleModelViewSet):
+    """
+    A viewset for listing and retrieving organization members.
+
+    Admin-only endpoint — lists both active and inactive members of the caller's
+    organization, suitable for a datatable view. Non-admin members get 403.
+    """
+
+    queryset = OrganizationMembership.objects.select_related("user", "user__profile")
```
hugobessa, line 132: `select_related("user", "user__profile")` keeps the datatable to a single query regardless of member count — the serializer's flattened `user.email` / `user.profile.first_name` fields would otherwise N+1 per row.

organizations/views.py
```diff
@@ -117,6 +121,31 @@ def perform_destroy(self, instance):
...
+        """Org-scoped queryset: return members of the caller's organization only."""
+        user = self.request.user
+        membership = get_active_organization_membership(user)
+        if membership:
+            return (
+                OrganizationMembership.objects.filter(organization_id=membership.organization_id)
+                .select_related("user", "user__profile")
+                .order_by("id")
+            )
+        return OrganizationMembership.objects.none()
```
hugobessa, line 146: Unlike the hard-gate (which hides an inactive *caller's* own access), this list intentionally returns the org's inactive members too — the admin needs to see disabled teammates in order to reactivate them in Phase 3. Scoping is still strictly the caller's own org via the active-membership helper.

PR #45, opened 2026-06-05
feat(organizations): add member deactivate/reactivate actions

hugobessa:
## Summary
Phase 3 of the **REST API Frontend Gaps** plan — the admin "disable a user" action. Adds two admin-only actions on `OrganizationMembershipViewSet`:
- `POST /organization-members/{id}/deactivate/` — sets `is_active=False` (tenant-scoped disable; the member is then gated everywhere via the Phase 1 helper).
- `POST /organization-members/{id}/reactivate/` — sets `is_active=True` (the supported un-disable path).

Both are detail actions, so `IsOrganizationAdmin.has_object_permission` enforces admin + same-org (cross-org → 404, non-admin → 403). Both are idempotent.

## Guards
- **Self-lockout**: an admin cannot deactivate their own membership (403).
- **Last-admin**: defense-in-depth check that refuses removing the org's last active admin (400). Documented as currently unreachable through this endpoint — see the inline comment.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 3 + API Design 4.1.

## Review history
Layer-3 review flagged a "missing last-admin-refusal test", which led to a useful finding: that 400 branch is actually unreachable via HTTP (the requester must be an active admin, so they're always counted as another admin unless they're the target — in which case the self-lockout guard fires first). The real invariant (org never loses its last admin) is enforced by the self-lockout guard and is tested. The docs commit clarifies this and adds `test_sole_admin_cannot_self_deactivate`.

## Test plan
- `uv run pytest organizations/tests/test_views.py -n auto`
- Asserts: admin deactivates a member → 200 + DB false + member then gated on `/organizations/current/`; reactivate restores access; admin cannot deactivate self (403, sole-admin case included); non-admin → 403; cross-org → 404; idempotent re-calls → 200.
- Outer gate green: ruff, format --check, mypy (108 full-project errors, unchanged from base — zero new), makemigrations --check, check --deploy, `pytest -n auto` (1285 passed).

### DIFF
--- a/organizations/views.py
+++ b/organizations/views.py
@@ -4,7 +4,7 @@ from dependency_injector.wiring import Provide, inject
 from drf_spectacular.utils import OpenApiResponse, extend_schema
 from rest_framework import generics, status
 from rest_framework.decorators import action
-from rest_framework.exceptions import NotFound
+from rest_framework.exceptions import NotFound, PermissionDenied, ValidationError
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
@@ -123,10 +123,15 @@ class OrganizationInvitationViewSet(NoUpdateVintaScheduleModelViewSet):
 
 class OrganizationMembershipViewSet(ReadOnlyVintaScheduleModelViewSet):
     """
-    A viewset for listing and retrieving organization members.
+    A viewset for listing, retrieving, and managing organization members.
 
     Admin-only endpoint — lists both active and inactive members of the caller's
     organization, suitable for a datatable view. Non-admin members get 403.
+
+    Actions:
+    - `deactivate`: POST to disable a member (prevent self-deactivation and
+      protect the last active admin).
+    - `reactivate`: POST to re-enable a member.
     """
 
     queryset = OrganizationMembership.objects.select_related("user", "user__profile")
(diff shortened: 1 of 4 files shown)


organizations/views.py
```diff
@@ -145,6 +150,92 @@ def get_queryset(self):
...
+        Idempotency: deactivating an already-inactive member is a no-op success.
+        """
+        target = (
+            self.get_object()
+        )  # Permission checks via IsOrganizationAdmin.has_object_permission
+        user = request.user
+
+        # Guard: prevent self-deactivation
+        if target.user_id == user.id:
+            raise PermissionDenied(detail="Cannot deactivate your own membership.")
```
hugobessa, line 179: The self-lockout guard is what actually guarantees the org never loses its last admin: the only way to drop to zero active admins through this endpoint is an admin deactivating themselves, which this 403 blocks. It runs before the last-admin guard below by design.

organizations/views.py
```diff
@@ -145,6 +150,92 @@ def get_queryset(self):
...
+                    organization_id=org_id,
+                    role=target.role,  # Same role filter (ADMIN)
+                    is_active=True,
+                )
+                .exclude(id=target.id)  # Exclude the target itself
+                .count()
+            )
+            if other_active_admin_count == 0:
+                raise ValidationError(
+                    detail="Cannot deactivate the last active admin of the organization."
```
hugobessa, line 200: Defense-in-depth, intentionally kept even though it is unreachable via this HTTP action: the requester must be an active admin of the org, so when target != requester the requester is counted as "another active admin" and the count is always >= 1; when target == requester the self-lockout guard above already returned 403. The guard is retained to protect any FUTURE non-self deactivation path (bulk action / service-layer call) that wouldn't run the self-check. The inline comment in the code says the same so nobody "fixes" it as dead code.

organizations/views.py
```diff
@@ -145,6 +150,92 @@ def get_queryset(self):
...
+
+        Idempotency: reactivating an already-active member is a no-op success.
+        """
+        target = (
+            self.get_object()
+        )  # Permission checks via IsOrganizationAdmin.has_object_permission
+
+        # Reactivate (idempotent: no-op if already active)
+        target.is_active = True
+        target.save(update_fields=["is_active"])
```
hugobessa, line 233: Reactivate has no guards — re-enabling is always safe — and is the supported counterpart to deactivate. This is the un-disable path the Phase 1 design note pointed to (re-provisioning a disabled member via a new invitation is intentionally refused; reactivation here is how you bring them back).

PR #46, opened 2026-06-05
feat(calendar): add request-import action for the caller's external calendars

hugobessa:
## Summary
Phase 4 of the **REST API Frontend Gaps** plan — the import half of "request a calendar to sync". Adds `POST /calendar/request-import/`: an authenticated user triggers import of their connected external calendars (Google/Microsoft) into the app. Each connected account's import runs as a Celery task, enqueued only after the request transaction commits.

## Behavior
- Imports **all** of the caller's connected `SocialAccount`s (not just one), so a user with Google + Microsoft gets both.
- 202 Accepted with a short detail message (the underlying `request_calendars_import` returns nothing pollable).
- No connected external account → 400. Membership-less/inactive → 403. Anonymous → 401.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 4 + API Design 4.2 + Open Questions (return-shape default).

## Review history
Layer-3 review flagged that the first cut imported only `SocialAccount...first()` (one provider). Fixed to import all accounts. The multi-account fix required care: naively looping `authenticate()` on one DI service instance and deferring `request_calendars_import()` via `on_commit` would make every callback read the last account's mutated state. Resolved by allocating a fresh `CalendarService` per account and binding it into each closure by default argument; a multi-account test asserts both accounts are imported.

## Test plan
- Asserts: single account → 202 + service invoked via on_commit; two accounts → 202 + both imported (distinct instances, two enqueues); no account → 400; membership-less → 403; anonymous → 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1290 passed).

### DIFF
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -1,4 +1,5 @@
 import datetime
+from collections.abc import Callable
 from typing import Annotated
 
 from django.db import transaction
@@ -123,11 +124,11 @@ class CalendarViewSet(VintaScheduleModelViewSet):
     def request_import(
         self,
         request,
-        calendar_service: Annotated[CalendarService, Provide["calendar_service"]],
+        calendar_service_factory: Annotated[
+            Callable[[], CalendarService], Provide["calendar_service.provider"]
+        ],
     ):
         """Request import of external calendars for the authenticated user."""
-        from di_core.containers import container
-
         user = request.user
 
         membership = get_active_organization_membership(user)
@@ -145,20 +146,14 @@ class CalendarViewSet(VintaScheduleModelViewSet):
             )
 
         try:
-            # Resolve a fresh CalendarService for each account to avoid closure bug
-            # where the mutated service state gets captured by the on_commit callback
             for social_account in social_accounts:
-                # Get a fresh service instance for this account
-                if not container:
-                    raise CalendarIntegrationError("DI container not initialized")
-                fresh_service = container.calendar_service()
+                fresh_service = calendar_service_factory()
 
                 fresh_service.authenticate(
                     account=social_account,
                     organization=membership.organization,
                 )
 
-                # Capture the fresh service by default argument to avoid closure
                 def enqueue_import(service=fresh_service):
                     service.request_calendars_import()
 
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -1,6 +1,7 @@
 import datetime
 from typing import Annotated
 
+from django.db import transaction
 from django.http import Http404
 
 from allauth.socialaccount.models import SocialAccount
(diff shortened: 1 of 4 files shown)


calendar_integration/views.py
```diff
@@ -107,6 +108,74 @@ def create_bundle_calendar(self, request, *args, **kwargs):
...
+                    account=social_account,
+                    organization=membership.organization,
+                )
+
+                # Capture the fresh service by default argument to avoid closure
+                def enqueue_import(service=fresh_service):
+                    service.request_calendars_import()
+
+                transaction.on_commit(enqueue_import)
+
```
hugobessa, line 161: Imports every connected account, and the per-account fresh service is the load-bearing detail. `calendar_service` is a DI Factory and `authenticate()` mutates instance state (self.account/self.organization). If we looped with a single instance and deferred `request_calendars_import()` via on_commit, all the deferred closures would read the SAME instance's final state at commit time — only the last provider would import (possibly twice). Allocating a fresh service per account and binding it via the closure's default argument (`service=fresh_service`) gives each enqueue its own authenticated instance.

calendar_integration/views.py
```diff
@@ -107,6 +108,74 @@ def create_bundle_calendar(self, request, *args, **kwargs):
...
+                fresh_service.authenticate(
+                    account=social_account,
+                    organization=membership.organization,
+                )
+
+                # Capture the fresh service by default argument to avoid closure
+                def enqueue_import(service=fresh_service):
+                    service.request_calendars_import()
+
+                transaction.on_commit(enqueue_import)
```
hugobessa, line 160: `request_calendars_import` only does `task.delay(...)`, so deferring it to `transaction.on_commit` is exactly right under ATOMIC_REQUESTS=True: the Celery worker can't pick up the task before the request's transaction has committed (which would otherwise race the worker against uncommitted state).

calendar_integration/views.py
```diff
@@ -107,6 +108,74 @@ def create_bundle_calendar(self, request, *args, **kwargs):
...
+        methods=["post"],
+        detail=False,
+        url_path="request-import",
+        url_name="request-import",
+    )
+    @inject
+    def request_import(
+        self,
+        request,
+        calendar_service: Annotated[CalendarService, Provide["calendar_service"]],
```
hugobessa: Minor known nit: this action still declares the injected `calendar_service` param but now resolves a fresh service per account instead, so the injected one is unused here. Left in place to keep the @inject signature consistent with the other CalendarViewSet actions; safe to drop in a later cleanup.

PR #47, opened 2026-06-05
feat(calendar): add request-sync action for owned calendars

hugobessa:
## Summary
Phase 5 of the **REST API Frontend Gaps** plan — the sync half of "request a calendar to sync". Adds `POST /calendar/{id}/request-sync/`: an authenticated user syncs a calendar they OWN over a date range. Returns the created `CalendarSync` (id + status) so the frontend can poll progress.

## Behavior
- Owner-only: `get_object()` org-scopes (cross-org → 404) and an explicit `CalendarOwnership` check restricts to owners (non-owner active member → 403). Admin cross-user sync is Phase 6.
- Body validated by `CalendarSyncRequestSerializer` (`start_datetime`, `end_datetime` required; `should_update_events` optional bool) → 400 on missing/invalid.
- No linked account for the calendar's provider → 400 (explicit guard, avoids a 500).
- 202 Accepted + serialized `CalendarSync`.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 5 + API Design 4.2 + Open Questions (return sync id+status).

## Review history
Layer-3 review caught two issues, both fixed in the second commit: (1) the new `CalendarSyncRequestSerializer` was dead — the action now uses it for input validation (which also gives proper boolean coercion of `should_update_events`); (2) the `social_account` lookup could be `None` and reach `authenticate()` as a 500 — now guarded with a 400. Regression test added for the no-account case.

## Out-of-scope note
`CalendarService.request_calendar_sync` enqueues `sync_calendar_task.delay(...)` without `transaction.on_commit` (pre-existing, shared by Phase 6's admin-sync). Not changed here to avoid destabilizing task/test callers; flagged as a follow-up refactor candidate.

## Test plan
- Asserts: owner → 202 + CalendarSync id/status; non-owner → 403; cross-org → 404; missing/invalid datetimes → 400; no linked account → 400; anonymous → 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1297 passed).

### DIFF
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -208,27 +208,26 @@ class CalendarViewSet(VintaScheduleModelViewSet):
                 status=status.HTTP_403_FORBIDDEN,
             )
 
-        # Parse and validate datetime fields
-        start_datetime_str = request.data.get("start_datetime")
-        end_datetime_str = request.data.get("end_datetime")
-
-        if not start_datetime_str or not end_datetime_str:
-            raise ValidationError(
-                {"non_field_errors": ["start_datetime and end_datetime are required"]}
-            )
-
-        try:
-            start_datetime = datetime.datetime.fromisoformat(
-                start_datetime_str.replace("Z", "+00:00")
-            )
-            end_datetime = datetime.datetime.fromisoformat(end_datetime_str.replace("Z", "+00:00"))
-        except (ValueError, CalendarIntegrationError) as e:
-            raise ValidationError({"non_field_errors": [f"Invalid datetime format: {e!s}"]}) from e
-
-        should_update_events = request.data.get("should_update_events", False)
+        # Validate request input using serializer
+        input_serializer = CalendarSyncRequestSerializer(data=request.data)
+        input_serializer.is_valid(raise_exception=True)
+        data = input_serializer.validated_data
+        start_datetime = data["start_datetime"]
+        end_datetime = data["end_datetime"]
+        should_update_events = data["should_update_events"]
 
         # Get social account for authentication
         social_account = SocialAccount.objects.filter(user=user, provider=calendar.provider).first()
+
+        # Guard against missing social account
+        if social_account is None:
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        f"No linked account found for provider '{calendar.provider}'."
+                    ]
+                }
+            )
 
         membership = get_active_organization_membership(user)
         if not membership:
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -29,6 +29,7 @@ from calendar_integration.models import (
     Calendar,
     CalendarEvent,
     CalendarGroup,
+    CalendarOwnership,
 )
 from calendar_integration.permissions import (
     CalendarAvailabilityPermission,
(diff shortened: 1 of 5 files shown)


calendar_integration/views.py
```diff
@@ -176,6 +179,86 @@ def enqueue_import(service=fresh_service):
...
+        if not CalendarOwnership.objects.filter(
+            calendar=calendar,
+            user=user,
+            organization_id=calendar.organization_id,
+        ).exists():
+            return Response(
+                {"detail": "You do not own this calendar."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
```
hugobessa, line 210: Two-layer scoping: get_object() already restricts to the caller's org (so a cross-org id is a 404 before we get here), and this explicit CalendarOwnership check makes the endpoint owner-only. That's the deliberate boundary between this self-sync action and Phase 6's admin-sync — an org admin syncing someone else's calendar goes through the separate admin endpoint, not this one.

calendar_integration/views.py
```diff
@@ -176,6 +179,86 @@ def enqueue_import(service=fresh_service):
...
+        # Validate request input using serializer
+        input_serializer = CalendarSyncRequestSerializer(data=request.data)
+        input_serializer.is_valid(raise_exception=True)
+        data = input_serializer.validated_data
+        start_datetime = data["start_datetime"]
```
hugobessa, line 215: Input goes through CalendarSyncRequestSerializer rather than manual request.data parsing — this gives DRF-native required-field + ISO datetime validation and, importantly, real BooleanField coercion of should_update_events (a raw `request.data.get(...)` would treat the string "false" as truthy).

calendar_integration/views.py
```diff
@@ -176,6 +179,86 @@ def enqueue_import(service=fresh_service):
...
+        # Get social account for authentication
+        social_account = SocialAccount.objects.filter(user=user, provider=calendar.provider).first()
+
+        # Guard against missing social account
+        if social_account is None:
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        f"No linked account found for provider '{calendar.provider}'."
+                    ]
```
hugobessa, line 228: Explicit guard for the no-linked-account case (e.g. an internal calendar, or a provider the user never connected). Without it, a None account reaches `authenticate()` and surfaces as a 500; here it's a clear 400 naming the provider.

calendar_integration/views.py
```diff
@@ -176,6 +179,86 @@ def enqueue_import(service=fresh_service):
...
+
+            calendar_sync = calendar_service.request_calendar_sync(
+                calendar=calendar,
+                start_datetime=start_datetime,
+                end_datetime=end_datetime,
+                should_update_events=should_update_events,
+            )
+
+            serializer = CalendarSyncSerializer(calendar_sync)
+            return Response(serializer.data, status=status.HTTP_202_ACCEPTED)
```
hugobessa, line 253: Returns the created CalendarSync (id + status) at 202 so the frontend can poll the sync's progress — the Open Questions default in the plan. The row is created in-transaction by the service, so its id is available immediately.

PR #48, opened 2026-06-05
feat(calendar): add admin-sync action for org calendars

hugobessa:
## Summary
Phase 6 of the **REST API Frontend Gaps** plan — an org admin manually syncs **any** calendar in their organization (not just their own). Adds `POST /calendar/{id}/admin-sync/`, admin-only, returning the created `CalendarSync` (id + status) at 202.

The security-critical detail: syncing someone else's external calendar requires the **calendar owner's** OAuth credentials — the admin doesn't hold them. So the action authenticates the `CalendarService` with the owner's `SocialAccount`, never the admin's.

## Behavior
- `permission_classes=[IsOrganizationAdmin]` on the action → non-admin member 403; cross-org calendar 404 (org-scoped queryset).
- Owner resolved via `CalendarOwnership` (default owner first); no owner → 400; owner has no linked account for the provider → 400.
- Body validated by `CalendarSyncRequestSerializer` → 400 on invalid.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 6 + API Design 4.2.

## Review history
Layer-3 review: clean, no findings. The happy-path test explicitly asserts `authenticate()` received the owner's account (and that it is not the admin's).

## Out-of-scope note
Shares `CalendarService.request_calendar_sync`, whose `.delay()` is not `on_commit`-wrapped (pre-existing; flagged in Phase 5). Not changed here.

## Test plan
- Asserts: admin syncs another in-org user's calendar → 202 + authenticate called with the OWNER's account; non-admin → 403; cross-org → 404; no owner → 400; owner has no linked account → 400; invalid datetimes → 400; anon → 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1304 passed).

### DIFF
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -64,6 +64,7 @@ from calendar_integration.services.calendar_group_service import CalendarGroupSe
 from calendar_integration.services.calendar_service import CalendarService
 from common.utils.view_utils import VintaScheduleModelViewSet
 from organizations.models import get_active_organization_membership
+from organizations.permissions import IsOrganizationAdmin
 
 
 class CalendarViewSet(VintaScheduleModelViewSet):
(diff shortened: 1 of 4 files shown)


calendar_integration/views.py
```diff
@@ -259,6 +260,97 @@ def request_sync(
...
+                {"detail": "User is not an active member of any organization."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        try:
+            # Authenticate with the OWNER's account, not the admin's
+            calendar_service.authenticate(
+                account=owner_social_account,
+                organization=membership.organization,
+            )
```
hugobessa, line 335: Security-critical: the service authenticates with the calendar OWNER's SocialAccount, not the requesting admin's. An admin can trigger the sync but the actual provider call runs on the owner's credentials — the admin never needs (and never uses) the target user's OAuth token. The happy-path test asserts the account passed to authenticate() is the owner's, not the admin's.

calendar_integration/views.py
```diff
@@ -259,6 +260,97 @@ def request_sync(
...
+    )
+    @action(
+        methods=["post"],
+        detail=True,
+        url_path="admin-sync",
+        url_name="admin-sync",
+        permission_classes=[IsOrganizationAdmin],
+    )
+    @inject
+    def admin_sync(
```
hugobessa, line 272: `permission_classes=[IsOrganizationAdmin]` on the action overrides the viewset's default calendar permission for admin-sync only: has_permission requires an active admin (non-admin member → 403) and has_object_permission + the org-scoped get_queryset together make a cross-org calendar a 404. The regular self-sync action (Phase 5) keeps the owner-only rule; this one is the admin escalation path.

calendar_integration/views.py
```diff
@@ -259,6 +260,97 @@ def request_sync(
...
+            .first()
+        )
+
+        if not ownership:
+            return Response(
+                {"detail": "Calendar has no owner; cannot sync."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        calendar_owner = ownership.user
```
hugobessa, line 307: Owner resolution orders by `-is_default, id` so a calendar's default owner is chosen when several users co-own it, falling back to the earliest ownership. No owner at all (e.g. an orphaned/resource calendar) → 400 rather than a confusing auth failure downstream.

PR #49, opened 2026-06-05
feat(organizations): add rooms-sync trigger, on-enable transition, and admin org update

hugobessa:
## Summary
Phase 7 of the **REST API Frontend Gaps** plan. Delivers the admin "sync rooms" capability and — necessarily — makes org configuration actually editable by admins:
- `POST /organizations/{id}/sync-rooms/` (admin) — enqueue an organization calendar-resources (rooms) import. Optional `start_time`/`end_time` (ISO; default now..+365d).
- `should_sync_rooms` False→True transition on org update auto-triggers the import exactly once.
- `OrganizationService.request_rooms_sync(...)` — shared method; `create_organization` now delegates to it (DRY).

## Why the permission change was required
The org `update`/`partial_update` endpoint was previously **unreachable for everyone** (`OrganizationManagementPermission.has_permission` only admits membership-less users, while `has_object_permission` requires a membership — a contradiction). That made both the should_sync_rooms transition AND the original requirement *"an admin must be able to configure their org to sync rooms"* impossible. This phase adds a `get_permissions()` override gating `update`/`partial_update` with `IsOrganizationAdmin` (create/retrieve/destroy unchanged), so admins can configure their own org. Members still get 403; cross-org 404.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 7 + API Design 4.3 + the "configure org to sync rooms" use-case.

## Review history
Two review passes. First surfaced that the transition lived on an unreachable endpoint and its tests only "passed" by patching the permission away — fixed by opening admin update and rewriting the tests with real admin clients. Second pass hardened the update path: removed a dead `try/except` around `on_commit` (it can't catch async service errors), corrected the OpenAPI for `sync_rooms`, and added a `select_for_update` lock so concurrent False→True PATCHes can't double-fire the import.

## Out-of-scope note
The underlying `request_organization_calendar_resources_import` enqueues `.delay` without `on_commit` internally (pre-existing). The new reachable paths (sync_rooms action + transition) defer via `on_commit` at the view layer; `create_organization`'s timing is unchanged.

## Test plan
- Asserts: admin sync-rooms 202 (explicit + default window); non-admin 403; cross-org 404; anon 401; bad datetime 400. Transition: False→True fires once; already-True / True→False / unchanged fire zero; non-admin 403 fires zero. Admin can PATCH should_sync_rooms and it persists. create_organization(should_sync_rooms=True) still triggers.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1320 passed).

### DIFF
--- a/organizations/services.py
+++ b/organizations/services.py
@@ -65,15 +65,36 @@ class OrganizationService:
         )
 
         if should_sync_rooms:
-            self.calendar_service.initialize_without_provider(
-                user_or_token=creator, organization=self.organization
-            )
-            now = datetime.datetime.now(tz=datetime.UTC)
-            self.calendar_service.request_organization_calendar_resources_import(
-                start_time=now,
-                end_time=now + datetime.timedelta(days=365),
+            self.request_rooms_sync(
+                organization=self.organization,
+                requested_by=creator,
             )
         return self.organization
+
+    def request_rooms_sync(
+        self,
+        organization: Organization,
+        requested_by: User,
+        start_time: datetime.datetime | None = None,
+        end_time: datetime.datetime | None = None,
+    ) -> None:
+        """
+        Initialize the calendar service without a provider and enqueue a
+        calendar resources import for the given organization.
+
+        :param organization: The organization to sync rooms for.
+        :param requested_by: The user (or token) authorizing the sync.
+        :param start_time: Import window start; defaults to now.
+        :param end_time: Import window end; defaults to now + 365 days.
+        """
+        self.calendar_service.initialize_without_provider(
+            user_or_token=requested_by, organization=organization
+        )
+        now = datetime.datetime.now(tz=datetime.UTC)
+        self.calendar_service.request_organization_calendar_resources_import(
+            start_time=start_time or now,
+            end_time=end_time or (now + datetime.timedelta(days=365)),
+        )
 
     @transaction.atomic()
     def invite_user_to_organization(
--- a/organizations/views.py
+++ b/organizations/views.py
@@ -85,20 +85,28 @@ class OrganizationViewSet(NoListVintaScheduleModelViewSet):
         return Organization.objects.none()
 
     def update(self, request, *args, **kwargs):
-        """Override update to trigger rooms sync when should_sync_rooms flips False→True."""
+        """Override update to trigger rooms sync when should_sync_rooms flips False→True.
+
+        Uses select_for_update to lock the row during snapshot + write, serializing
+        concurrent PATCHes and preventing double-fire of the sync on False→True transition.
+        """
         partial = kwargs.get("partial", False)
-        instance = self.get_object()
 
-        # Snapshot the old value BEFORE the serializer writes to the DB.
-        old_should_sync_rooms = instance.should_sync_rooms
+        # Lock the row during snapshot + write.
+        with transaction.atomic():
+            instance = Organization.objects.select_for_update().get(pk=self.get_object().pk)
+            old_should_sync_rooms = instance.should_sync_rooms
 
-        serializer = self.get_update_serializer(instance, data=request.data, partial=partial)
-        serializer.is_valid(raise_exception=True)
-        self.perform_update(serializer)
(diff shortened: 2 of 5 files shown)


organizations/views.py
```diff
@@ -50,13 +53,74 @@ class OrganizationViewSet(NoListVintaScheduleModelViewSet):
...
+        - update / partial_update: admin-only (IsOrganizationAdmin).  An admin
+          can only reach their own org because get_queryset is scoped by
+          membership, so cross-org attempts return 404.
+        - All other actions keep the class-level defaults (IsAuthenticated +
+          OrganizationManagementPermission), which gates onboarding (create)
+          to membership-less users and locks down retrieve/destroy.
+        """
+        if self.action in ("update", "partial_update"):
+            return [IsAuthenticated(), IsOrganizationAdmin()]
+        return super().get_permissions()
```
hugobessa, line 68: This get_permissions override is what actually unlocks the phase. The org update path was previously unreachable for everyone (OrganizationManagement- Permission.has_permission only admits membership-LESS users, but has_object_permission requires a membership — a contradiction). Gating update/partial_update with IsOrganizationAdmin lets an admin configure their OWN org (get_queryset scopes to their org → cross-org 404; non-admin → 403), which is required both for the should_sync_rooms transition below and for the "admin configures org to sync rooms" use-case. create/retrieve/destroy keep their existing (locked) behavior.

organizations/views.py
```diff
@@ -50,13 +53,74 @@ class OrganizationViewSet(NoListVintaScheduleModelViewSet):
...
+        # Register the on_commit callback AFTER the atomic block (so it fires post-commit).
+        if fire:
+            org = serializer.instance
+            user = request.user
+            org_service = self.organization_service
+
+            def _trigger():
+                org_service.request_rooms_sync(organization=org, requested_by=user)
+
+            transaction.on_commit(_trigger)
```
hugobessa, line 117: Exactly-once transition with a row lock: select_for_update on the org during the snapshot-of-old-value + write serializes concurrent admin PATCHes, so two simultaneous False→True updates can't both read old=False and double-enqueue the import. The on_commit registration is deliberately placed AFTER the atomic block so the Celery task fires only once the write has committed (ATOMIC_- REQUESTS=True). get_object() still runs first for object-permission + org scoping; the select_for_update re-fetch by pk only adds the lock.

organizations/services.py
```diff
@@ -65,16 +65,37 @@ def create_organization(
...
+        :param start_time: Import window start; defaults to now.
+        :param end_time: Import window end; defaults to now + 365 days.
+        """
+        self.calendar_service.initialize_without_provider(
+            user_or_token=requested_by, organization=organization
+        )
+        now = datetime.datetime.now(tz=datetime.UTC)
+        self.calendar_service.request_organization_calendar_resources_import(
+            start_time=start_time or now,
+            end_time=end_time or (now + datetime.timedelta(days=365)),
```
hugobessa, line 96: request_rooms_sync centralizes the "initialize service without a provider + request resources import" setup that create_organization used inline. Both create_organization and the two new view paths (sync_rooms action + transition) call it, so the rooms-import wiring lives in one place. Defaults preserve the original now..+365d window.

organizations/views.py
```diff
@@ -77,6 +141,64 @@ def current(self, request):
...
+
+        def _trigger():
+            org_service.request_rooms_sync(
+                organization=org,
+                requested_by=user,
+                start_time=start_time,
+                end_time=end_time,
+            )
+
+        transaction.on_commit(_trigger)
```
hugobessa, line 197: sync_rooms is admin-only (permission_classes=[IsOrganizationAdmin] overrides the viewset default for this action; anonymous → 401 since the permission rejects unauthenticated users and authenticators are configured). Bad ISO datetimes are rejected as 400 BEFORE the enqueue; the enqueue itself is deferred via on_commit (no try/except around it — on_commit can't surface an async service error, so wrapping it would be dead code).

PR #50, opened 2026-06-05
feat(calendar): add admin event transfer action

hugobessa:
## Summary
Phase 8 of the **REST API Frontend Gaps** plan. Adds `POST /calendar-events/{id}/transfer/` (admin-only): moves an event from one calendar to another within the org, via `CalendarService.transfer_event`. Returns the new event (200).

## Behavior
- Admin-only (`IsOrganizationAdmin`); event org-scoped (cross-org → 404); non-admin → 403.
- `target_calendar_id` required, must resolve within the event's organization (else 400); same-calendar transfer is a 400 no-op.
- Authenticates with the **source** calendar owner's account (transfer reads the source event from its provider) — not the admin's. No owner / no linked account → 400.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 8 + API Design 4.4.

## Review history
Layer-3 review: no blockers. Added the missing test for the same-calendar no-op guard. Noted (not enforced) that the target isn't restricted to PERSONAL calendars — left open since an admin may legitimately transfer to a virtual/bundle calendar.

## Test plan
- `uv run pytest calendar_integration/tests/test_views.py -k transfer -vs`
- Asserts: admin transfer success (transfer_event called with event+target, authenticated with the SOURCE owner's account); non-admin 403; cross-org event 404; missing/invalid/cross-org target 400; same-calendar no-op 400 (service not called); no source owner/account 400; anon 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1329 passed).

### DIFF
--- a/calendar_integration/views.py
+++ b/calendar_integration/views.py
@@ -661,6 +661,116 @@ class CalendarEventViewSet(VintaScheduleModelViewSet):
         except (ValueError, CalendarIntegrationError) as e:
             raise ValidationError({"non_field_errors": [str(e)]}) from e
 
+    @extend_schema(
+        summary="Transfer event to another calendar (admin)",
+        description=(
+            "Move an event from its current calendar to a target calendar within the same "
+            "organization. The service authenticates with the SOURCE calendar owner's credentials "
+            "to read and delete the event from the provider. Admin only."
+        ),
+        request={"type": "object", "properties": {"target_calendar_id": {"type": "integer"}}},
+        responses={200: CalendarEventSerializer},
+    )
+    @action(
+        detail=True,
+        methods=["post"],
+        url_path="transfer",
+        url_name="transfer",
+        permission_classes=[IsOrganizationAdmin],
+    )
+    @inject
+    def transfer(
+        self,
+        request,
+        pk: str | None = None,
+        calendar_service: Annotated[CalendarService, Provide["calendar_service"]] = None,  # type: ignore
+    ) -> Response:
+        """Transfer an event to a different calendar (admin-only)."""
+        event = self.get_object()  # org-scoped → cross-org yields 404; non-admin → 403
+
+        # --- Parse and validate body ---
+        target_calendar_id = request.data.get("target_calendar_id")
+        if target_calendar_id is None:
+            return Response(
+                {"detail": "target_calendar_id is required."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        # Reject no-op transfers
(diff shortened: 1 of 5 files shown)


calendar_integration/views.py
```diff
@@ -666,6 +666,116 @@ def bulk_modify(
...
+                {"detail": "User is not an active member of any organization."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        try:
+            calendar_service.authenticate(
+                account=owner_social_account,
+                organization=membership.organization,
+            )
+            new_event = calendar_service.transfer_event(
```
hugobessa, line 744: Security-critical, same shape as admin-sync: transfer_event reads the source event from its provider (calendar_adapter.get_event), so the service is authenticated with the SOURCE calendar OWNER's account, never the admin's. The admin authorizes the move; the provider calls run on the owner's credentials. The success test asserts authenticate() got the owner's account.

calendar_integration/views.py
```diff
@@ -666,6 +666,116 @@ def bulk_modify(
...
+        # Resolve the target calendar within the event's organization
+        try:
+            target_calendar = Calendar.objects.filter_by_organization(event.organization_id).get(
+                id=target_calendar_id
+            )
+        except Calendar.DoesNotExist:
+            return Response(
+                {"detail": "target_calendar_id invalid or not in your organization."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
hugobessa: Target is resolved with filter_by_organization(event.organization_id) — the tenant-safe manager, so a cross-org calendar id can't be targeted (→ 400, not a leak). Using the event's org (rather than the admin's) makes the same-org invariant explicit even though get_object already guarantees they match.

calendar_integration/views.py
```diff
@@ -666,6 +666,116 @@ def bulk_modify(
...
+
+        # Reject no-op transfers
+        if str(target_calendar_id) == str(event.calendar_fk_id):
+            return Response(
+                {"detail": "Event is already on the target calendar."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
```
hugobessa: Same-calendar transfers short-circuit to 400 before any provider round-trip or owner-account lookup — a transfer to the event's current calendar is a no-op, not an error worth hitting Google/Microsoft for. Covered by test_transfer_event_same_calendar_no_op (asserts the service is never called).

PR #51, opened 2026-06-05
feat(calendar): add calendar soft-disable (is_active)

hugobessa:
## Summary
Phase 9 of the **REST API Frontend Gaps** plan — "disable a calendar of their own". Adds `Calendar.is_active`; `DELETE /calendar/{id}/` now **soft-disables** (sets is_active=False, row persists) instead of hard-deleting. Disabled calendars are hidden from default list/detail; `?include_inactive=true` surfaces them.

## What changed
- `Calendar.is_active` BooleanField (default True, db_default True, db_index) + additive migration `0011_calendar_is_active`.
- `CalendarViewSet.get_queryset` filters is_active=True unless `?include_inactive=true` (applied uniformly — so admins can still reach a disabled calendar on action routes by passing the flag).
- `destroy` override → is_active=False, 204, idempotent (no hard delete; bundle cascade deferred to a later phase).
- Serializer exposes is_active (read-only).

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 9 + Data Model 3.2 + Guiding Decision (Calendar soft-disable).

## Review history
Layer-3 review: no blockers. Clarified + tested that the default-active filter is opt-out via `?include_inactive=true` uniformly (admins can still act on a disabled calendar on action routes with the flag; 404 without). Removed an internal "Phase 11" phrase from the public API description.

## Follow-up note (for the bundle phases)
Some serializer querysets (bundle-create, resource-allocation) and the public GraphQL calendars query resolve calendars via `filter_by_organization` WITHOUT an is_active filter, so a disabled calendar could still be selected/listed there. Acceptable for this phase (soft-disable only hides from CalendarViewSet defaults); worth revisiting when Phase 10/11 touch bundles.

## Test plan
- Asserts: DELETE → 204 + row persists + is_active False; default list hides disabled; `?include_inactive=true` shows it; retrieve-of-disabled 404 by default / 200 with flag; create defaults active; admin action reaches disabled calendar only with the flag; org-scoping intact.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1338 passed).


calendar_integration/migrations/0011_calendar_is_active.py
```diff
@@ -0,0 +1,26 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="calendar",
+            name="is_active",
+            field=models.BooleanField(
+                db_default=True,
+                db_index=True,
+                default=True,
+                help_text="Whether this calendar is active. Inactive calendars are hidden from default list/detail queries. Use DELETE /calendar/{id}/ to soft-disable (sets this to False) instead of deleting the row. Opt-in to see disabled calendars via ?include_inactive=true. Default True keeps every existing read unchanged.",
+            ),
```
hugobessa, line 24: Additive, single-step column: db_default=True backfills existing rows at the DB level (no data migration, no table rewrite) and default=True keeps every pre-feature read unchanged. Calendar is non-partitioned, so no CONCURRENTLY dance needed. Mirrors the OrganizationMembership.is_active migration from Phase 1.

calendar_integration/views.py
```diff
@@ -85,7 +89,31 @@ def get_queryset(self):
...
             # Membership-less or inactive members get an empty queryset, not a 500.
             return Calendar.original_manager.none()
-        return super().get_queryset().filter_by_organization(membership.organization_id)
+
+        qs = super().get_queryset().filter_by_organization(membership.organization_id)
+
+        include_inactive = self.request.query_params.get("include_inactive", "").lower() == "true"
+        if not include_inactive:
+            qs = qs.filter(is_active=True)
+
```
hugobessa, line 100: The default-active filter is applied AFTER org-scoping and reads include_inactive from query params uniformly for every action — so the same flag that surfaces disabled rows in the list also lets an admin reach a disabled calendar on an action route (e.g. admin-sync?include_inactive=true). Without the flag, disabled calendars 404 everywhere. This keeps "disabled = hidden by default" consistent rather than special-casing per action.

calendar_integration/views.py
```diff
@@ -85,7 +89,31 @@ def get_queryset(self):
...
+            "Bundle calendars may additionally cascade to their child representations."
+        ),
+        responses={204: None},
+    )
+    def destroy(self, request, *args, **kwargs):
+        """Soft-disable the calendar (set is_active=False) instead of hard-deleting."""
+        calendar = self.get_object()
+        calendar.is_active = False
+        calendar.save(update_fields=["is_active"])
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
hugobessa, line 118: DELETE is soft-disable: set is_active=False, 204, row persists. Idempotent (a second DELETE 404s because the row is now hidden). Kept deliberately generic so the bundle phase can extend it to cascade to a bundle's child representations without rewriting the base behavior.

PR #52, opened 2026-06-05
feat(calendar): add bundle update action

hugobessa:
## Summary
Phase 10 of the **REST API Frontend Gaps** plan. Adds `PATCH /calendar/{id}/bundle/` (admin-only): reconcile a bundle calendar's child calendars and primary designation. Complements the existing `POST /calendar/bundle/` create.

## What changed
- `CalendarService.update_bundle_calendar(bundle, child_calendars, primary_calendar)` — atomic reconciliation of `ChildrenCalendarRelationship` rows (add new, remove dropped, exactly one primary). Validates bundle type, primary∈children, same-org, no nested bundles.
- `CalendarBundleUpdateSerializer` — org-scoped child/primary fields; `update_bundle` PATCH action gated by `IsOrganizationAdmin`; non-bundle target → 400.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 10 + API Design 4.6.

## Review history
Layer-3 review: production-ready, no blockers. Fixed one UX trap it flagged — a bundle child disabled after creation could no longer be resubmitted (the active-only queryset rejected it, blocking all bundle edits). The serializer queryset now allows the bundle's existing children regardless of is_active, while still barring brand-new disabled children. Two tests added for both sides.

## Test plan
- `uv run pytest calendar_integration/tests/test_views.py -k bundle -vs`
- Asserts: add/remove child + change primary → 200 with exactly one primary; non-bundle 400; <2 children 400; primary-not-in-children 400; cross-org child 400; NEW disabled child rejected 400; EXISTING disabled child kept on update 200; non-admin 403; cross-org bundle 404; anon 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1349 passed).

### DIFF
--- a/calendar_integration/serializers.py
+++ b/calendar_integration/serializers.py
@@ -23,6 +23,7 @@ from calendar_integration.models import (
     CalendarGroupSlotMembership,
     CalendarOwnership,
     CalendarSync,
+    ChildrenCalendarRelationship,
     EventAttendance,
     EventExternalAttendance,
     EventRecurrenceException,
@@ -261,19 +262,39 @@ class CalendarBundleUpdateSerializer(serializers.Serializer):
         active_membership = (
             get_active_organization_membership(user) if user and user.is_authenticated else None
         )
-        active_org_qs = (
-            Calendar.objects.filter_by_organization(
-                organization_id=active_membership.organization_id
-            ).filter(is_active=True)
-            if active_membership
-            else Calendar.original_manager.none()
-        )
+
+        # Build queryset: active calendars + existing children (even if disabled)
+        org_id = active_membership.organization_id if active_membership else None
+
+        if org_id:
+            from django.db.models import Q
+
+            # Get the target bundle from context (passed from the view)
+            bundle = self.context.get("bundle") if self.context else None
+
+            # Fetch existing child IDs
+            existing_child_ids = []
+            if bundle:
+                existing_child_ids = list(
+                    ChildrenCalendarRelationship.objects.filter(
+                        bundle_calendar=bundle,
+                        organization_id=org_id,
+                    ).values_list("child_calendar_fk_id", flat=True)
+                )
+
+            # Build the queryset: (active OR in existing_child_ids) AND org-scoped
+            qs = Calendar.objects.filter_by_organization(organization_id=org_id).filter(
+                Q(is_active=True) | Q(id__in=existing_child_ids)
+            )
+        else:
+            qs = Calendar.original_manager.none()
+
         self.fields["bundle_calendars"] = serializers.PrimaryKeyRelatedField(
             many=True,
-            queryset=active_org_qs,
+            queryset=qs,
         )
         self.fields["primary_calendar"] = serializers.PrimaryKeyRelatedField(
-            queryset=active_org_qs,
+            queryset=qs,
             allow_null=True,
             required=False,
         )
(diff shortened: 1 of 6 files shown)


calendar_integration/services/calendar_service.py
```diff
@@ -736,6 +736,91 @@ def create_bundle_calendar(
...
+        :param primary_calendar: The child to designate as primary; must be in child_calendars.
+        :return: The (unchanged) bundle_calendar instance after reconciliation.
+        :raises ValueError: If bundle_calendar is not a BUNDLE type, children are cross-org,
+                            or primary_calendar is not in child_calendars.
+        """
+        if not is_initialized_or_authenticated_calendar_service(self):
+            raise
+
+        if bundle_calendar.calendar_type != CalendarType.BUNDLE:
+            raise ValueError("Calendar is not a bundle.")
```
hugobessa, line 764: Reconciliation lives in the service (project convention: business logic out of views) and runs in @transaction.atomic so a partial failure can't leave a half-rewired bundle. It validates bundle type, primary∈children, same-org for every child, and rejects nesting a bundle inside a bundle — then diffs existing vs desired children (add/remove) and forces exactly one is_primary when a primary is given.

calendar_integration/serializers.py
```diff
@@ -245,6 +246,73 @@ def create(self, validated_data):
...
+                        bundle_calendar=bundle,
+                        organization_id=org_id,
+                    ).values_list("child_calendar_fk_id", flat=True)
+                )
+
+            # Build the queryset: (active OR in existing_child_ids) AND org-scoped
+            qs = Calendar.objects.filter_by_organization(organization_id=org_id).filter(
+                Q(is_active=True) | Q(id__in=existing_child_ids)
+            )
+        else:
```
hugobessa, line 300: The child queryset is (active org calendars) UNION (this bundle's existing children, regardless of is_active). This fixes a trap a reviewer caught: once calendar soft-disable (Phase 9) exists, a bundle child could later be disabled, and an active-only queryset would reject re-submitting the current set — locking the admin out of editing the bundle at all. New disabled calendars still can't be added (a disabled calendar that isn't already a child won't be in the queryset); only existing ones can be kept.

calendar_integration/views.py
```diff
@@ -140,6 +142,75 @@ def create_bundle_calendar(self, request, *args, **kwargs):
...
+        calendar_service: Annotated[CalendarService, Provide["calendar_service"]] = None,  # type: ignore
+    ) -> Response:
+        """Update the children and primary calendar of an existing bundle (admin only)."""
+        calendar = self.get_object()
+
+        if calendar.calendar_type != CalendarType.BUNDLE:
+            return Response(
+                {"detail": "Calendar is not a bundle."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
hugobessa: Admin-only (IsOrganizationAdmin); the target is org-scoped via get_object (cross-org → 404) and must be a BUNDLE calendar (else 400). The action passes the bundle into the serializer context so the serializer can union in the existing children (see the serializer comment). PATCH at detail url_path "bundle" → /calendar/{id}/bundle/, distinct from the create action's collection POST /calendar/bundle/.

PR #54, opened 2026-06-05
feat(public_api): add admin REST endpoint to create API tokens

hugobessa:
## Summary
Phase 12 of the **REST API Frontend Gaps** plan. Adds the first private-REST surface to the `public_api` app (previously GraphQL-only): `POST /public-api-tokens/` (admin-only) creates a `SystemUser` + its `ResourceAccess` grants and returns the plaintext token **once**.

## What changed
- `SystemUserTokenViewSet` (create-only, `IsOrganizationAdmin`), org-scoped to the caller.
- `SystemUserTokenCreateSerializer` (integration_name + available_resources∈PublicAPIResources, non-empty) and `SystemUserTokenResponseSerializer` (id, integration_name, is_active, available_resources, write-once token — never the hash).
- New `public_api/routes.py` wired into `vinta_schedule_api/urls.py`.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 12 + API Design 4.7 + Guiding Decision (token surface = private REST admin-only). Create-only; list/revoke/edit are Phases 13–15.

## Review history
Layer-3 review found a **production blocker**: catching `IntegrityError` (duplicate integration_name) without a savepoint poisons the request transaction under `ATOMIC_REQUESTS=True`, turning the intended 400 into a 500 (the test only passed because test settings don't enable ATOMIC_REQUESTS). Fixed by wrapping the create in a nested `transaction.atomic()` savepoint. Also corrected the OpenAPI 201 schema (was the input serializer — token field undocumented), removed a dead validator + dead TYPE_CHECKING block, guarded get_queryset for anon, and added id/required-field tests.

## Test plan
- Asserts: admin create → 201 + plaintext token + id/is_active/available_resources, ResourceAccess rows created, hash never in response, SystemUser org-scoped; invalid/empty resources → 400; missing integration_name → 400; duplicate integration_name → 400 (not 500); non-admin → 403; anon → 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1376 passed).

### DIFF
--- a/public_api/serializers.py
+++ b/public_api/serializers.py
@@ -1,15 +1,24 @@
-from rest_framework import serializers
+from typing import Annotated
 
+from django.db import IntegrityError, transaction
+
+from dependency_injector.wiring import Provide, inject
+from rest_framework import serializers
+from rest_framework.exceptions import PermissionDenied
+
+from organizations.models import get_active_organization_membership
 from public_api.constants import PublicAPIResources
 from public_api.models import ResourceAccess, SystemUser
+from public_api.services import PublicAPIAuthService
 
 
 class SystemUserTokenCreateSerializer(serializers.Serializer):
     """Input serializer for creating a new public-API token (SystemUser + ResourceAccess rows).
 
     Accepts ``integration_name`` and ``available_resources`` (a non-empty list of valid
-    ``PublicAPIResources`` values).  On a successful create the view adds a write-once
-    ``token`` field to the response data; this serializer never stores or exposes
+    ``PublicAPIResources`` values).  ``create()`` provisions the ``SystemUser`` via the
+    injected auth service and bulk-creates the ``ResourceAccess`` grants, attaching the
+    write-once plaintext ``token`` to the returned instance.  Never stores or exposes
     ``long_lived_token_hash``.
     """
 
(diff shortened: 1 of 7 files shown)


public_api/views.py
```diff
@@ -0,0 +1,119 @@
...
+            with transaction.atomic():
+                system_user, plaintext_token = self.public_api_auth_service.create_system_user(
+                    integration_name=integration_name,
+                    organization=organization,
+                )
+        except IntegrityError:
+            return Response(
+                {
+                    "integration_name": [
+                        f"A token with integration_name '{integration_name}' already exists."
```
hugobessa: The nested savepoint is the production-correctness fix. Under ATOMIC_REQUESTS=True the whole request is one transaction; a duplicate integration_name raises IntegrityError that poisons it, so catching the error without a savepoint would make the final commit raise TransactionManagementError (a 500) instead of the intended 400. Wrapping only create_system_user in transaction.atomic() rolls back just that savepoint and lets the 400 return cleanly. (The original test passed only because test settings don't set ATOMIC_REQUESTS — a real gap Layer-3 review caught.)

public_api/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+    Never exposes ``long_lived_token_hash``.
+    """
+
+    available_resources = serializers.SerializerMethodField()
+    token = serializers.CharField(read_only=True)
+
+    class Meta:
+        model = SystemUser
+        fields = ("id", "integration_name", "is_active", "available_resources", "token")
+        read_only_fields = fields
```
hugobessa, line 94: The response serializer enumerates fields explicitly and never includes long_lived_token_hash — so the secret hash can't leak even though it lives on the model. The plaintext `token` is attached as a transient attribute on the instance only in the create response (write-once); no list/retrieve endpoint exists this phase, so the token is unrecoverable afterward.

public_api/views.py
```diff
@@ -0,0 +1,119 @@
...
+    """Admin-only viewset for creating public-API tokens (SystemUser + ResourceAccess rows).
+
+    Phase 12: create only.  List / retrieve / revoke / edit-grants are future phases.
+
+    ``POST /public-api-tokens/`` creates a new ``SystemUser`` for the caller's
+    organisation, persists the requested ``ResourceAccess`` rows, and returns the
+    plaintext token **once**.  The token is never recoverable after this response.
+    """
+
+    permission_classes = (IsOrganizationAdmin,)
```
hugobessa, line 31: Admin-only create; the SystemUser's organization comes from the caller's membership (never a request field), so an admin can only mint tokens for their own org. This is the private management surface for the GraphQL API tokens — distinct from the public GraphQL endpoint those tokens authenticate against.

PR #55, opened 2026-06-05
feat(public_api): add admin list/retrieve for API tokens

hugobessa:
## Summary
Phase 13 of the **REST API Frontend Gaps** plan. Adds `GET /public-api-tokens/` and `GET /public-api-tokens/{id}/` (admin-only) so admins can view their org's tokens — without any secret. Read side of the token-management UI.

## What changed
- `SystemUserTokenViewSet` now mixes in list + retrieve; `get_serializer_class` routes per action (create → input/response serializers with the write-once token; list/retrieve → `SystemUserTokenSerializer`).
- `SystemUserTokenSerializer` exposes only id/integration_name/is_active/available_resources — never token or long_lived_token_hash.
- `get_queryset` prefetches `available_resources` to keep the list to a bounded query count.
- `IsOrganizationAdmin.has_object_permission` gains a `hasattr(obj, "organization_id")` fallback so it can object-gate SystemUser (still requires org match + admin).

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 13 + API Design 4.7.

## Review history
Layer-3 review: clean, no blockers. Confirmed no secret leak on list/retrieve, org-scoping (cross-org → excluded/404), the permission fallback stays safe (org-match + admin still required and pre-existing isinstance consumers are unaffected), and the prefetch removes the N+1.

## Test plan
- Asserts: admin list 200 (active+inactive, no token/hash); retrieve 200 (no secrets); cross-org excluded + retrieve 404; non-admin 403; anon 401; create still returns the write-once token.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1393 passed).

### DIFF
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -113,7 +113,11 @@ class IsOrganizationAdmin(BasePermission):
         elif isinstance(obj, OrganizationModel):
             obj_organization_id = obj.organization_id
         else:
-            return False
+            # Handle SystemUser and other objects with an organization FK
+            if hasattr(obj, "organization_id"):
+                obj_organization_id = obj.organization_id
+            else:
+                return False
 
         # Membership org must match object org; user must be an admin
         if membership.organization_id != obj_organization_id:
--- a/public_api/views.py
+++ b/public_api/views.py
@@ -2,6 +2,7 @@ import logging
 
 from drf_spectacular.utils import extend_schema
 from rest_framework import status
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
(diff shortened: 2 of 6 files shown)


public_api/views.py
```diff
@@ -49,15 +55,27 @@ def __init__(
...
+            return SystemUser.objects.filter(
+                organization_id=membership.organization_id
+            ).prefetch_related("available_resources")
         return SystemUser.objects.none()
 
+    def get_serializer_class(self):  # type: ignore[override]
+        """Use SystemUserTokenSerializer for list/retrieve; create-response uses SystemUserTokenResponseSerializer."""
+        if self.action == "create":
+            return SystemUserTokenCreateSerializer
+        return SystemUserTokenSerializer
```
hugobessa, line 60: Per-action serializer routing is what keeps the secret write-once: only the create action serializes the plaintext token (via the response serializer it instantiates directly); list/retrieve use SystemUserTokenSerializer, which omits both `token` and `long_lived_token_hash` from its fields entirely, so a secret can't be read back after creation.

public_api/views.py
```diff
@@ -49,15 +55,27 @@ def __init__(
...
+        """
         user = self.request.user
         if not user.is_authenticated:
             return SystemUser.objects.none()
         membership = get_active_organization_membership(user)
         if membership:
-            return SystemUser.objects.filter(organization_id=membership.organization_id)
+            return SystemUser.objects.filter(
+                organization_id=membership.organization_id
+            ).prefetch_related("available_resources")
```
hugobessa, line 53: prefetch_related("available_resources") (the related_name on ResourceAccess.system_user) plus the serializer reading obj.available_resources.all() keeps the list to a bounded query count — the create-response serializer's per-instance ResourceAccess query would have been an N+1 across a token list.

organizations/permissions.py
```diff
@@ -113,7 +113,11 @@ def has_object_permission(self, request, view, obj) -> bool:
...
             obj_organization_id = obj.organization_id
         else:
-            return False
+            # Handle SystemUser and other objects with an organization FK
+            if hasattr(obj, "organization_id"):
+                obj_organization_id = obj.organization_id
+            else:
+                return False
 
         # Membership org must match object org; user must be an admin
```
hugobessa, line 122: The hasattr(organization_id) fallback lets this permission object-gate SystemUser (which carries an organization FK but isn't an OrganizationModel subclass) without weakening anything: it still requires membership.organization_id == obj.organization_id AND user.is_organization_admin(...). Existing consumers (Organization, OrganizationMembership, Calendar) keep hitting their explicit isinstance branches above, so their behavior is unchanged.

PR #56, opened 2026-06-05
feat(public_api): add admin token revoke action

hugobessa:
## Summary
Phase 14 of the **REST API Frontend Gaps** plan. Adds `POST /public-api-tokens/{id}/revoke/` (admin-only): invalidates a token by setting its `SystemUser.is_active=False`. Revoked tokens then fail `check_system_user_token` (which already returns `(user, False)` for inactive users), so the GraphQL API rejects them.

## Behavior
- Admin-only detail action; cross-org token → 404; non-admin → 403; anon → 401.
- Idempotent: revoking an already-revoked token is a 200 no-op.
- Returns the no-secret serializer (is_active=False); never the token/hash.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 14 + API Design 4.7.

## Review history
Layer-3 review: clean, no blockers. The key end-to-end test creates a token, confirms `check_system_user_token` is True, revokes via the endpoint, then asserts it returns False.

## Test plan
- Asserts: admin revoke → 200 + is_active False + auth then rejected; idempotent re-revoke 200; cross-org 404; non-admin 403; anon 401; no secret in response.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1403 passed).

### DIFF
--- a/public_api/views.py
+++ b/public_api/views.py
@@ -2,6 +2,7 @@ import logging
 
 from drf_spectacular.utils import extend_schema
 from rest_framework import status
+from rest_framework.decorators import action
 from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
 from rest_framework.request import Request
 from rest_framework.response import Response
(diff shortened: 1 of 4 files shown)


public_api/views.py
```diff
@@ -135,3 +136,22 @@ def create(self, request: Request, *args, **kwargs) -> Response:
...
+        This is idempotent: revoking an already-revoked token is a 200 no-op.
+
+        Returns HTTP 200 with the updated token serialized via SystemUserTokenSerializer.
+        HTTP 403 is returned for non-admin callers; HTTP 404 if the token does not
+        exist or belongs to another organization; HTTP 401 for unauthenticated.
+        """
+        system_user = self.get_object()
+        system_user.is_active = False
+        system_user.save(update_fields=["is_active"])
+
```
hugobessa, line 105: Revoke leans on existing auth logic: check_system_user_token already returns (user, False) when the SystemUser is inactive, so flipping is_active=False is the whole invalidation — no token rotation or row deletion needed, and it's reversible if needed later. The end-to-end test proves a previously-valid token verifies False after this call.

public_api/views.py
```diff
@@ -135,3 +136,22 @@ def create(self, request: Request, *args, **kwargs) -> Response:
             system_user, context={"request": request}
         )
         return Response(response_serializer.data, status=status.HTTP_201_CREATED)
+
+    @extend_schema(responses={200: SystemUserTokenSerializer})
+    @action(detail=True, methods=["post"], url_path="revoke")
+    def revoke(self, request: Request, pk=None) -> Response:
```
hugobessa, line 92: Detail action under the viewset's IsOrganizationAdmin: has_object_permission enforces admin + org-match and the org-scoped get_queryset makes a cross-org token a 404. The 200 response uses the no-secret serializer, so revoke (like list/retrieve) never echoes the token or its hash.

PR #57, opened 2026-06-05
feat(public_api): add admin token permission edit (PATCH)

hugobessa:
## Summary
Phase 15 of the **REST API Frontend Gaps** plan. Adds `PATCH/PUT /public-api-tokens/{id}/` (admin-only): edit a token's resource grants (`ResourceAccess`) without re-issuing the secret. Completes the token lifecycle (create → list → revoke → edit).

## What changed
- `SystemUserTokenUpdateSerializer` — accepts `available_resources` ONLY (validated against PublicAPIResources, non-empty). integration_name / token / is_active are never accepted or changed.
- `update`/`partial_update` on `SystemUserTokenViewSet` reconcile ResourceAccess transactionally (add new, remove dropped via set-diff); response is the no-secret serializer.
- Data-integrity hardening: added a `UniqueConstraint(system_user, resource_name)` on `ResourceAccess` (migration `0003`) and deduped the Phase 12 create input so repeated resources can't create duplicate grants.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Phase 15 + API Design 4.7. (Resolves the earlier "create-only" open question — grants are now editable.)

## Review history
Layer-3 review: no blockers. Acted on its data-integrity finding — `ResourceAccess` had no uniqueness guard, so duplicate grants could persist (and the create loop could mint them from duplicate input). Added the unique constraint + create-time dedupe, plus a test.

## Test plan
- Asserts: admin PATCH/PUT → 200, grants reconciled to exactly the new set (no dupes); long_lived_token_hash + integration_name + is_active unchanged (even if sent in body); invalid/empty resources → 400; non-admin 403; cross-org 404; anon 401; create dedupes duplicate input.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1426 passed).


public_api/views.py
```diff
@@ -155,3 +163,79 @@ def revoke(self, request: Request, pk=None) -> Response:
...
+
+            # Remove dropped resources
+            removed_resources = current_resources - desired_set
+            if removed_resources:
+                ResourceAccess.objects.filter(
+                    system_user=system_user, resource_name__in=removed_resources
+                ).delete()
+
+            # Add newly-granted resources
+            added_resources = desired_set - current_resources
```
hugobessa, line 162: Grants are reconciled by set-difference inside a transaction: delete the rows whose resource_name dropped out, create the ones newly added. The secret is never touched — long_lived_token_hash and integration_name aren't in the update serializer's fields, so a PATCH can only change the resource grants (tests assert the hash is byte-identical in the DB afterward).

public_api/models.py
```diff
@@ -38,3 +40,11 @@ class ResourceAccess(BaseModel):
         on_delete=models.CASCADE,
     )
     resource_name = models.CharField(max_length=150, choices=PublicAPIResources, db_index=True)
+
+    class Meta:
+        constraints: ClassVar = [
+            models.UniqueConstraint(
+                fields=["system_user", "resource_name"],
```
hugobessa, line 47: UniqueConstraint(system_user, resource_name) closes a data-integrity gap a reviewer caught: without it, duplicate grant rows could exist (and the create loop could mint them from a request that repeated a resource). Paired with deduping the create input (dict.fromkeys), grants are now guaranteed unique per token.

public_api/serializers.py
```diff
@@ -65,3 +65,19 @@ class Meta:
...
+
+class SystemUserTokenUpdateSerializer(serializers.Serializer):
+    """Input serializer for updating a public-API token's resource grants (Phase 15).
+
+    Accepts ``available_resources`` (a non-empty list of valid ``PublicAPIResources`` values)
+    only.  ``integration_name`` and ``token`` are never accepted or changed.
+    The view reconciles ResourceAccess rows: adds rows for newly-granted resources,
+    removes rows for dropped resources.
+    """
+
```
hugobessa, line 136: The update serializer accepts only available_resources — this is what makes "edit permissions without re-issuing the secret" structurally true: there's no field through which integration_name or the token could be mutated on PATCH.

PR #59, opened 2026-06-05
feat(organizations): add invitation resend action

hugobessa:
## Summary
Follow-up gap (plan amendment). Adds `POST /invitations/{id}/resend/`: re-sends a pending invitation, regenerating the token and extending expiry. Reuses `OrganizationService.invite_user_to_organization`, whose get_or_create reset path already regenerates `token_hash`, extends `expires_at` by 7 days, and re-sends the email.

## Behavior
- Detail action on `OrganizationInvitationViewSet`; org-scoped (cross-org → 404); gated by `OrganizationInvitationPermission` (active member; non-member 403, anon 401).
- Refuses an already-accepted invitation (400).

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Amendment, Phase 17.

## Review history
Layer-3 review found a BLOCKER — the service call was wrapped in `except Exception` (would mask DB errors / real bugs as 400). `invite_user_to_organization` only raises `DuplicateInvitationError` (a DRF ValidationError subclass that already renders 400), so the try/except was removed entirely and unexpected errors now propagate as 500. Also added a real (non-mocked) test asserting token_hash changes + expires_at extends + the email is re-sent. (The initial commit also carried an AI co-author trailer — stripped to comply with the project's no-AI-trailer policy.)

## Test plan
- `uv run pytest organizations/tests/test_views.py -k "resend or invitation" -vs`
- Asserts: resend pending → 200 + token_hash changed + expires_at extended + email re-sent; accepted → 400; cross-org → 404; non-member → 403; anon → 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1442 passed).

### DIFF
--- a/organizations/views.py
+++ b/organizations/views.py
@@ -242,6 +242,54 @@ class OrganizationInvitationViewSet(NoUpdateVintaScheduleModelViewSet):
         """Revoke invitation by calling the service method."""
         self.organization_service.revoke_invitation(str(instance.id))
 
+    @extend_schema(
+        summary="Resend a pending organization invitation",
+        responses={
+            200: OrganizationInvitationSerializer,
+            400: OpenApiResponse(description="Invitation already accepted or service error"),
+            403: OpenApiResponse(description="Not an active member"),
+            404: OpenApiResponse(description="Invitation not found or cross-org"),
+        },
+    )
+    @action(detail=True, methods=["post"], url_path="resend")
+    def resend(self, request, pk=None):
+        """POST /invitations/{id}/resend/ — regenerate token and re-send a pending invitation.
+
+        Guards:
+        - Invitation must not be accepted (accepted_at is None).
+        - User must be an active member of the invitation's organization.
+
+        Returns the re-serialized invitation with the new token_hash and extended expires_at.
+        """
+        invitation = self.get_object()  # org-scoped; raises 404 if cross-org
+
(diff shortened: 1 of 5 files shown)


organizations/views.py
```diff
@@ -242,6 +242,51 @@ def perform_destroy(self, instance):
...
+        - Invitation must not be accepted (accepted_at is None).
+        - User must be an active member of the invitation's organization.
+
+        Returns the re-serialized invitation with the new token_hash and extended expires_at.
+        """
+        invitation = self.get_object()  # org-scoped; raises 404 if cross-org
+
+        # Guard: refuse if invitation is already accepted
+        if invitation.accepted_at is not None:
+            raise ValidationError(detail="Invitation already accepted.")
```
hugobessa, line 268: Resend is intentionally thin: it re-invokes invite_user_to_organization with the invitation's own email/name, and the service's get_or_create reset path regenerates the token, extends expiry, and re-sends the email. The only guard here is "not already accepted" (400). No try/except wrapper — the service's DuplicateInvitationError is already a DRF ValidationError (400); anything else is a real error and should surface as 500 rather than be masked.

PR #60, opened 2026-06-05
feat(organizations): configure rooms-sync creds via org PATCH and fix the trigger

hugobessa:
## Summary
Follow-up gap (plan amendment). Makes rooms-syncing actually usable:
- **Configure via org partial-update**: `PATCH /organizations/{id}/` accepts a nested `google_service_account` (email, audience, public_key + **write-only** private_key/private_key_id), upserting the org's `GoogleCalendarServiceAccount`. Per requester decision, config persists through the org Partial Update — no separate endpoint. Secrets are never returned (read shows email/audience + `configured`).
- **Fix the trigger**: `OrganizationService.request_rooms_sync` now authenticates the calendar service with that `GoogleCalendarServiceAccount` (the org-level one, `calendar_fk__isnull=True`) before `request_organization_calendar_resources_import`. Previously it called `initialize_without_provider` (account=None) → the import **raised at runtime**.

## Why this was a real bug
The import requires a provider-authenticated service (`is_authenticated_calendar_service` needs account + adapter). Phase 7's trigger and the pre-existing `create_organization` rooms path both used `initialize_without_provider`, so triggering a rooms import would 500. Now: configured → runs; not configured → **400** (never 500); `create_organization(should_sync_rooms=True)` with no creds no longer crashes (skips + logs).

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Amendment, Phase 18.

## Review history
Layer-3 review (credentials surface): 0 blockers — confirmed secrets can't leak through any response/exception/log. Fixed its should-fixes: TOCTOU on the post-commit trigger (now try/except + log so a deleted account can't raise unhandled in the on_commit callback), and reordered `update()` so the "enabling sync needs creds" check runs **before** any write (a rename+enable-without-creds PATCH now 400s without persisting the rename). Added authenticate + no-`public_key` assertions and a no-partial-write test.

## Test plan
- Asserts: PATCH creds → 200, stored encrypted, response has NO private_key/private_key_id/public_key; rotate on 2nd PATCH; configured → sync-rooms authenticates with the SA + enqueues; unconfigured → 400 (not 500); enable-without-creds → 400 with no partial write; create_organization no-creds → no crash; non-admin 403; cross-org 404; anon 401.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1456 passed).

### DIFF
--- a/organizations/serializers.py
+++ b/organizations/serializers.py
@@ -3,6 +3,7 @@ from typing import Annotated
 from dependency_injector.wiring import Provide, inject
 from rest_framework import serializers
 
+from calendar_integration.models import GoogleCalendarServiceAccount
 from common.utils.serializer_utils import VirtualModelSerializer
 from organizations.models import Organization, OrganizationInvitation, OrganizationMembership
 from organizations.services import OrganizationService
@@ -12,7 +13,51 @@ from organizations.virtual_models import (
 )
 
 
+class GoogleServiceAccountWriteSerializer(serializers.Serializer):
+    """Write-only nested serializer for configuring a Google Calendar service account.
+
+    Used within OrganizationSerializer's ``google_service_account`` field.
+    Accepts all five credential fields; ``private_key`` and ``private_key_id``
+    are write-only and are never echoed back in any response.
+    """
+
+    email = serializers.EmailField()
+    audience = serializers.CharField(max_length=255)
+    public_key = serializers.CharField()
+    private_key_id = serializers.CharField(max_length=255, write_only=True)
+    private_key = serializers.CharField(write_only=True)
+
+
+class GoogleServiceAccountReadSerializer(serializers.Serializer):
+    """Read-only nested serializer for the Google Calendar service account status.
+
+    Exposes only non-secret fields plus a ``configured`` boolean flag so the
+    frontend can display whether credentials are set without ever returning
+    ``private_key`` or ``private_key_id``.
+    """
+
+    email = serializers.CharField(read_only=True)
+    audience = serializers.CharField(read_only=True)
+    configured = serializers.SerializerMethodField()
+
+    def get_configured(self, obj: GoogleCalendarServiceAccount) -> bool:
+        """Return True always — presence of the object means it is configured."""
+        return True
+
+
 class OrganizationSerializer(VirtualModelSerializer):
+    """Serializer for Organization instances.
+
+    The ``google_service_account`` field supports both reading and writing:
+    - **Write**: accepts ``email``, ``audience``, ``public_key``,
+      ``private_key_id`` (write-only), and ``private_key`` (write-only).
+      Omitting the field on PATCH leaves existing credentials unchanged.
+    - **Read**: returns ``email``, ``audience``, and ``configured: true/false``.
+      Secret fields are never returned.
+    """
+
+    google_service_account = serializers.SerializerMethodField()
+
     class Meta:
         model = Organization
         virtual_model = OrganizationVirtualModel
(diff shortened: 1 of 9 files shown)


organizations/serializers.py
```diff
@@ -12,18 +13,74 @@
...
+    Accepts all five credential fields; ``private_key`` and ``private_key_id``
+    are write-only and are never echoed back in any response.
+    """
+
+    email = serializers.EmailField()
+    audience = serializers.CharField(max_length=255)
+    public_key = serializers.CharField()
+    private_key_id = serializers.CharField(max_length=255, write_only=True)
+    private_key = serializers.CharField(write_only=True)
+
```
hugobessa, line 29: Credentials surface: private_key and private_key_id are write_only on the WRITE serializer, and the WRITE serializer is used only for input validation in the view — it is never serialized into a response. Reads go through a separate read serializer (email/audience/`configured`) that has no secret fields at all, so the private key can't echo back through the org PATCH/GET.

organizations/services.py
```diff
@@ -78,18 +95,30 @@ def request_rooms_sync(
...
+        service_account = (
+            GoogleCalendarServiceAccount.objects.filter_by_organization(organization.id)
+            .filter(calendar_fk__isnull=True)
+            .first()
         )
+        if service_account is None:
+            raise NoServiceAccountConfiguredError()
+
+        self.calendar_service.authenticate(account=service_account, organization=organization)
         now = datetime.datetime.now(tz=datetime.UTC)
```
hugobessa, line 122: The actual bug fix: resolve the org-level GoogleCalendarServiceAccount (calendar_fk__isnull=True, org-scoped) and authenticate with it before the resource import — which requires a provider-authenticated service. The old initialize_without_provider left account=None, so the import raised. No account configured → NoServiceAccountConfiguredError (a 400), so callers never surface a 500.

organizations/views.py
```diff
@@ -112,7 +165,14 @@ def update(self, request, *args, **kwargs):
...
             def _trigger():
-                org_service.request_rooms_sync(organization=org, requested_by=user)
+                try:
+                    org_service.request_rooms_sync(organization=org, requested_by=user)
+                except NoServiceAccountConfiguredError:
+                    logger.warning(
+                        "rooms-sync trigger skipped: no service account configured for org %s "
+                        "(account may have been deleted between pre-flight check and commit)",
+                        org.id,
+                    )
```
hugobessa, line 175: update() validates "enabling should_sync_rooms requires configured-or- provided creds" BEFORE perform_update, so a PATCH that renames AND enables sync without creds returns 400 without persisting the rename (no partial write). Creds upsert runs before the transition trigger, so supplying creds in the same PATCH succeeds. The on_commit trigger is wrapped in try/except NoServiceAccountConfiguredError + log to avoid an unhandled raise if the account is deleted between the pre-check and commit (TOCTOU).

organizations/services.py
```diff
@@ -65,10 +71,21 @@ def create_organization(
...
+            try:
+                self.request_rooms_sync(
+                    organization=self.organization,
+                    requested_by=creator,
+                )
+            except NoServiceAccountConfiguredError:
+                logger.warning(
+                    "Skipping rooms sync for new organization %s: no service account configured.",
+                    self.organization.id,
+                )
```
hugobessa, line 88: create_organization(should_sync_rooms=True) catches NoServiceAccountConfigured Error and logs — a brand-new org never has creds yet, so org creation must not crash on the rooms-import attempt. The admin configures creds afterward via PATCH, which then triggers the sync.

PR #61, opened 2026-06-05
refactor(calendar): defer task enqueues via on_commit at service layer; filter disabled calendars

hugobessa:
## Summary
Final follow-up phase (plan amendment) — two pieces of hardening flagged across earlier phases:

1. **on_commit at the service layer**: the Celery `.delay(...)` calls inside `CalendarService.request_calendar_sync`, `request_organization_calendar_resources_import`, and `request_calendars_import` are now wrapped in `transaction.on_commit(...)`, so tasks enqueue only after the surrounding (ATOMIC_REQUESTS) transaction commits — closing the pre-existing race noted in Phases 4/5/7. The redundant view-layer `on_commit` wrappers (request_import; sync_rooms; the org update transition) were removed so each task defers exactly once, owned by the service. Synchronous rows (CalendarSync, the resources-import workflow) are still created before the deferral, so request-sync/admin-sync responses are unchanged.
2. **is_active filtering** (Phase 9 follow-up): disabled calendars no longer leak into selection/listing surfaces — `CalendarBundleCreateSerializer`, `ResourceAllocationSerializer`, and the public GraphQL `calendars` query now filter `is_active=True`.

## Deliberate non-change
Read-side **filtersets** (`?calendar=<id>` on events/blocked-times/groups/available-times) are NOT is_active-filtered: a user must still be able to filter by a *disabled* calendar's id (the data exists). Review caught an over-filter here and it was reverted. The bundle UPDATE serializer's union (keep existing disabled children) is untouched.

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Amendment, Phase 19.

## Review history
Layer-3 review (shared service, many callers): 0 blockers — confirmed single deferral on every caller path, closure args captured as locals (not mutable self), and rollback safety. Fixed its should-fixes: reverted the read-side filterset over-filter, added rollback tests (task NOT enqueued on transaction rollback) for the three methods, tightened a weak assertion, fixed a stale docstring.

## Test plan
- `uv run pytest calendar_integration/tests/ organizations/tests/ public_api/tests/ -n auto`
- Asserts: each method enqueues via on_commit + NOT before commit + NOT on rollback; request-sync/admin-sync still return the CalendarSync; disabled calendar excluded from bundle-create / resource-allocation / public GraphQL calendars; read filters by a disabled calendar id still work; existing Phase 4/7/18 flows green after removing the view wraps.
- Outer gate green: ruff, format --check, mypy (108, unchanged), makemigrations --check, check --deploy, `pytest -n auto` (1462 passed).


calendar_integration/services/calendar_service.py
```diff
@@ -2301,8 +2319,16 @@ def request_calendar_sync(
...
-            account_type, self.account.id, calendar_sync.id, calendar.organization_id
+        # Capture ids by value so the closure is independent of mutable self state.
+        _account_type = account_type
+        _account_id = self.account.id
+        _calendar_sync_id = calendar_sync.id
+        _organization_id = calendar.organization_id
+
+        transaction.on_commit(
+            lambda: sync_calendar_task.delay(  # type: ignore
+                _account_type, _account_id, _calendar_sync_id, _organization_id
```
hugobessa, line 2330: The .delay is deferred to on_commit and the closure references LOCALS (_account_type/_account_id/_calendar_sync_id/_organization_id) captured before the lambda — not self.account.id — so the per-account fresh-service loop in request_import (and any later mutation) can't corrupt the enqueued args. The CalendarSync row is still created synchronously above and returned, so the request-sync/admin-sync responses are unchanged; only the task enqueue moves to post-commit.

calendar_integration/services/calendar_service.py
```diff
@@ -408,13 +408,23 @@ def request_organization_calendar_resources_import(
...
-            account_id=self.account.id,
-            organization_id=self.organization.id,
-            import_workflow_state_id=import_workflow_state.id,
+            else "social_account"
+        )
+        _account_id = self.account.id
+        _organization_id = self.organization.id
+        _import_workflow_state_id = import_workflow_state.id
+
+        transaction.on_commit(
```
hugobessa, line 421: Single deferral, service-owned: the view-layer transaction.on_commit wrappers (request_import, sync_rooms, the org update transition) were removed, so each task is deferred exactly once here in the service. Under ATOMIC_REQUESTS it fires after the request commits; in a Celery/management context (no atomic) it fires immediately — both correct. Rollback tests assert the task does NOT enqueue if the surrounding transaction rolls back.

calendar_integration/serializers.py
```diff
@@ -197,7 +197,7 @@ def __init__(
             queryset=(
                 Calendar.objects.filter_by_organization(
                     organization_id=active_membership.organization_id
-                )
+                ).filter(is_active=True)
```
hugobessa, line 208: is_active=True on the bundle-CREATE choice querysets (and ResourceAllocation, and the public GraphQL calendars query) keeps disabled calendars out of selection/listing. Deliberately NOT applied to the read-side filtersets — filtering events by a disabled calendar's id must still work — and the bundle UPDATE serializer keeps its union (existing disabled children stay editable).

PR #62, opened 2026-06-06
feat(organizations): service account CRUD + org-wide calendar sync trigger

hugobessa:
## Summary
Follow-up gap (plan amendment, Phase 20). Two admin-only additions:

1. **Service Account CRUD** — a dedicated `ServiceAccountViewSet` at `/service-accounts/` for managing the organization's **org-level** Google Calendar service account (`calendar_fk IS NULL`, the one used for rooms sync). Full CRUD: list / retrieve / create / update (PUT rotate, PATCH partial) / delete. Secrets (`private_key`, `private_key_id`) are write-only and **never** returned; reads expose only `email`, `audience`, `configured`. One org-level account per organization — `create` refuses a duplicate (rotate via PUT/PATCH or DELETE first).
2. **Org-wide calendar sync** — `POST /organizations/{id}/sync-calendars/` enqueues a sync for every active calendar in the organization. Each calendar is authenticated with its **owner's** linked provider account (mirroring `admin-sync`); calendars with no owner or no linked account are reported under `skipped` rather than failing the whole request. Returns `{"synced": [...], "skipped": [{"calendar_id", "reason"}]}`.

## Decisions (confirmed with requester)
- **SA CRUD = org-level only.** Per-calendar service accounts remain internal/auto-assigned by the calendar auth flow and are not exposed (manual delete/reassign would silently break that calendar's sync).
- **Sync-all auth = each calendar's owner account** (skip + report un-syncable), not the org service account.
- **Sync-all endpoint lives on `OrganizationViewSet`** next to `sync-rooms`.

## Scope / security
- All endpoints admin-only (`IsOrganizationAdmin`); org-scoped `get_queryset` ⇒ cross-org id → 404; non-admin → 403; anon → 401.
- No model/migration changes (reuses the existing `GoogleCalendarServiceAccount` model).

## Plan reference
Plan `ai-plans/2026-06-05-REST_API_FRONTEND_GAPS_IMPLEMENTATION_PLAN.md` — Amendment, Phase 20.

## Test plan
- SA CRUD: create (201, encrypted, no secrets in body), duplicate → 400, list/retrieve (no secrets), PUT rotate, PATCH partial (email only, secret retained), DELETE (204); cross-org → 404; non-admin → 403; anon → 401.
- Sync-calendars (view, service patched): 202 + summary; bad datetime → 400; non-admin → 403; cross-org → 404; anon → 401.
- Sync-calendars (service unit, calendar service mocked): syncs calendars with owner+linked account; skips no-owner and no-linked-account with reasons; empty org → empty summary; inactive calendars excluded.
- Outer gate: ruff, format --check, mypy (no new errors), makemigrations --check (no changes), `pytest -n auto`, schema regenerated.


organizations/views.py
```diff
@@ -270,6 +275,165 @@ def sync_rooms(self, request, pk=None):
...
+
+        input_serializer = CalendarSyncRequestSerializer(data=request.data)
+        input_serializer.is_valid(raise_exception=True)
+        data = input_serializer.validated_data
+
+        result = self.organization_service.request_all_calendars_sync(
+            organization=org,
+            requested_by=request.user,
+            start_datetime=data["start_datetime"],
+            end_datetime=data["end_datetime"],
```
hugobessa, line 323: create() enforces the org-level singleton: at most one account with calendar_fk IS NULL per org. A duplicate create is a 400 directing the caller to PUT/PATCH (rotate) or DELETE first — this keeps the resource a true singleton while still exposing a REST-shaped collection. Responses always use ServiceAccountReadSerializer so private_key/private_key_id/public_key never echo.

organizations/services.py
```diff
@@ -125,6 +130,68 @@ def request_rooms_sync(
...
+                continue
+
+            self.calendar_service.authenticate(account=social_account, organization=organization)
+            self.calendar_service.request_calendar_sync(
+                calendar=calendar,
+                start_datetime=start_datetime,
+                end_datetime=end_datetime,
+                should_update_events=should_update_events,
+            )
+            synced.append(calendar.id)
```
hugobessa, line 197: request_all_calendars_sync mirrors admin-sync's auth model: per active calendar, resolve the default CalendarOwnership → owner's SocialAccount for the calendar's provider, then authenticate + request_calendar_sync (which already defers the Celery enqueue via on_commit). Un-syncable calendars (no owner / no linked account) are collected into `skipped` with a reason instead of raising, so one bad calendar never fails the whole org sync. Inactive calendars are excluded.

organizations/serializers.py
```diff
@@ -45,6 +45,40 @@ def get_configured(self, obj: GoogleCalendarServiceAccount) -> bool:
...
+
+    private_key_id = serializers.CharField(max_length=255, write_only=True)
+    private_key = serializers.CharField(max_length=255, write_only=True)
+
+    class Meta:
+        model = GoogleCalendarServiceAccount
+        fields = ("email", "audience", "public_key", "private_key_id", "private_key")
+
+
 class OrganizationSerializer(VirtualModelSerializer):
```
hugobessa, line 86: Two serializers split the credentials surface: ServiceAccountWriteSerializer declares private_key/private_key_id as write_only (input only), and every response is rendered with ServiceAccountReadSerializer, which has no secret fields at all (id, email, audience, configured, created, modified). The secret can't round-trip back out through any CRUD response.

organizations/views.py
```diff
@@ -270,6 +275,165 @@ def sync_rooms(self, request, pk=None):
...
+        """Org-scoped queryset limited to the org-level service account."""
+        user = self.request.user
+        if not user.is_authenticated:
+            return GoogleCalendarServiceAccount.objects.none()
+        membership = get_active_organization_membership(user)
+        if membership is None:
+            return GoogleCalendarServiceAccount.objects.none()
+        return GoogleCalendarServiceAccount.objects.filter_by_organization(
+            membership.organization_id
+        ).filter(calendar_fk__isnull=True)
```
hugobessa, line 356: ServiceAccountViewSet manages ONLY the org-level account (get_queryset filters calendar_fk__isnull=True), so per-calendar service accounts — which the calendar auth flow assigns automatically — are never reachable here. Deleting one of those would silently break that calendar's sync, so they are deliberately out of scope (requester decision). Org-scoped queryset means a cross-org id is a 404, not a 403.

PR #64, opened 2026-06-13
fix(calendar): make Google service-account rooms sync work (DWD + Admin SDK)

hugobessa:
## Summary
Fixes the Google Workspace **rooms sync** for service accounts, which has never produced correct output — every attempt failed with "credentials do not contain the necessary fields". Three phases, one PR (modular commits — read the commit log top-to-bottom as a table of contents):

1. **Phase 1 — rename `audience` → `admin_email`** (this commit). The field on `GoogleCalendarServiceAccount` stored the Google Workspace admin email used as the domain-wide-delegation (DWD) impersonation subject, but was misnamed `audience` (a leftover from the broken JWT-`aud` flow). Renamed across model, migration, both REST serializers, the org PATCH view, the adapter `TypedDict`, the service-layer credentials dict, and all tests. **No auth behaviour change yet.**
2. **Phase 2 — replace the broken JWT auth** with `google.oauth2.service_account.Credentials.from_service_account_info(...).with_subject(admin_email)` + DWD. Removes `_generate_jwt` + `from_service_account_credentials`; adds `from_service_account` (bypasses `__init__` via `cls.__new__`, builds Calendar API `client` + Admin SDK `admin_client` from one DWD cred). **Done.**
3. **Phase 3 — list Workspace resource calendars via the Admin SDK** (`resources().calendars().list(customer="my_customer")`, paginated, `maxResults=500`) instead of the SA's always-empty personal `calendarList`. Both `get_calendar_resources` / `get_calendar_resource` guard with an eager `NotImplementedError` on non-SA adapters. **Done.**

## Why this is a bug, not a feature
The old SA path generated a JWT and used it directly as a bearer token; Google returns 401 and the client's auto-refresh fails (`refresh_token` is None). And `get_calendar_resources()` listed the SA's own (empty) calendar list, so even a working auth would import zero rooms. Both are replaced in Phases 2–3.

## No feature flag
Per the plan's Goals/Non-goals: rooms sync has never worked, so there is no pre-existing behaviour to gate off.

## Plan reference
Plan `ai-plans/2026-06-09-GOOGLE_SA_ROOMS_SYNC_IMPLEMENTATION_PLAN.md`.

## Review history
Each phase: 3-layer review (mechanical + plan-compliance + independent adversarial reviewer), 0 BLOCKERs across all three.
- **Phase 1** — kept the column at `varchar(255)` (model + migration `max_length=255`) so the rename is catalog-only with no row-narrowing scan; read serializer field → `EmailField` for a correct `format: email` OpenAPI annotation. The two legacy-JWT should-fixes were deferred to Phase 2, where that code is deleted outright.
- **Phase 2** — strengthened the unit tests to assert the exact `from_service_account_info` payload + `scopes=_SA_SCOPES`, and the integration test to assert the SA branch calls `from_service_account` with `admin_email`/`email`.
- **Phase 3** — made the `admin_client` guard eager (public `get_calendar_resources` wrapper raises before delegating to the private `_iter_calendar_resources` generator), so a non-SA adapter fails at call time like `get_calendar_resource`. **Documented deviation:** the plan listed updating `test_execute_organization_calendar_resources_import_calls_adapter`; that test mocks at the `get_available_calendar_resources` boundary (returns `CalendarResourceData`, above the Admin-SDK layer), so it already passes unchanged and the Admin-SDK field mapping is fully covered by the adapter unit tests — updating it would be hollow churn.

## Test plan
- `uv run pytest calendar_integration/tests/ organizations/tests/ -n auto`
- Outer gate green after every phase: `ruff check`, `ruff format --check`, `makemigrations --check` ("No changes detected"), `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1552 passed** at final phase. The single failure `accounts/tests/test_account_adapters.py::...test_send_unknown_account_sms_success` is pre-existing on `main` (mocks a removed `get_twilio_client`) and unrelated to this change.
- Adapter unit tests cover: both DWD clients built + payload/scopes, `with_subject(admin_email)`, Admin-SDK field mapping, pagination via `nextPageToken`, and `NotImplementedError` on non-SA adapters.

## Post-merge / ops note
DWD must be granted to the service account in the Google Admin Console (Security → API Controls → Domain-wide delegation) for both scopes (`admin.directory.resource.calendar.readonly` + `calendar.readonly`) before rooms sync will succeed; without it the Admin SDK call returns 403. After merge, admins re-trigger `/organizations/{id}/sync-rooms/` per org to import rooms as `CalendarType.RESOURCE`.


calendar_integration/migrations/0018_rename_audience_admin_email.py
```diff
@@ -0,0 +1,38 @@
...
+            name="admin_email",
+            field=models.EmailField(
+                blank=True,
+                help_text=(
+                    "Google Workspace super-admin email used as the DWD impersonation subject. "
+                    "The service account must have domain-wide delegation granted for the Admin SDK "
+                    "and Calendar API scopes in the Google Admin Console."
+                ),
+                max_length=255,
+            ),
```
hugobessa, line 36: Two ops, both catalog-only on Postgres. RenameField is a metadata rename of a low-traffic column (one row per org). The AlterField keeps max_length=255 on purpose: the field switches CharField→EmailField, and EmailField's Django default is 254 — emitting 254 would narrow varchar(255)→varchar(254) and force a full-row length-validation scan. Pinning 255 makes it a pure catalog update (just help_text + the app-level EmailValidator). Reverse path restores the original CharField(max_length=255) + name `audience`.

calendar_integration/models.py
```diff
@@ -1614,7 +1614,15 @@ class GoogleCalendarServiceAccount(OrganizationModel):
...
     email = models.EmailField()
-    audience = models.CharField(max_length=255, blank=True)
+    admin_email = models.EmailField(
+        max_length=255,
+        blank=True,
+        help_text=(
+            "Google Workspace super-admin email used as the DWD impersonation subject. "
+            "The service account must have domain-wide delegation granted for the Admin SDK "
+            "and Calendar API scopes in the Google Admin Console."
+        ),
```
hugobessa, line 1624: Rename rationale: this field is the DWD impersonation *subject* — a Google Workspace super-admin email passed to `.with_subject(...)` in Phase 2 — not a JWT `aud` claim. The old name `audience` came from the broken bearer-JWT flow being removed. max_length=255 mirrors the migration (see the migration comment) to avoid a column narrow.

organizations/serializers.py
```diff
@@ -80,17 +80,17 @@ class ServiceAccountWriteSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = GoogleCalendarServiceAccount
-        fields = ("email", "audience", "public_key", "private_key_id", "private_key")
+        fields = ("email", "admin_email", "public_key", "private_key_id", "private_key")
```
hugobessa, line 83: Beyond the two serializers named in the plan, the two ModelSerializers here (`ServiceAccountReadSerializer` / `ServiceAccountWriteSerializer`) list the field explicitly in their `fields` tuples — leaving `audience` there would raise ImproperlyConfigured at import once the model field is gone. So these edits are required by the rename, not scope creep. `public_key` is deliberately retained (API compat; unused by the new auth).

calendar_integration/services/calendar_adapters/google_calendar_adapter.py
```diff
@@ -66,7 +66,7 @@ class GoogleCredentialTypedDict(TypedDict):
 class GoogleServiceAccountCredentialsTypedDict(TypedDict):
     account_id: str
     email: str
-    audience: str
+    admin_email: str
```
hugobessa, line 72: Cross-phase coupling: this TypedDict key is the contract between CalendarService and the adapter. Phase 2 adds `from_service_account()` which reads `credentials["admin_email"]` and feeds it to `.with_subject(...)`, so the key name has to land in Phase 1 before the auth rewrite consumes it.

calendar_integration/services/calendar_service.py
```diff
@@ -332,7 +332,7 @@ def get_calendar_adapter_for_account(
                     "public_key": account.public_key,
                     "private_key_id": account.private_key_id,
                     "private_key": account.private_key,
-                    "audience": account.audience,
+                    "admin_email": account.admin_email,
```
hugobessa, line 335: This dict is built from the stored GoogleCalendarServiceAccount row and handed to the adapter. Phase 1 only renames the key here; Phase 2 changes the *consumer* (`from_service_account_credentials` → `from_service_account`) in the same `get_calendar_adapter_for_account` branch.

hugobessa:
All 3 phases complete on `plan/google-sa-rooms-sync` (modular commits — one PR). Phase 1 rename → Phase 2 DWD auth → Phase 3 Admin SDK listing. Outer gate green each phase (final `pytest -n auto`: 1552 passed; the lone `accounts` SMS failure is pre-existing on `main`). Each phase passed a 3-layer review with 0 BLOCKERs. **Before this works in prod:** grant the service account domain-wide delegation in the Google Admin Console for both readonly scopes, then re-trigger `/organizations/{id}/sync-rooms/` per org.

PR #65, opened 2026-06-13
refactor(calendar): extract shared context + utils, fix multi-tenant calendar-lookup cache

hugobessa:
## Summary
Phase 0 of the calendar-service refactor (splitting the 4,726-line `CalendarService` monolith into a thin facade + focused sub-services). This is **pure scaffolding** that later phases build on, plus one sanctioned correctness fix:

1. **`CalendarServiceContext`** (`calendar_service_context.py`) — a frozen dataclass snapshot of the auth context (organization, user_or_token, account, calendar_adapter, permission/side-effects services). Built once in `authenticate()` / `initialize_without_provider()`; consumed by the sub-services extracted in Phases 2–6.
2. **`calendar_service_utils.py`** — moves the cross-cutting helpers out of the class as module-level functions: timezone conversion, the event-serialization family, permission-granting, and the two calendar lookups. The facade methods now delegate to them (kept as thin wrappers this phase; shrunk in Phase 7).
3. **Multi-tenant `lru_cache` fix** — the calendar lookups were `@lru_cache`-decorated instance methods keyed only on `id`/`external_id`. On a reused service instance that could return **another organization's** `Calendar`. Replaced with a per-instance dict keyed on `(organization_id, id_or_external_id)`, reset on every (re)authentication.

## Behavior preservation
Everything except the lru fix is byte-for-byte. The full `calendar_integration` suite (1014 tests) passes with **no edits to existing assertions**. Where the original code carried a latent bug, it is preserved deliberately and pinned by a test (see inline comments) — out of scope to fix here.

## No feature flag
Pure refactor — correctness is proven by the unchanged existing suite. The lone behavior change (lru fix) is guarded by a new regression test, not a flag. Per the plan's Goals/Non-goals.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 0.

## Review history
Layer-3 adversarial review: 0 blockers. Should-fixes addressed in-phase: reverted an unsanctioned behavior change in `serialize_event_data_input` (the refactor had "fixed" a latent bug — restored byte-for-byte and pinned it with a test); hoisted two function-local imports to module level; added the plan-mandated delegation test + a `serialize_event_data_input` resource-path test; typed the test cache dicts. NIT deferred: the per-instance cache is unbounded vs the old `maxsize=128` — bounded in practice by distinct calendars per org per instance lifetime (noted, not a leak on current call sites).

## Test plan
- `uv run pytest calendar_integration/tests/ -n auto` → 1014 passed.
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → 1563 passed. The single failure `accounts/tests/test_account_adapters.py::...test_send_unknown_account_sms_success` is pre-existing on `main` (unrelated SMS-adapter test) and the diff is calendar-only.

### DIFF
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -2,9 +2,7 @@ import copy
 import datetime
 import json
 import logging
-import zoneinfo
 from collections.abc import Callable, Iterable
-from functools import lru_cache
 from typing import Annotated, Any, Literal, TypedDict, cast
 
 from django.conf import settings
@@ -61,6 +59,34 @@ from calendar_integration.models import (
 from calendar_integration.querysets import CalendarEventQuerySet
 from calendar_integration.recurrence_utils import OccurrenceValidator, RecurrenceRuleSplitter
 from calendar_integration.services.calendar_permission_service import CalendarPermissionService
+from calendar_integration.services.calendar_service_context import CalendarServiceContext
+from calendar_integration.services.calendar_service_utils import (
+    convert_naive_utc_datetime_to_timezone as _convert_naive_utc_datetime_to_timezone,
+)
+from calendar_integration.services.calendar_service_utils import (
+    get_calendar_by_external_id as _get_calendar_by_external_id_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    get_calendar_by_id as _get_calendar_by_id_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    grant_calendar_owner_permissions as _grant_calendar_owner_permissions_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    grant_event_attendee_permissions as _grant_event_attendee_permissions_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event as _serialize_event_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_data_input as _serialize_event_data_input_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_external_attendee as _serialize_event_external_attendee_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_internal_attendee as _serialize_event_internal_attendee_util,
+)
 from calendar_integration.services.calendar_side_effects_service import CalendarSideEffectsService
 from calendar_integration.services.dataclasses import (
     ApplicationCalendarData,
@@ -130,6 +156,13 @@ class CalendarService(BaseCalendarService):
         self.calendar_adapter = None
         self.calendar_side_effects_service = calendar_side_effects_service
         self.calendar_permission_service = calendar_permission_service
+        # Per-instance calendar lookup cache: keyed on (organization_id, id_or_external_id).
+        # This replaces the @lru_cache approach which was keyed only on id/external_id and
+        # could return a cached Calendar from a different organization when the service
+        # instance is reused across organizations (multi-tenant safety bug).
+        self._calendar_cache: dict[tuple[int, str | int], Calendar] = {}
+        # Shared auth-context snapshot; set by authenticate() / initialize_without_provider().
+        self._context: CalendarServiceContext | None = None
 
     def _grant_calendar_owner_permissions(self, calendar: Calendar) -> None:
         """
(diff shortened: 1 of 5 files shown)


calendar_integration/services/calendar_service.py
```diff
@@ -130,6 +156,13 @@ def __init__(
         self.calendar_adapter = None
         self.calendar_side_effects_service = calendar_side_effects_service
         self.calendar_permission_service = calendar_permission_service
+        # Per-instance calendar lookup cache: keyed on (organization_id, id_or_external_id).
+        # This replaces the @lru_cache approach which was keyed only on id/external_id and
+        # could return a cached Calendar from a different organization when the service
+        # instance is reused across organizations (multi-tenant safety bug).
+        self._calendar_cache: dict[tuple[int, str | int], Calendar] = {}
+        # Shared auth-context snapshot; set by authenticate() / initialize_without_provider().
+        self._context: CalendarServiceContext | None = None
```
hugobessa, line 165: The multi-tenant fix. The old lookups were `@lru_cache(maxsize=128)` on instance methods, keyed only on the positional `id`/`external_id` — so a `CalendarService` instance reused across organizations (a pattern the DI Factory provider allows) could return org A's cached `Calendar` for an org-B lookup. This per-instance dict keys on `(organization_id, id_or_external_id)` so a cross-org hit is impossible. Trade-off: unbounded vs the old `maxsize=128` — acceptable because growth is bounded by distinct calendars per org per instance lifetime and reset on every (re)auth (see below).

calendar_integration/services/calendar_service.py
```diff
@@ -424,6 +330,20 @@ def authenticate(
...
+        # Build the immutable auth-context snapshot consumed by sub-services.
+        self._context = CalendarServiceContext(
+            organization=self.organization,
+            user_or_token=self.user_or_token,
+            account=self.account,
+            calendar_adapter=self.calendar_adapter,
+            calendar_permission_service=self.calendar_permission_service,
+            calendar_side_effects_service=self.calendar_side_effects_service,
+        )
+
```
hugobessa, line 346: Two things happen on every authenticate: the calendar cache is reset (so no stale cross-org entry survives a re-auth to a different org) and the immutable `CalendarServiceContext` snapshot is built. The same block is mirrored in `initialize_without_provider()`. The existing instance attributes (`self.organization` / `self.account` / `self.calendar_adapter`) are still set exactly as before — the `type_guards` check them by attribute, so they must stay on the facade.

calendar_integration/services/calendar_service_context.py
```diff
@@ -0,0 +1,49 @@
...
+        CalendarSideEffectsService,
+    )
+    from calendar_integration.services.protocols.calendar_adapter import CalendarAdapter
+    from organizations.models import Organization
+    from public_api.models import SystemUser
+    from users.models import User
+
+
+@dataclasses.dataclass(frozen=True)
+class CalendarServiceContext:
```
hugobessa, line 34: Frozen dataclass, built but **not yet consumed** in this phase. It's the seam the sub-services extracted in Phases 2–6 (`CalendarEventService`, `CalendarBundleService`, `AvailabilityService`, `CalendarSyncService`, `CalendarWebhookService`) receive instead of re-authenticating — that's the plan's performance guardrail (authenticate once → one adapter construction / token refresh). Cross-phase coupling: don't drop it as "unused" before Phase 2 lands.

calendar_integration/services/calendar_service_utils.py
```diff
@@ -0,0 +1,421 @@
...
+                title=resource_allocation.calendar.name,
+                email=resource_allocation.calendar.email,
+                external_id=resource_allocation.calendar.external_id,
+                status=cast(
+                    Literal["accepted", "declined", "pending"],
+                    resource_allocation.status,
+                ),
+            )
+            # For recurring instances, get resource allocations from the parent event; for regular events, use their own
+            for resource_allocation in Calendar.objects.filter(
```
hugobessa, line 262: Preserved-as-is on purpose. This iterates `Calendar.objects.filter(...)` but the body accesses `resource_allocation.calendar.*` / `.status` — a latent pre-existing bug (the loop variable is a `Calendar`, which has no `.calendar` / `.status`). It only fires when `resource_allocations` is non-empty and matches a RESOURCE calendar, which no current code path exercises. Phase 0 is byte-for-byte, so this is **deliberately not fixed here** and is pinned by a test (next comment). Fixing it belongs in a separate scoped change.

calendar_integration/tests/services/test_calendar_service_utils.py
```diff
@@ -0,0 +1,370 @@
...
+    iterates ``Calendar`` objects (the ``Calendar.objects.filter(...)`` result) but then
+    accesses ``resource_allocation.calendar.name`` / ``.status`` as if each item were a
+    ``ResourceAllocation``. A ``Calendar`` has no ``.calendar`` attribute, so when the
+    resource_allocations match a RESOURCE calendar in the org, evaluating the resources
+    list raises ``AttributeError``. We assert that here so the preserved behavior is
+    explicit and any future "fix" is a deliberate, reviewed change rather than an
+    accidental Phase 0 deviation.
+    """
+
+    def test_matching_resource_allocation_raises_attribute_error(self, db, organization):
```
hugobessa, line 313: This test asserts `serialize_event_data_input` *raises* `AttributeError` on a matching resource allocation — i.e. it pins the latent bug noted above as the current behavior, so a future "cleanup" that silently changes it gets caught. The companion regression test (`test_same_external_id_different_orgs...`) pins the lru fix: two orgs sharing an `external_id` must each get their own Calendar through a shared cache.

PR #66, opened 2026-06-13
feat(organizations): allow users to belong to multiple organizations (Phase 1)

hugobessa:
## Summary
Phase 1 of the **Multi-Organization Membership** plan — the foundation. Converts
`OrganizationMembership.user` from `OneToOneField` to `ForeignKey` so one user can hold
memberships in many organizations, guarded by a composite `unique(user, organization)`. All
existing single-membership behavior is preserved; later phases add header-driven active-org
resolution.

## What changed
- `OrganizationMembership.user`: `OneToOneField` → `ForeignKey`, `related_name` `organization_membership` → `organization_memberships`. New `UniqueConstraint(user, organization)` named `uniq_membership_user_organization` (replaces the implicit `UNIQUE(user_id)` the OneToOne carried).
- Migration `0006`: `AlterField` + `AddConstraint`, reversible (the reverse restores the OneToOne + single unique).
- `get_active_organization_membership(user)` rewired from the OneToOne reverse descriptor to a manager query — Phase 1 returns the user's single active membership (`is_active=True`, deterministic `order_by("created")` fallback). Phase 2a layers header-driven resolution on top of this same seam.
- `User.is_organization_admin(organization)` now resolves admin rights **per-org** via the membership manager instead of inspecting a single membership.
- The `accept_invitation` / `provision_tenant_for_user` provisioning guards keep their exact pre-Phase-1 semantics (refuse re-provisioning when the user has **any** membership, active or inactive). Phase 4 deliberately relaxes this to per-org.
- Swept every direct singular `user.organization_membership` reverse read in app code (`calendar_integration/filtersets.py`, `calendar_integration/serializers.py`, `public_api/queries.py`) and test files to the new manager.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 1 + Data Model Changes (Membership FK) + Guiding Decisions (no feature flag — pre-production, 1:1 migration).

## Review history
Layer-3 adversarial review: no blockers. One SHOULD-FIX caught + fixed — the provisioning guards had been narrowed to active-only, which would have let a deactivated user self-provision a second membership; restored to "any membership row" to preserve Phase-1 semantics. Stale docstring + nondeterministic ordering also fixed; added a test for the active-vs-inactive resolution branch.

## Test plan
- `uv run pytest organizations/tests/test_models.py organizations/tests/test_services.py -vs`
- Asserts: a user holds memberships in two orgs; `unique(user, organization)` rejects a same-org duplicate (`IntegrityError`); `is_organization_admin` is per-org; `get_active_organization_membership` ignores an inactive membership in another org.
- Outer gate green: ruff, format --check, `makemigrations --check` (no changes), `check --deploy`, `pytest -n auto` (1555 passed; 1 sanctioned pre-existing failure `test_send_unknown_account_sms_success`, red on base too).


organizations/models.py
```diff
@@ -130,24 +135,29 @@ class OrganizationMembership(BaseModel):
...
+        scoping code — always go through ``get_active_organization_membership`` so
+        the resolution seam is centralised.
     """
 
-    user = models.OneToOneField(
+    user = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE,
-        related_name="organization_membership",
+        related_name="organization_memberships",
```
hugobessa, line 160: Core of the feature: OneToOne → ForeignKey. This drops the implicit UNIQUE(user_id) the OneToOne carried; the composite UniqueConstraint(user, organization) below is the replacement backstop. The hard-gate invariant in the class docstring was rewritten — "gated" now means **zero active memberships**, not "no membership row". Resolution of *which* org is active for a multi-org user is deferred to Phase 2a (header-driven); Phase 1 keeps every existing single-membership flow byte-for-byte.

organizations/models.py
```diff
@@ -27,15 +27,20 @@ def get_active_organization_membership(
...
-        return None
-    if not membership.is_active:
+    if user is None:
         return None
-    return membership
+    # Stable ordering: once later phases allow multiple active memberships, an
+    # accidental two-active-row state must resolve deterministically. Multi-active
+    # resolution becomes header-driven in Phase 2a, so this ordering is only a
+    # deterministic fallback.
+    return user.organization_memberships.filter(is_active=True).order_by("created").first()  # type: ignore[union-attr]
```
hugobessa, line 43: Temporary Phase-1 resolution. The reverse accessor is now a manager, so this can no longer be the OneToOne descriptor read. Returns the single active membership; `order_by("created")` is a deterministic fallback only — multi-active-membership is not reachable via any guarded write path this phase, and Phase 2a replaces this with the `_active_membership` stash set from the `X-Organization-Id` header. Keeping the same function signature is what lets the ~60 existing call sites stay untouched.

organizations/services.py
```diff
@@ -273,17 +273,18 @@ def accept_invitation(self, token: str, user: User) -> OrganizationMembership:
...
-        # DESIGN: hasattr check intentionally matches BOTH active and inactive memberships.
-        # An inactive membership still has a DB row, so accept_invitation must refuse
-        # re-provisioning here.  The supported un-disable path is admin REACTIVATION
-        # (setting is_active=True); not re-accepting an invitation.
-        if hasattr(user, "organization_membership"):
+        # DESIGN: this guard refuses re-provisioning when the user already has ANY
+        # membership (active or inactive) — preserving the original OneToOne-era
+        # semantics and the admin-reactivation-only un-disable path. Phase 4
+        # deliberately relaxes this to per-org for multi-org invite accept.
+        if user.organization_memberships.exists():
```
hugobessa, line 287: Deliberately checks `.exists()` (ANY membership, active OR inactive), not `get_active_organization_membership(...) is not None` (active-only). The original OneToOne-era guard refused re-provisioning for inactive memberships too — the supported un-disable path is admin reactivation, not re-accepting an invite. Phase 1 must preserve that exact semantic; Phase 4 is where this relaxes to "refuse only a duplicate in the *same* org" for multi-org invite accept.

users/models.py
```diff
@@ -52,15 +52,14 @@ def is_organization_admin(self, organization) -> bool:
...
+
         organization_id = getattr(organization, "id", organization)
-        return membership.organization_id == organization_id and getattr(
-            membership, "is_admin", False
-        )
+        return self.organization_memberships.filter(  # type: ignore[attr-defined]
+            organization_id=organization_id,
+            is_active=True,
+            role=OrganizationRole.ADMIN,
+        ).exists()
```
hugobessa, line 62: `is_organization_admin` is now genuinely per-org — it filters the membership manager by the passed `organization_id` rather than reading one membership and comparing its org. This is what makes a user able to be admin of org A and a plain member of org B. The local import of `OrganizationRole` avoids the organizations↔users circular import (same pattern the method already used).

PR #67, opened 2026-06-13
feat(notifications): in-app notifications via vintasend — adapter, DI wiring & REST endpoints

hugobessa:
## Summary
Adds in-app notifications backed by vintasend's native service methods, plus the internal REST surface to consume them. Single plan branch, atomic commits per logical unit (read the commit log top-to-bottom as a table of contents).

Phases (each a group of commits on this branch):
- **Phase 0 — adapter, renderer & DI wiring** *(this update)*: makes `notification_type=IN_APP` a working channel so `create_notification(... IN_APP ...)` and `get_in_app_unread` stop raising. No user-visible endpoint yet.
- **Phase 1 — list unread endpoint** (`GET /notifications/unread/`, native `get_in_app_unread`).
- **Phase 2 — list all endpoint** (`GET /notifications/`, read + unread via thin ORM).
- **Phase 3 — mark-as-read endpoint** (`POST /notifications/{id}/mark-read/`, native `mark_read`, ownership-enforced).

## Phase 0 detail
1. **`DjangoInAppNotificationAdapter`** (`notifications/notification_adapters/django_in_app.py`) — `notification_type = IN_APP`; `send()` renders the body template to validate it, then returns. Persistence + status transitions stay with the backend. vintasend_django ships no in-app adapter (only a `FakeInAppAdapter` test stub), so this is repo-local.
2. **`DjangoTemplatedInAppRenderer`** (`notifications/notification_template_renderers/django_in_app_renderer.py`) — single-body Django-template renderer (SMS-style), conforming to vintasend's `BaseNotificationTemplateRenderer`.
3. **`FixedDjangoDbNotificationBackend`** (`notifications/notification_backends.py`) — a one-method subclass fixing a real bug in the vendored backend (see inline comment): the unread queryset filtered by the raw `NotificationTypes.IN_APP` enum, which Django serializes via `str()` to `"NotificationTypes.IN_APP"` ≠ the stored `"IN_APP"`, so the query always returned empty.
4. **DI wiring** (`di_core/containers.py`) — appends the IN_APP adapter and swaps the service-level `notification_backend` to the fixed backend. Email/SMS entries untouched.
5. **Context registration** — `notifications/apps.py` `ready()` imports `notification_contexts`, registering an example `in_app_generic_context` + `templates/notifications/in_app/example.body.txt`.

## No feature flag
Purely additive surface: a new IN_APP channel and (later phases) brand-new endpoints. Appending an adapter to the DI list cannot change Email/SMS behavior. Per the plan's Goals/Non-goals.

## No migration
Reuses the existing `vintasend_django.Notification` model. Zero DDL.

## Plan reference
Plan `ai-plans/2026-06-13-IN_APP_NOTIFICATIONS_IMPLEMENTATION_PLAN.md`.

## Review history (Phase 0)
Layer-3 adversarial review: **0 blockers**. The `FixedDjangoDbNotificationBackend` deviation was verified correct, minimal, and Email/SMS-safe. Three should-fixes applied in-phase: removed a redundant `filter_in_app_unread_notifications` override (parent already delegates to the fixed queryset method); corrected a misleading adapter docstring about `context_used` (it stores the raw context, not the rendered body — the body is re-rendered on read); added a DI-container end-to-end unread test guarding the service-level backend swap. NITs (TypeVar bound widened, hollow test assertion made explicit) also applied.

## Test plan
- `uv run pytest notifications/tests/ -n auto` → 9 passed.
- Outer gate: `ruff check` clean, `ruff format --check` clean, `manage.py check --deploy` (5 pre-existing security warnings only), `uv run pytest -n auto` → 1559 passed. The single failure `accounts/tests/test_account_adapters.py::TestAccountAdapter::test_send_unknown_account_sms_success` is pre-existing on `main` (SMS adapter test, fails because phone verification is temporarily disabled) and unrelated to this notifications-only diff.


notifications/notification_backends.py
```diff
@@ -0,0 +1,40 @@
...
+    which would otherwise serialise to "NotificationTypes.IN_APP" via str().
+    """
+
+    def _get_all_in_app_unread_notifications_queryset(
+        self, user_id: int | str | uuid.UUID
+    ) -> QuerySet[NotificationModel]:
+        return NotificationModel.objects.filter(
+            user_id=str(user_id),
+            status=NotificationStatus.SENT.value,
+            notification_type=NotificationTypes.IN_APP.value,
```
hugobessa: This one-method subclass is the load-bearing deviation. The vendored DjangoDbNotificationBackend._get_all_in_app_unread_notifications_queryset filters `notification_type=NotificationTypes.IN_APP` — the raw Enum instance. `NotificationTypes` is a plain `enum.Enum` (not a str-Enum), so Django's CharField.get_prep_value calls `str()` on it and produces "NotificationTypes.IN_APP", which never matches the stored "IN_APP". The unread query therefore always returns empty upstream. We override only the queryset method to use `.value`; the parent `filter_in_app_unread_notifications` and `filter_all_in_app_unread_notifications` both delegate here, so they're fixed transitively with no further overrides.

di_core/containers.py
```diff
@@ -70,8 +75,12 @@ class AppContainer(containers.DeclarativeContainer):
                 DjangoTemplatedSMSRenderer(),
                 DjangoDbNotificationBackend(),
             ),
+            DjangoInAppNotificationAdapter(
+                DjangoTemplatedInAppRenderer(),
+                FixedDjangoDbNotificationBackend(),
+            ),
         ],
-        notification_backend=DjangoDbNotificationBackend(),
+        notification_backend=FixedDjangoDbNotificationBackend(),
```
hugobessa: Two additive changes: (1) append the IN_APP adapter so `get_in_app_unread` doesn't raise NotificationError("No in-app notification adapter found"); (2) swap the service-level `notification_backend` to FixedDjangoDbNotificationBackend — this is the one that matters for reads, because `get_in_app_unread` calls `self.notification_backend.filter_in_app_unread_notifications`. The Email/SMS adapter entries above are untouched, and the fixed backend overrides only the IN_APP query path, so existing channels behave byte-for-byte as before.

notifications/notification_adapters/django_in_app.py
```diff
@@ -0,0 +1,51 @@
...
+
+        The notification is already persisted before send() is called.
+        DjangoDbNotificationBackend.mark_pending_as_sent() transitions the status
+        after this method returns without raising.
+
+        :param notification: The notification to send.
+        :param context: The context dict generated from the registered context function.
+        :raises NotificationBodyTemplateRenderingError: When body template rendering fails.
+        """
+        self.template_renderer.render(notification, context)
```
hugobessa, line 51: Render-on-send is validation only — it confirms the body template renders and then discards the output. The rendered body is NOT persisted; vintasend's backend stores the raw `context_used` dict. Read paths (Phases 1–2) re-render `body_template` against `context_used` at serialization time. Persistence and the PENDING_SEND→SENT transition are owned by the backend, not this adapter.

notifications/apps.py
```diff
@@ -5,3 +5,8 @@ class NotificationsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "notifications"
     verbose_name = "Notifications"
+
+    def ready(self) -> None:
+        # Import the notification contexts module so @register_context decorators
+        # run and register functions with vintasend's Contexts singleton on app load.
+        import notifications.notification_contexts  # noqa: F401
```
hugobessa, line 12: Context registration hook: importing the module at app-ready runs the `@register_context("in_app_generic_context")` decorator against vintasend's Contexts singleton. Mirrors the pattern in users/apps.py. `register_context` is an idempotent dict assignment, so the import is safe even if triggered more than once.

notifications/tests/test_in_app_send.py
```diff
@@ -0,0 +1,150 @@
...
+        has_in_app = any(
+            a.notification_type == NotificationTypes.IN_APP
+            for a in notification_service.notification_adapters
+        )
+        assert has_in_app, (
+            "DI container notification_service is missing the IN_APP adapter; "
+            "get_in_app_unread would raise NotificationError"
+        )
+
+    def test_di_get_in_app_unread_returns_sent_notification(self, di_container, user) -> None:
```
hugobessa, line 128: Regression guard for the service-level backend swap in di_core/containers.py. It exercises create→get_in_app_unread through the REAL DI container (not a hand-built service), so reverting `notification_backend` to the buggy DjangoDbNotificationBackend would make this fail rather than silently returning an empty unread list.

notifications/views.py
```diff
@@ -0,0 +1,162 @@
...
+        parsed = int(value)
+    except (ValueError, TypeError) as exc:
+        raise ValidationError({param_name: f"Must be a positive integer, got {value!r}."}) from exc
+    if parsed < 1:
+        raise ValidationError({param_name: f"Must be >= 1, got {parsed}."})
+    return parsed
+
+
+@extend_schema(tags=["Notifications"])
+class NotificationViewSet(GenericViewSet):
```
hugobessa: Deliberately `GenericViewSet` only — NOT `ListModelMixin`. Inheriting the list mixin would register a live `GET /notifications/` route this phase, and its `get_queryset` has no status filter, so it would leak PENDING_SEND/FAILED/ CANCELLED rows. The list endpoint is Phase 2's job (it adds status__in=[SENT, READ]). `get_queryset` is kept here only for Phase 2/3 reuse; the `unread` action does not read from it.

notifications/views.py
```diff
@@ -0,0 +1,162 @@
...
+            param_name="page",
+            default=_DEFAULT_PAGE,
+        )
+        page_size = min(
+            _parse_positive_int(
+                request.query_params.get("page_size", ""),
+                param_name="page_size",
+                default=_DEFAULT_PAGE_SIZE,
+            ),
+            _MAX_PAGE_SIZE,
```
hugobessa, line 234: page_size is clamped to _MAX_PAGE_SIZE (100). The native get_in_app_unread → _paginate_queryset slices by the raw page_size with no upper bound, so without this an authenticated caller could request page_size=100000 and force an unbounded fetch. The clamped value is echoed back in the envelope so the client sees what was actually applied.

notifications/serializers.py
```diff
@@ -0,0 +1,101 @@
...
+
+    id = serializers.CharField(read_only=True)
+    title = serializers.CharField(read_only=True)
+    notification_type = serializers.CharField(read_only=True)
+    status = serializers.CharField(read_only=True)
+    body = serializers.SerializerMethodField()
+    created = serializers.SerializerMethodField()
+    modified = serializers.SerializerMethodField()
+
+    def get_body(self, obj: object) -> str:
```
hugobessa, line 48: Plain `Serializer` (not ModelSerializer) on purpose: it serializes BOTH the vintasend `Notification` dataclasses returned by get_in_app_unread (Phase 1) AND the ORM model rows Phase 2 will list. `body` is re-rendered from `body_template` against `context_used`/`context_kwargs` at read time (no rendered body is persisted); `created`/`modified` are model-only and resolve to None for dataclasses via getattr.

notifications/views.py
```diff
@@ -0,0 +1,207 @@
...
+        list-all and Phase 3's mark-read endpoints.  The unread @action bypasses
+        this queryset and calls get_in_app_unread() directly via the service.
+        """
+        if not self.request.user.is_authenticated:
+            return Notification.objects.none()
+        return Notification.objects.filter(
+            user=self.request.user,
+            notification_type=NotificationTypes.IN_APP.value,
+            status__in=[NotificationStatus.SENT.value, NotificationStatus.READ.value],
+        ).order_by("-created", "-id")
```
hugobessa: The thin-ORM "list all" query. No native vintasend method lists read+unread, so this filters the vendored Notification model directly: user-scoped, notification_type=IN_APP, and status in (SENT, READ) — deliberately EXCLUDING PENDING_SEND/FAILED/CANCELLED so internal pipeline states never reach the user. `order_by("-created", "-id")` — the `-id` tiebreaker keeps "newest first" deterministic when two rows share a created timestamp.

notifications/views.py
```diff
@@ -0,0 +1,207 @@
...
+        parsed = int(value)
+    except (ValueError, TypeError) as exc:
+        raise ValidationError({param_name: f"Must be a positive integer, got {value!r}."}) from exc
+    if parsed < 1:
+        raise ValidationError({param_name: f"Must be >= 1, got {parsed}."})
+    return parsed
+
+
+@extend_schema(tags=["Notifications"])
+class NotificationViewSet(ListModelMixin, GenericViewSet):
```
hugobessa: Phase 2 re-adds `ListModelMixin` (Phase 1 had dropped it to avoid shipping an unfiltered list route early). `GET /notifications/` uses the project default LimitOffsetPagination → `{count, next, previous, results}`, intentionally a DIFFERENT envelope + query params (`limit`/`offset`) than the unread action's `page`/`page_size` passthrough, because the native get_in_app_unread can't provide a total count.

notifications/views.py
```diff
@@ -0,0 +1,263 @@
...
+        owned), then calls the native mark_read service method. If the notification
+        is already READ, returns 200 idempotently without calling mark_read (which
+        would raise NotificationUpdateError).
+
+        Returns the serialized updated notification.
+        """
+        # get_object() uses get_queryset(), which filters by user, notification_type,
+        # and status (SENT/READ). This automatically enforces ownership + 404 for
+        # non-existent, wrong-user, or wrong-status notifications (e.g., PENDING_SEND).
+        notification = self.get_object()
```
hugobessa: Ownership + 404 boundary. `get_object()` runs against the user-scoped `get_queryset` (user=request.user, notification_type=IN_APP, status in SENT/READ), so another user's notification, a non-existent id, or a PENDING/FAILED/CANCELLED row all 404 here — the native `mark_read` (which has NO ownership check) is never reached with an unvetted id. This is the IDOR guard.

notifications/views.py
```diff
@@ -0,0 +1,263 @@
...
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        # Status is SENT — call mark_read to transition to READ.
+        # A concurrent request may win the race and mark it READ first; in that
+        # case mark_read raises NotificationUpdateError (0-row update).  Treat
+        # that as idempotent success — the row is already READ.
+        try:
+            self.notification_service.mark_read(notification.pk)
+        except NotificationUpdateError:
+            pass
```
hugobessa: Two-layer idempotency. (1) An already-READ row short-circuits to 200 without calling the native method — `mark_sent_as_read` filters `status=SENT` and raises NotificationUpdateError on 0 rows updated, so calling it on a READ row would 500. (2) The try/except handles the concurrent race: if a parallel request flips the row to READ between our check and the update, the native call raises and we swallow it, re-fetch, and still return an idempotent 200.

di_core/containers.py
```diff
@@ -70,6 +74,10 @@ class AppContainer(containers.DeclarativeContainer):
                 DjangoTemplatedSMSRenderer(),
                 DjangoDbNotificationBackend(),
             ),
+            DjangoInAppNotificationAdapter(
+                DjangoTemplatedInAppRenderer(),
+                DjangoDbNotificationBackend(),
+            ),
         ],
         notification_backend=DjangoDbNotificationBackend(),
```
hugobessa, line 82: Reverted to the stock `DjangoDbNotificationBackend` on both the IN_APP adapter and the service-level backend. vintasend-django 1.2.0 fixes the IN_APP enum-filter bug that 1.1.3 had (the unread queryset filtered by the raw `NotificationTypes.IN_APP` enum → `"NotificationTypes.IN_APP"` ≠ stored `"IN_APP"`), so the repo-local `FixedDjangoDbNotificationBackend` workaround is deleted.

notifications/views.py
```diff
@@ -0,0 +1,359 @@
...
+                description=(
+                    "Passthrough-paginated list of all in-app notifications. "
+                    "Envelope: {results: [...], page: int, page_size: int, count: int}."
+                )
+            ),
+            400: OpenApiResponse(description="Invalid page / page_size parameter"),
+            401: OpenApiResponse(description="Unauthenticated"),
+        },
+    )
+    def list(self, request: Request, *args, **kwargs) -> Response:  # type: ignore[override]
```
hugobessa, line 149: List-all now uses native `get_in_app_notifications(user_id, page, page_size)` + `get_in_app_notifications_count` (1.2.0) instead of the old ORM + LimitOffsetPagination. Both list endpoints now share `_paginated_envelope` → `{results, page, page_size, count}` with `page_size` clamped to 100. `pagination_class = None` (line 76) stops DRF injecting limit/offset into the schema. `get_queryset` is kept only for the single mark-read `get_object()` lookup.

notifications/views.py
```diff
@@ -0,0 +1,359 @@
...
+                    "Notifications marked as read. "
+                    "Envelope: {results: [...serialized notifications that are READ after the op...]}."
+                )
+            ),
+            400: OpenApiResponse(description="Invalid request body (empty or missing ids list)"),
+            401: OpenApiResponse(description="Unauthenticated"),
+        },
+    )
+    @action(detail=False, methods=["post"], url_path="mark-read-bulk")
+    def mark_read_bulk(self, request: Request) -> Response:
```
hugobessa, line 308: Bulk mark-read uses a DISTINCT `url_path="mark-read-bulk"` (not `mark-read`) on purpose: sharing the detail action's `url_path` makes drf-spectacular derive the same operationId for both → a collision warning + an order-dependent `_2` suffix. A distinct path gives stable, unique operationIds with no global schema hooks. `mark_read_bulk(ids, user_id=request.user.id)` is ownership-scoped (foreign ids silently skipped — IDOR-safe) and idempotent.

notifications/serializers.py
```diff
@@ -0,0 +1,118 @@
...
+    Input serializer for the bulk mark-as-read endpoint.
+
+    Validates that the request body contains a non-empty list of notification ids.
+    Empty list or missing field → DRF 400 validation error.
+    """
+
+    ids = serializers.ListField(
+        child=serializers.IntegerField(),
+        allow_empty=False,
+        max_length=100,
```
hugobessa, line 117: `BulkMarkReadSerializer.ids` is bounded with `max_length=100`, symmetric with the list endpoints' `page_size` cap — an unbounded id list would mean a huge `IN (...)` UPDATE. Empty/missing list → 400; over 100 ids → 400.

notifications/views.py
```diff
@@ -0,0 +1,330 @@
...
+        parsed = int(value)
+    except (ValueError, TypeError) as exc:
+        raise ValidationError({param_name: f"Must be a positive integer, got {value!r}."}) from exc
+    if parsed < 1:
+        raise ValidationError({param_name: f"Must be >= 1, got {parsed}."})
+    return parsed
+
+
+@extend_schema(tags=["Notifications"])
+class NotificationViewSet(ViewSet):
```
hugobessa, line 51: Plain `ViewSet` (not `GenericViewSet`) on purpose: there is no model queryset to expose — every endpoint reads/writes through the native vintasend service only. `serializer_class` (below) is kept solely as a drf-spectacular schema hint; the serializer is instantiated directly in each action. Zero ORM in this view.

notifications/views.py
```diff
@@ -0,0 +1,330 @@
...
+        - already-READ owned id → returned unchanged (idempotent 200).
+        No ORM query, no get_object() — the native method is the only data path.
+        """
+        user_id = request.user.id
+        if user_id is None:
+            return Response(status=status.HTTP_401_UNAUTHORIZED)
+
+        results = list(self.notification_service.mark_read_bulk([pk], user_id=user_id))
+        if not results:
+            raise NotFound("Notification not found.")
```
hugobessa, line 283: Single mark-read is now fully native: it delegates to `mark_read_bulk([pk], user_id=request.user.id)` instead of an ORM get_object()/get_queryset. mark_read_bulk scopes the update + read-back to the user, so a foreign / missing / non-SENT-non-READ id yields an empty result → 404 (IDOR guard), a SENT id is marked + returned, and an already-READ owned id is returned unchanged (idempotent 200). This also removed the previous TOCTOU try/except — bulk is idempotent by construction, so there is no 0-row-update 500 to guard against.

PR #68, opened 2026-06-13
refactor(calendar): extract recurrence exception + bulk-modification engines into RecurrenceManager

hugobessa:
## Summary
Phase 1 of the calendar-service refactor. Extracts the two generic recurrence template-method engines out of the `CalendarService` monolith into a new **stateless** `RecurrenceManager` (`calendar_integration/services/recurrence_manager.py`):
- `create_recurring_exception_generic` — the shared engine behind recurring **event**, **blocked-time**, and **available-time** exception creation.
- `create_recurring_bulk_modification_generic` — the shared engine behind "modify/cancel from date onward" for the same three entity families.

The facade keeps every public method signature and now delegates at 6 call sites: `self._recurrence_manager.create_recurring_*_generic(self._context, ...)`. `calendar_service.py` shrinks ~222 lines.

## Why a stateless helper (not a service)
Per the plan's Guiding Decisions: the recurrence machinery is shared by three entity families, so it belongs in one DRY helper — but it carries no auth state, so it's a plain class constructed once in `__init__`, not a DI provider. The engines used `self` for exactly one thing — the auth guard — so they now take the `CalendarServiceContext` as their first parameter and run the guard against it. The per-type logic was already fully externalized as caller-supplied callbacks (truncate / continuation / record builders), which stay as parameters.

## Behavior preservation
Byte-for-byte. The engine bodies were moved verbatim (only `self.<x>` → param substitution); reviewer confirmed via line-by-line `diff` (only a trailing blank line differs). The full `calendar_integration` suite (1022) passes with no assertion edits.

## No feature flag
Pure refactor. Per the plan's Goals/Non-goals.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 1.

## Review history
Layer-3 adversarial review: 0 blockers (engine bodies confirmed byte-identical, guard semantics preserved, all 6 call sites wired to the correct per-type callbacks, removed imports confirmed unused). One should-fix addressed in-phase: added a direct test for the master-date exception branch (`exception_date == parent.start_time.date()`), which had been covered only indirectly through the facade suite. NITs left as-is: a duplicated comment and the bare-`raise` defensive line are verbatim from the original (byte-for-byte scope).

## Test plan
- New `calendar_integration/tests/services/test_recurrence_manager.py` (9 tests) — exercises both engines directly through real model objects, incl. the master-date branch + the unauthenticated-context guard.
- `uv run pytest calendar_integration/tests/ -n auto` → 1022 passed (pre-fix; 1023 with the added test).
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → 1572 passed + the one known pre-existing `accounts/.../test_send_unknown_account_sms_success` failure (unrelated; diff is calendar-only).

### DIFF
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -2,9 +2,7 @@ import copy
 import datetime
 import json
 import logging
-import zoneinfo
 from collections.abc import Callable, Iterable
-from functools import lru_cache
 from typing import Annotated, Any, Literal, TypedDict, cast
 
 from django.conf import settings
@@ -61,6 +59,34 @@ from calendar_integration.models import (
 from calendar_integration.querysets import CalendarEventQuerySet
 from calendar_integration.recurrence_utils import OccurrenceValidator, RecurrenceRuleSplitter
 from calendar_integration.services.calendar_permission_service import CalendarPermissionService
+from calendar_integration.services.calendar_service_context import CalendarServiceContext
+from calendar_integration.services.calendar_service_utils import (
+    convert_naive_utc_datetime_to_timezone as _convert_naive_utc_datetime_to_timezone,
+)
+from calendar_integration.services.calendar_service_utils import (
+    get_calendar_by_external_id as _get_calendar_by_external_id_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    get_calendar_by_id as _get_calendar_by_id_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    grant_calendar_owner_permissions as _grant_calendar_owner_permissions_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    grant_event_attendee_permissions as _grant_event_attendee_permissions_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event as _serialize_event_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_data_input as _serialize_event_data_input_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_external_attendee as _serialize_event_external_attendee_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_internal_attendee as _serialize_event_internal_attendee_util,
+)
 from calendar_integration.services.calendar_side_effects_service import CalendarSideEffectsService
 from calendar_integration.services.dataclasses import (
     ApplicationCalendarData,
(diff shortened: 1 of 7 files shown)


calendar_integration/services/recurrence_manager.py
```diff
@@ -0,0 +1,273 @@
...
+if TYPE_CHECKING:
+    from calendar_integration.services.calendar_service_context import CalendarServiceContext
+    from calendar_integration.services.protocols.base_calendar_service import BaseCalendarService
+
+
+class RecurrenceManager:
+    """Stateless engine for recurring-object exceptions and bulk modifications."""
+
+    def create_recurring_exception_generic(
+        self,
```
hugobessa, line 56: Stateless by design: constructed once in `CalendarService.__init__`, holds no auth state. The original engine methods used `self` for exactly one thing — the auth guard — so the context is threaded in as the first parameter instead. Everything per-entity-type (the truncate / continuation / record callbacks) is still passed by the caller, so the same engine serves events, blocked-times, and available-times with no branching inside.

calendar_integration/services/recurrence_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        :param create_new_recurring_callback: Callback to create new recurring object after master modification
+        :param create_modified_object_callback: Callback to create modified object for non-cancelled exceptions
+        :param exception_manager_update_callback: Callback to update exception manager references
+        :param exception_manager_delete_callback: Callback to delete exception manager references
+        :return: Created/modified object or None if cancelled
+        """
+        if not is_initialized_or_authenticated_calendar_service(
+            cast("BaseCalendarService", context)
+        ):
+            raise
```
hugobessa, line 92: Guard equivalence: the original ran `is_initialized_or_authenticated_calendar_service(self)` against `self.organization`. Here it runs against the threaded `CalendarServiceContext` (cast to the protocol via a string literal so no runtime import / cycle). When the facade is unauthenticated `self._context` is `None`, and `hasattr(None, "organization")` is `False`, so the guard raises the same `CalendarServiceOrganizationNotSetError` as before. The bare `raise` on the next line is dead defensive code preserved verbatim — the guard always raises internally first; don't "fix" it here (byte-for-byte).

calendar_integration/services/calendar_service.py
```diff
@@ -1926,7 +1820,8 @@ def delete_exception_manager(parent_obj: RecurringMixin) -> None:
             "timezone": modified_timezone,
         }
 
-        result = self._create_recurring_exception_generic(
+        result = self._recurrence_manager.create_recurring_exception_generic(
```
hugobessa, line 1823: Representative delegation site (one of 6: 3 exception + 3 bulk-mod, across event / blocked-time / available-time). Only the receiver changed and `self._context` was inserted as the first arg — every following kwarg + the per-type callbacks are byte-identical to what the in-class engine received. `self._context` is the snapshot Phase 0 builds in `authenticate()` / `initialize_without_provider()`.

PR #69, opened 2026-06-13
feat(organizations): resolve active org from X-Organization-Id header

hugobessa:
## Summary
Phase 2a of the **Multi-Organization Membership** plan. Adds a request-scoped resolver that
selects the active organization from the `X-Organization-Id` header and stashes it where the
existing tenant-scoping call sites already read — so multi-org switching works without touching
~60 call sites. Only the happy-path rows of the resolution table are implemented; the 400
(absent + multi-org) and 403 (non-member org) rows are deliberate no-regression stubs for
Phases 2b/2c.

## What changed
- New `TenantScopedViewMixin` in `common/utils/view_utils.py`, prepended to all 8 shared base viewsets. Its `initial()` runs `super().initial()` first (so DRF auth has populated `request.user`), then resolves `X-Organization-Id` → the caller's active membership and stashes it on `request.organization`, `request.organization_membership`, and `request.user._active_membership`.
- `get_active_organization_membership(user)` reads the `_active_membership` stash via an `_UNSET` sentinel (distinguishes "resolved to gated/None" from "not on a DRF request"); off-request callers (Celery, mgmt commands, direct tests) fall back to the single-membership DB query.
- The header lookup is strictly scoped to `request.user.organization_memberships` — no request can resolve to an org the caller isn't an active member of.
- Malformed (non-integer) header falls through to the fallback instead of 500.
- `CreateModelMixin.create` re-resolves the active org after `perform_create` (a service may have just created the user's first membership) so the post-create re-fetch honors the header.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 2a + Guiding Decisions (resolution seam = stash on request/user) + API Design (resolution table). Use-case 2.

## Review history
Layer-3 adversarial review caught one BLOCKER: the original post-create `del user._active_membership` dropped to the header-blind fallback, so a multi-org user creating under header B then re-fetching via `get_queryset` (scoped to the first membership) would 500 / cross-org. Fixed by re-resolving after `perform_create`, with a fail-before/pass-after regression test through `CalendarViewSet`. Also fixed: tests now exercise real tenant data (list-A vs list-B + create-under-B 201) not just the `current` echo; malformed-header 500 guarded. Note: org PK is an integer `BigAutoField` (the plan said "uuid" — corrected here).

## Test plan
- Asserts: header A → org-A data, header B → org-B data through `CalendarViewSet`; create under `X-Organization-Id: <B>` returns 201 and lands in org B (regression for the BLOCKER); single-membership absent-header behavior unchanged; gated + unauthenticated pass through; malformed header does not 500.
- Outer gate green: ruff, format --check, mypy (138, unchanged baseline), `makemigrations --check`, `check --deploy`, `pytest -n auto` (1570 passed; 1 sanctioned pre-existing failure).

### DIFF
--- a/common/utils/view_utils.py
+++ b/common/utils/view_utils.py
@@ -1,9 +1,181 @@
+from __future__ import annotations
+
+import logging
+from typing import Any
+
 from django.shortcuts import get_object_or_404
 
 import django_virtual_models as v
 from rest_framework import generics, mixins, status
+from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet, ViewSetMixin
+
+
+logger = logging.getLogger(__name__)
+
+#: Header name used to select the active organization for a request.
+ACTIVE_ORG_HEADER = "X-Organization-Id"
+
+
+class TenantScopedViewMixin:
+    """Resolve the active organization for every DRF request.
+
+    This mixin must be included in every base viewset so that all internal REST
+    endpoints automatically pick up the ``X-Organization-Id`` header.  The resolver
+    runs **after** ``super().initial()`` so that DRF authentication has already
+    populated ``request.user`` — the JWT user is not available at Django-middleware
+    time.
+
+    After this mixin runs, three attributes are available on every DRF request:
+
+    - ``request.organization_membership`` — the resolved ``OrganizationMembership``
+      or ``None`` (gated / unauthenticated user).
+    - ``request.organization`` — the resolved ``Organization`` or ``None``.
+    - ``request.user._active_membership`` — same value as
+      ``request.organization_membership``.  ``get_active_organization_membership``
+      reads this stash so all ~60 existing call sites are header-aware without
+      change.
+
+    Resolution table (Phase 2a implements the happy-path rows; stubs for 2b/2c are noted):
+
+    +-----------------------+---------------------------------+------------------------------------------+
+    | Memberships (active)  | Header                          | Result (Phase 2a)                        |
+    +-----------------------+---------------------------------+------------------------------------------+
+    | 0                     | any                             | gated (membership = None)                |
+    | 1                     | absent                          | resolve to that membership               |
+    | 1                     | present, matches                | resolve to it                            |
+    | 2+                    | present, matches member         | resolve to named org                     |
+    | 2+                    | absent                          | fallback to first [Phase 2b → 400]       |
+    | any                   | present, non-member org         | fallback to first [Phase 2c → 403]       |
+    | any                   | present, non-integer            | fallback to first (malformed, not 500)   |
+    +-----------------------+---------------------------------+------------------------------------------+
+
+    Unauthenticated requests pass through untouched (the mixin sets ``None`` on
+    all three attributes so downstream code doesn't KeyError); DRF's own
+    authentication / permission stack returns 401 before any business logic runs.
+    """
+
+    def initial(self, request: Request, *args: Any, **kwargs: Any) -> None:
+        """Run DRF initialisation, then resolve and stash the active org."""
+        super().initial(request, *args, **kwargs)  # type: ignore[misc]
+        self._resolve_active_organization(request)
+
+    def _resolve_active_organization(self, request: Request) -> None:  # noqa: C901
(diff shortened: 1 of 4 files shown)


common/utils/view_utils.py
```diff
@@ -1,11 +1,183 @@
...
+
+    Unauthenticated requests pass through untouched (the mixin sets ``None`` on
+    all three attributes so downstream code doesn't KeyError); DRF's own
+    authentication / permission stack returns 401 before any business logic runs.
+    """
+
+    def initial(self, request: Request, *args: Any, **kwargs: Any) -> None:
+        """Run DRF initialisation, then resolve and stash the active org."""
+        super().initial(request, *args, **kwargs)  # type: ignore[misc]
+        self._resolve_active_organization(request)
```
hugobessa, line 62: The whole multi-org seam lives here. `super().initial()` MUST run first — the JWT user is set by DRF authentication during `initial()`, not by Django middleware, so resolving before it would read AnonymousUser. The mixin is prepended to all 8 base viewsets so this wires once. Endpoints NOT on these bases (ServiceAccountViewSet, AcceptInvitationView) are onboarding/raw DRF and intentionally fall back to the off-request single-membership query.

common/utils/view_utils.py
```diff
@@ -1,11 +1,183 @@
...
+                # Header present and is a valid integer: try to find a matching active membership.
+                matching = (
+                    user.organization_memberships.filter(  # type: ignore[union-attr]
+                        is_active=True,
+                        organization_id=org_id_header,
+                    )
+                    .select_related("organization")
+                    .first()
+                )
+                if matching is not None:
```
hugobessa, line 118: Tenant isolation: the header lookup filters by `user.organization_memberships` AND `organization_id=<header>`, so a caller can only ever resolve to an org they are an active member of — a forged/guessed header for someone else's org matches nothing and falls through to their own membership. Phase 2c upgrades that no-match case from silent fallback to an explicit 403.

common/utils/view_utils.py
```diff
@@ -197,6 +369,14 @@ def create(self, request, *args, **kwargs):
         self.perform_create(serializer)
         instance = serializer.instance
 
+        # A service may have created the user's first membership during perform_create
+        # (e.g. OrganizationService.create_organization), making the stash set in
+        # TenantScopedViewMixin.initial() stale. Re-resolve so the post-create re-fetch
+        # honors the X-Organization-Id header (and any newly-created membership) instead
+        # of silently dropping to the header-blind single-membership fallback.
+        if hasattr(self, "_resolve_active_organization"):
+            self._resolve_active_organization(request)
```
hugobessa, line 378: Why re-resolve instead of just clearing: a service called during perform_create (e.g. OrganizationService.create_organization) can create the user's first membership, making the stash set in initial() stale. The earlier version `del`-eted the stash, which dropped the post-create re-fetch into the header-blind single-membership fallback and 500'd a multi-org create under header B. Re-resolving honors both the new membership and the header. The `test_create_under_header_b_returns_201_and_lands_in_org_b` test locks this.

organizations/models.py
```diff
@@ -27,19 +33,30 @@ def get_active_organization_membership(
...
-    # accidental two-active-row state must resolve deterministically. Multi-active
-    # resolution becomes header-driven in Phase 2a, so this ordering is only a
-    # deterministic fallback.
+
+    stashed = getattr(user, "_active_membership", _UNSET)
+    if stashed is not _UNSET:
+        # DRF request path: the resolver has already run; trust its result
+        # (may be an OrganizationMembership or None for gated users).
+        return stashed  # type: ignore[return-value]
+
```
hugobessa, line 56: The `_UNSET` sentinel is load-bearing: a DRF request that resolves to a gated user stashes `None`, and we must NOT fall back to the DB query in that case (it would re-derive a membership the resolver deliberately withheld). `getattr(user, "_active_membership", _UNSET)` + `is not _UNSET` distinguishes "resolver ran, answer is None" from "no resolver ran" (Celery/mgmt/test). This is what keeps the header authoritative on the request path.

PR #70, opened 2026-06-13
feat(organizations): require X-Organization-Id for multi-org requests

hugobessa:
## Summary
Phase 2b of the **Multi-Organization Membership** plan. A caller with 2+ active memberships who
omits `X-Organization-Id` now gets a **400** instead of silently resolving to an arbitrary org.
Single-membership callers and gated (onboarding) users are unaffected.

## What changed
- `TenantScopedViewMixin._resolve_active_organization`: the absent-header + 2+ active-membership row now raises `ValidationError` → 400 `{"detail": "X-Organization-Id header required."}`. Rows kept unchanged: 0 → gated, 1 + absent → resolve, header-present-match → resolve, non-member org → still the Phase 2c fallback stub.
- A **malformed (non-integer) header** is now treated identically to a missing header — single → resolve, multi-org → 400, gated → gated. This replaced the earlier "fall back to the first membership" stub, which would have silently served org A's data to a multi-org caller who sent garbage (a tenant-ambiguity bug).
- New opt-out class attribute `active_org_resolution_optional: bool = False`. A view that must serve multi-org callers without the header (Phase 3's `GET /organizations/mine/`, onboarding) sets it `True` to skip the 400 and resolve to gated. No shipped view sets it this phase.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 2b + API Design resolution table. Use-case 3.

## Review history
Layer-3 review caught one edge BLOCKER: malformed-header handling for a multi-org caller. Resolved by unifying malformed → absent-header semantics (the safe 400) and removing the dead fallback, plus a multi-org-malformed test asserting 400 and tightening the single-membership-malformed test to assert 200.

## Test plan
- Asserts: two memberships + no header → 400 through `CalendarViewSet`; one membership + no header → 200; gated user + no header → not 400 (onboarding intact); two memberships + valid header → 200; multi-org + malformed header → 400; single membership + malformed header → 200; opt-out view → not 400.
- Outer gate fully green: ruff, format, mypy (baseline), `makemigrations --check`, `check --deploy`, `pytest -n auto` (1578 passed, 0 failures).

### DIFF
--- a/common/utils/view_utils.py
+++ b/common/utils/view_utils.py
@@ -7,6 +7,7 @@ from django.shortcuts import get_object_or_404
 
 import django_virtual_models as v
 from rest_framework import generics, mixins, status
+from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet, ViewSetMixin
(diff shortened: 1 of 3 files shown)


common/utils/view_utils.py
```diff
@@ -153,14 +168,17 @@ def _resolve_active_organization(self, request: Request) -> None:  # noqa: C901
...
-                    # first membership so nothing regresses.
-                    resolved_membership = active_memberships[0]
+                    # Multi-org caller with no header: the active org is ambiguous.
+                    # Reject with 400 so we never silently pick one — unless the
+                    # concrete view opted out (org discovery / onboarding), in which
+                    # case resolution falls through to gated (None).
+                    if not getattr(self, "active_org_resolution_optional", False):
+                        raise ValidationError(
+                            {"detail": "X-Organization-Id header required."},
+                        )
```
hugobessa, line 178: This is the core of Phase 2b: a multi-org caller who omits the header can never resolve to an implicit org — we 400 rather than guess. Raising inside `_resolve_active_organization` (called from both `initial()` and the post-perform_create re-resolve) means the 400 fires consistently on reads and writes. DRF renders ValidationError as 400 for every base viewset.

common/utils/view_utils.py
```diff
@@ -37,25 +38,45 @@ class TenantScopedViewMixin:
...
     Unauthenticated requests pass through untouched (the mixin sets ``None`` on
     all three attributes so downstream code doesn't KeyError); DRF's own
     authentication / permission stack returns 401 before any business logic runs.
     """
 
+    #: When ``True``, a multi-org caller that omits ``X-Organization-Id`` is *not*
+    #: rejected with a 400 — the active org resolves to ``None`` instead.  Concrete
+    #: views that must function without the header (org discovery, onboarding) opt
+    #: in.  See the class docstring's resolution table for the affected row.
+    active_org_resolution_optional: bool = False
```
hugobessa, line 78: The opt-out. Phase 3's `GET /organizations/mine/` is the org-discovery endpoint — a multi-org user calls it precisely because they don't yet know which id to put in the header, so it must NOT 400. Setting this True makes the absent-header-multi-org row resolve to gated (None) instead of raising. Defaults False; no shipped view opts out this phase.

common/utils/view_utils.py
```diff
@@ -37,25 +38,45 @@ class TenantScopedViewMixin:
...
     +-----------------------+---------------------------------+------------------------------------------+
     | 0                     | any                             | gated (membership = None)                |
     | 1                     | absent                          | resolve to that membership               |
     | 1                     | present, matches                | resolve to it                            |
     | 2+                    | present, matches member         | resolve to named org                     |
-    | 2+                    | absent                          | fallback to first [Phase 2b → 400]       |
+    | 2+                    | absent                          | **400** (X-Organization-Id required)     |
     | any                   | present, non-member org         | fallback to first [Phase 2c → 403]       |
-    | any                   | present, non-integer            | fallback to first (malformed, not 500)   |
+    | any                   | present, non-integer            | treated as absent header                 |
```
hugobessa, line 52: A malformed (non-integer) header is deliberately treated as a missing header, not as "fall back to the user's first org". Falling back would silently serve one tenant's data to a multi-org caller who sent garbage — the exact ambiguity the 400 exists to prevent. So malformed inherits the absent-header rules: 1 → resolve, 2+ → 400, 0 → gated.

PR #71, opened 2026-06-13
feat(organizations): reject non-member X-Organization-Id with 403

hugobessa:
## Summary
Phase 2c of the **Multi-Organization Membership** plan — the last resolver phase. A request whose
`X-Organization-Id` names an org the caller is NOT an active member of is now refused with **403**,
instead of silently falling back to the caller's first membership. This closes the resolution
table.

## What changed
- `TenantScopedViewMixin._resolve_active_organization`: the valid-integer-but-no-matching-active-membership branch raises `PermissionDenied` → 403, replacing the Phase 2a/2b fallback stub. Covers both "not a member of that org" and "membership in that org is `is_active=False`" (the lookup filters `is_active=True`).
- The `active_org_resolution_optional` opt-out now skips BOTH the 2b 400 and the 2c 403, resolving to gated `None` — so Phase 3's `GET /organizations/mine/` and onboarding flows are unaffected by a stray header.
- Untouched: malformed-header handling (absent-header semantics), absent-header 400, happy-path resolution.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 2c + API Design resolution table. Use-case 4.

## Review history
Layer-3 review: no blockers. Confirmed the 403 fires on exactly the non-member branch (no spurious 403 for members, no surviving wrong-org fallback), inactive-membership → 403 is correct, 401 still precedes 403 for unauthenticated callers, and the opt-out can't re-enable a silent wrong-org resolution. One noted latent concern (post-`perform_create` re-resolve could raise after commit) is unreachable on the strict path — `initial()` rejects before `perform_create` — and is carried as a Phase 5 tracking note for when multi-org create becomes reachable.

## Test plan
- Asserts: non-member org header → 403 (real `CalendarViewSet`); inactive-membership org header → 403; member org header → 200; opt-out view + non-member header → not 403 (resolves gated); unauthenticated → 401 (precedes 403).
- Outer gate fully green: ruff, format, mypy (baseline), `makemigrations --check`, `check --deploy`, `pytest -n auto` (1583 passed, 0 failures).

### DIFF
--- a/common/utils/view_utils.py
+++ b/common/utils/view_utils.py
@@ -7,7 +7,7 @@ from django.shortcuts import get_object_or_404
 
 import django_virtual_models as v
 from rest_framework import generics, mixins, status
-from rest_framework.exceptions import ValidationError
+from rest_framework.exceptions import PermissionDenied, ValidationError
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet, ViewSetMixin
@@ -38,7 +38,7 @@ class TenantScopedViewMixin:
       reads this stash so all ~60 existing call sites are header-aware without
       change.
 
-    Resolution table (Phase 2b implements the multi-org-no-header 400; 2c stubbed):
+    Resolution table (Phase 2b: multi-org-no-header 400; Phase 2c: non-member 403):
 
     +-----------------------+---------------------------------+------------------------------------------+
     | Memberships (active)  | Header                          | Result                                   |
@@ -48,7 +48,7 @@ class TenantScopedViewMixin:
     | 1                     | present, matches                | resolve to it                            |
     | 2+                    | present, matches member         | resolve to named org                     |
     | 2+                    | absent                          | **400** (X-Organization-Id required)     |
-    | any                   | present, non-member org         | fallback to first [Phase 2c → 403]       |
+    | any                   | present, non-member org         | **403** (PermissionDenied)               |
     | any                   | present, non-integer            | treated as absent header                 |
     |                       |                                 | (1 → resolve · 2+ → 400 · 0 → gated)     |
     +-----------------------+---------------------------------+------------------------------------------+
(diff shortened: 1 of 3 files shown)


common/utils/view_utils.py
```diff
@@ -133,22 +135,25 @@ def _resolve_active_organization(self, request: Request) -> None:  # noqa: C901
...
+                        logger.debug(
+                            "X-Organization-Id header '%s' did not match any active membership for "
+                            "user %s; raising PermissionDenied (403).",
+                            org_id_header,
+                            user.pk,  # type: ignore[union-attr]
+                        )
+                        raise PermissionDenied(
+                            "X-Organization-Id header names an organization you are not an "
+                            "active member of."
                         )
```
hugobessa, line 153: Closes the resolution table: a header for an org you're not an active member of is a hard 403, not a silent fall-back to your first org. The old fallback was a tenant-ambiguity risk — a client bug (wrong id) would have quietly served different-org data. Note this also 403s when you DO have a membership in the named org but it's inactive (the `matching` lookup filters is_active=True), which is the intended "deactivated = no access" semantics.

common/utils/view_utils.py
```diff
@@ -133,22 +135,25 @@ def _resolve_active_organization(self, request: Request) -> None:  # noqa: C901
...
+                    # it, or the membership exists but is inactive).  Raise 403
+                    # unless the concrete view opted out of strict resolution
+                    # (active_org_resolution_optional = True).
+                    if not getattr(self, "active_org_resolution_optional", False):
+                        logger.debug(
+                            "X-Organization-Id header '%s' did not match any active membership for "
+                            "user %s; raising PermissionDenied (403).",
+                            org_id_header,
+                            user.pk,  # type: ignore[union-attr]
+                        )
```
hugobessa, line 149: The opt-out now short-circuits both the 2b 400 and this 403 → resolves to gated (None). That's what lets Phase 3's GET /organizations/mine/ accept a multi-org caller who sent a stale/foreign header without erroring — it just lists their memberships. Critically, this path leaves resolved_membership = None and does NOT re-enter the absent-header branch, so opt-out can never silently resolve to the wrong org.

PR #72, opened 2026-06-13
refactor(calendar): extract event CRUD into CalendarEventService

hugobessa:
## Summary
Phase 2 — the highest fan-in phase. Moves the 11 event methods (`create_event`, `update_event`, `delete_event`, `create_recurring_event`, `create_recurring_event_exception`, `get_recurring_event_instances`, `get_calendar_events_expanded`, `transfer_event`, `create_recurring_event_bulk_modification`, `modify_recurring_event_from_date`, `cancel_recurring_event_from_date`) out of the `CalendarService` monolith into a new `CalendarEventService` (`calendar_integration/services/calendar_event_service.py`). The facade delegates each via `self._get_event_service().<method>(...)`. `calendar_service.py` shrinks ~1000 lines.

Method bodies moved byte-for-byte (only the mechanical `self.<x>` → `self._context.<x>` / `self._host.<x>` substitutions + the Phase-1 guard-cast pattern). `@transaction.atomic()` moved with `create_event`/`update_event`/`delete_event`.

## The host seam (why a back-reference, not a pure snapshot)
`CalendarEventService` reaches the not-yet-extracted collaborators through an `EventServiceHost` Protocol (the facade is passed as `host=self`): bundle fan-out (`_*_bundle_event`, **Phase 3**), availability (`get_availability_windows_in_range`, **Phase 4**), and the shared `_get_write_adapter_for_calendar` / `_grant_event_attendee_permissions` helpers. This was chosen over a pure context-snapshot because the existing test suite patches these helpers **on the facade instance** and mutates `service.account`/`calendar_adapter` after `authenticate()` — a frozen one-time snapshot would have broken 6 existing tests (no assertion edits were made; the seam preserves the patch points). Phase 3/4 swap concrete sub-services in for the host methods without touching `CalendarEventService` call sites.

## Performance guardrail holds
`_get_event_service()` rebuilds a `CalendarServiceContext` from live facade attributes each call — a cheap dataclass construction, **no re-authentication / no adapter reconstruction / no token refresh**. The shared `_calendar_cache` dict + the `RecurrenceManager` are passed by reference, preserving cache continuity and the Phase-0 org-scoped-cache fix.

## Behavior preservation
Byte-for-byte. Reviewer confirmed the moved bodies via normalized diff (only ruff line-wrapping + the sanctioned guard substitution differ), `@transaction.atomic()` parity, and that `transfer_event` routing through `host.create_event`/`host.delete_event` matches the original `self.create_event`/`self.delete_event` call graph.

## No feature flag
Pure refactor.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 2.

## Review history
Layer-3 adversarial review: 0 blockers. One should-fix addressed in-phase: the initial extraction introduced ~30 new mypy `union-attr` errors (the guard narrowed a throwaway `cast(...)` argument, not the bare local, so `self._context.<field>` reads stayed `| None`). Fixed by narrowing a bare `context = cast("BaseCalendarService", self._context)` local through the existing `TypeGuard`s — runtime-identical, file mypy 50→20 (the remaining 20 match pre-existing facade patterns, incl. the deliberately-preserved latent bugs). NITs deferred to Phase 7: redundant facade-level `@transaction.atomic()` on the now one-line delegations; verbatim bare-`raise` guard lines.

## Test plan
- New `calendar_integration/tests/services/test_calendar_event_service.py` — constructs `CalendarEventService` directly against a context + fake adapter, asserts real DB create/update/delete/transfer + a recurring path + delegation.
- Outer gate (rebased onto current `main`, which now carries the SMS-adapter fix): `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1579 passed, 0 failed**.

### DIFF
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -1,0 +1,1412 @@
+"""Single + recurring calendar-event CRUD, transfer, and expansion reads.
+
+``CalendarEventService`` owns the event concern extracted from the
+``CalendarService`` facade. It is a plain class (not a DI-container provider):
+the facade constructs it once, *after* authentication, feeding it the shared
+:class:`CalendarServiceContext` so it never re-authenticates or re-builds a
+calendar adapter (the perf guardrail). Everything it needs arrives via the
+constructor:
+
+- ``context`` — the immutable auth snapshot (organization, user_or_token,
+  account, calendar_adapter, permission_service, side_effects_service). Read
+  through ``self._context``; the auth guards in ``type_guards.py`` inspect the
+  same ``organization`` / ``account`` / ``calendar_adapter`` attributes the
+  context exposes, so behavior is byte-for-byte identical to the former methods.
+- ``recurrence_manager`` — the stateless :class:`RecurrenceManager` the facade
+  also holds; the recurrence event methods delegate to it.
+- ``calendar_cache`` — the facade-owned, per-instance ``{(org_id, id): Calendar}``
+  cache (the lru_cache multi-tenant fix from Phase 0). Shared so lookups are not
+  duplicated across the facade and this service.
+- ``host`` — the :class:`EventServiceHost` (in Phase 2 the facade itself). The event
+  concern routes three things back through it: availability queries
+  (``get_availability_windows_in_range`` — extracted in Phase 4), bundle-event
+  fan-out (``_create_bundle_event`` / ``_update_bundle_event`` /
+  ``_delete_bundle_event`` — extracted in Phase 3), and the shared write-adapter /
+  attendee-permission helpers (``_get_write_adapter_for_calendar`` /
+  ``_grant_event_attendee_permissions``) that stay resident on the facade because the
+  sync/availability flows and the existing test suite reference them there. Routing
+  through the host keeps those single implementations and behavior byte-for-byte;
+  later phases swap concrete sub-services in without touching this service.
+"""
+
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING, Any, Protocol, cast
+
+from django.core.exceptions import PermissionDenied
+from django.db import transaction
+from django.db.models import Q
+
+from calendar_integration.constants import CalendarType
+from calendar_integration.exceptions import NoAvailableTimeWindowsError
+from calendar_integration.models import (
+    Calendar,
+    CalendarEvent,
+    CalendarManagementToken,
+    EventAttendance,
+    EventBulkModification,
+    EventExternalAttendance,
+    EventRecurrenceException,
+    ExternalAttendee,
+    RecurrenceRule,
+    RecurringMixin,
+    ResourceAllocation,
+)
+from calendar_integration.services.calendar_service_utils import (
+    convert_naive_utc_datetime_to_timezone as _convert_naive_utc_datetime_to_timezone,
+)
+from calendar_integration.services.calendar_service_utils import (
+    get_calendar_by_id as _get_calendar_by_id_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event as _serialize_event_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_data_input as _serialize_event_data_input_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_external_attendee as _serialize_event_external_attendee_util,
+)
+from calendar_integration.services.calendar_service_utils import (
+    serialize_event_internal_attendee as _serialize_event_internal_attendee_util,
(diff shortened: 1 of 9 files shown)


calendar_integration/services/calendar_event_service.py
```diff
@@ -0,0 +1,1407 @@
...
+class EventServiceHost(Protocol):
+    """The collaborator surface the event concern still routes back to the facade for.
+
+    Three concerns are not extracted in Phase 2 and stay on the facade:
+
+    - **availability** (``get_availability_windows_in_range``) — extracted in Phase 4;
+    - **bundle fan-out** (``_create_bundle_event`` / ``_update_bundle_event`` /
+      ``_delete_bundle_event``) — extracted in Phase 3;
+    - **write-adapter resolution + attendee-permission granting**
+      (``_get_write_adapter_for_calendar`` / ``_grant_event_attendee_permissions``) —
```
hugobessa, line 116: The `EventServiceHost` Protocol is the deliberate seam for the not-yet-extracted collaborators. The facade passes itself as `host=self`; the service calls `self._host._create_bundle_event(...)` (Phase 3 → `CalendarBundleService`), `self._host.get_availability_windows_in_range(...)` (Phase 4 → `AvailabilityService`), and the shared write-adapter / permission helpers. Chosen over a frozen context-only snapshot because the existing suite patches these on the facade instance and mutates `account`/`adapter` after `authenticate()` — a snapshot would have broken 6 tests without editing them.

calendar_integration/services/calendar_service.py
```diff
@@ -382,6 +372,41 @@ def initialize_without_provider(
...
+        (``host=self``). The context is rebuilt from the live facade attributes each
+        call (a cheap dataclass construction — no network / adapter rebuild), so it
+        never goes stale across re-authentication or direct attribute mutation.
+        """
+        return CalendarEventService(
+            context=self._build_context_snapshot(),
+            recurrence_manager=self._recurrence_manager,
+            calendar_cache=self._calendar_cache,
+            host=self,
+        )
```
hugobessa, line 408: Rebuilds a fresh `CalendarEventService` per call with a context snapshot from live facade attributes. This looks like it re-does work, but it's a cheap frozen-dataclass construction — NO re-authentication, NO adapter rebuild, NO token refresh — so the plan's "authenticate once" performance guardrail holds. The `_calendar_cache` and `_recurrence_manager` are shared by reference, so the Phase-0 org-scoped cache and Phase-1 recurrence engine are reused, not duplicated. Per-call rebuild (vs caching the sub-service) is what keeps the facade's post-auth attribute mutations visible — required for byte-for-byte.

calendar_integration/services/calendar_event_service.py
```diff
@@ -0,0 +1,1407 @@
...
+                    resource_id=r.calendar_fk_id,
+                )
+                for r in event.resource_allocations.all()
+                if r.calendar_fk_id
+            ],
+        )
+        # Route create/delete through the host so the facade's public methods (the
+        # original ``self.create_event`` / ``self.delete_event`` call targets) run —
+        # preserving the facade-level call graph the existing transfer tests assert on.
+        new_event = self._host.create_event(new_calendar.id, new_event_data)
```
hugobessa, line 1221: `transfer_event` routes the new-calendar create + old-event delete back through `self._host` (the facade's public `create_event`/`delete_event`). The original `transfer_event` called `self.create_event`/`self.delete_event`, so going through the host preserves the exact call graph — including the facade-level patches the transfer tests rely on. Don't "optimize" this into a direct `self.create_event` on the service; that would bypass the host and diverge from the original behavior.

calendar_integration/services/calendar_event_service.py
```diff
@@ -0,0 +1,1407 @@
...
+
+    @transaction.atomic()
+    def create_event(self, calendar_id: int, event_data: CalendarEventInputData) -> CalendarEvent:
+        """
+        Create a new event in the calendar.
+        :param calendar_id: Internal ID of the calendar
+        :param event_data: Dictionary containing event details.
+        :return: Response from the calendar client.
+        """
+        context = cast("BaseCalendarService", self._context)
```
hugobessa, line 234: The mypy-narrowing pattern (applied in the 6 guarded methods that read context fields): bind `context = cast("BaseCalendarService", self._context)` then run the `TypeGuard` on that bare local, so subsequent `context.calendar_adapter` / `.account` reads narrow to the authenticated protocol. A `cast(...)` passed directly as the guard argument does NOT narrow (TypeGuard narrows only its exact argument expression). `context is self._context` at runtime, so behavior is unchanged.

PR #73, opened 2026-06-13
feat(organizations): add GET /organizations/mine/ membership list

hugobessa:
## Summary
Phase 3 of the **Multi-Organization Membership** plan. Adds the org-switcher discovery endpoint:
`GET /organizations/mine/` returns the caller's active memberships so the frontend knows which
`X-Organization-Id` values it can send. No header required — this is what a multi-org user calls
*before* they have an id to put in the header.

## What changed
- `GET /organizations/mine/` (`mine` action on `OrganizationViewSet`): returns a bare JSON array of `{organization: {id, name}, role}` for the caller's active memberships. Gated user → `200 []`. Inactive memberships excluded. `permission_classes=[IsAuthenticated]` (so a normal member isn't blocked by the onboarding-only `OrganizationManagementPermission`).
- **Per-action opt-out** added to `TenantScopedViewMixin`: `active_org_optional_actions: tuple[str, ...]`. The resolver waives the 400/403 when `self.action` is in that tuple. `OrganizationViewSet` sets `active_org_optional_actions = ("mine",)` — so `mine` works header-less while `current`/`update`/`sync-rooms` keep full enforcement.
- New lightweight `OrganizationBriefSerializer` (`{id, name}`) + `MyMembershipSerializer` — avoids the heavier `OrganizationSerializer` (which loads the Google service account).
- New `OrganizationMembership.objects.active_for_user(user)` queryset/manager method (replaces an inline view query, per convention).
- `pagination_class=None` on the action so the response is a bare array; `schema.yml` regenerated to match (no `count/next/previous/results` envelope).

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 3 + API Design (`GET /organizations/mine/`). Use-case 5.

## Review history
Layer-3 review caught one BLOCKER: drf-spectacular inferred a paginated envelope (global `LimitOffsetPagination` + `many=True`) while the view returned a bare list — a generated client would read `.results` and break. Fixed by `pagination_class=None` + schema regen, with a test asserting the response is a bare array. Also added: a per-action-opt-out isolation test (multi-org + no header on `current` still 400), and moved the inline query to a manager method.

## Test plan
- `uv run pytest organizations/tests/test_views.py -k "Mine or mine" -vs`
- Asserts: multi-org user + no header → 200 listing both memberships (bare array); inactive excluded; gated → `200 []`; stale/foreign header on `mine` → still 200 (per-action opt-out); role reported correctly; `current` still 400 for multi-org no-header (opt-out isolation).
- Outer gate fully green: ruff, format, mypy (baseline), `makemigrations --check`, `check --deploy`, `backend-schema-local` hook, `pytest -n auto` (1594 passed, 0 failures).

### DIFF
--- a/organizations/serializers.py
+++ b/organizations/serializers.py
@@ -238,6 +238,36 @@ class CurrentMembershipSerializer(serializers.ModelSerializer):
         return OrganizationSerializer(obj.organization, context=self.context).data  # type: ignore[call-arg]
 
 
+class OrganizationBriefSerializer(serializers.ModelSerializer):
+    """Lightweight read-only serializer for an Organization.
+
+    Exposes only the fields needed for the org-switcher list: ``id`` and ``name``.
+    Intentionally avoids the heavier ``OrganizationSerializer`` (which loads the
+    Google service account) to keep ``GET /organizations/mine/`` fast.
+    """
+
+    class Meta:
+        model = Organization
+        fields = ("id", "name")
+        read_only_fields = ("id", "name")
+
+
+class MyMembershipSerializer(serializers.ModelSerializer):
+    """Read-only serializer for the caller's active organization memberships.
+
+    Used by ``GET /organizations/mine/`` to power the frontend org switcher.
+    Returns a list of ``{organization: {id, name}, role}`` entries — one per
+    active membership — without requiring the ``X-Organization-Id`` header.
+    """
+
+    organization = OrganizationBriefSerializer(read_only=True)
+
+    class Meta:
+        model = OrganizationMembership
+        fields = ("organization", "role")
+        read_only_fields = ("organization", "role")
+
+
 class OrganizationMembershipSerializer(serializers.ModelSerializer):
     """
     Read-only serializer for listing and retrieving organization members.
--- a/organizations/views.py
+++ b/organizations/views.py
@@ -47,6 +47,7 @@ from organizations.serializers import (
     AcceptInvitationSerializer,
     CurrentMembershipSerializer,
     GoogleServiceAccountWriteSerializer,
+    MyMembershipSerializer,
     OrganizationInvitationSerializer,
     OrganizationMembershipSerializer,
     OrganizationSerializer,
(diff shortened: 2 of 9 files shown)


organizations/views.py
```diff
@@ -67,6 +68,11 @@ class OrganizationViewSet(NoListVintaScheduleModelViewSet):
     queryset = Organization.objects.all()
     serializer_class = OrganizationSerializer
     permission_classes = (IsAuthenticated, OrganizationManagementPermission)
+    #: The ``mine`` action lists the caller's own memberships and must not require
+    #: the ``X-Organization-Id`` header — it is the endpoint the frontend uses to
+    #: *discover* which org ids are available.  All other actions on this viewset
+    #: keep the standard header enforcement (400 / 403).
+    active_org_optional_actions = ("mine",)
```
hugobessa, line 75: Per-action opt-out, not class-level. OrganizationViewSet has other actions (current, update, sync-rooms) that MUST keep the 400/403 header enforcement — only `mine` waives it, because it's the endpoint the client hits before it knows any org id. The resolver checks `self.action in active_org_optional_actions`. `test_current_still_returns_400_for_multi_org_user_without_header` locks the isolation so a future action added to this tuple can't silently weaken `current`.

organizations/views.py
```diff
@@ -206,6 +212,31 @@ def current(self, request):
...
+        },
+    )
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="mine",
+        permission_classes=[IsAuthenticated],
+        pagination_class=None,  # bare list — no count/next/previous envelope
+    )
+    def mine(self, request):
```
hugobessa, line 228: `pagination_class=None` is load-bearing for the API contract, not a style choice. The global default is LimitOffsetPagination, which made drf-spectacular publish a paginated `{count, next, previous, results}` schema for this `many=True` action while the view actually returns a bare list — a generated client would read `.results` and break. Disabling pagination here makes the regenerated schema a plain array, matching the runtime (`200 []` for gated users).

organizations/serializers.py
```diff
@@ -238,6 +238,36 @@ def get_organization(self, obj: OrganizationMembership) -> dict:
...
+    Google service account) to keep ``GET /organizations/mine/`` fast.
+    """
+
+    class Meta:
+        model = Organization
+        fields = ("id", "name")
+        read_only_fields = ("id", "name")
+
+
+class MyMembershipSerializer(serializers.ModelSerializer):
```
hugobessa, line 255: Deliberately a new lightweight serializer rather than reusing CurrentMembershipSerializer: that one nests the full OrganizationSerializer, which issues an extra query for the Google service account and needs request/DI context. The switcher list only needs {id, name, role}, and `active_for_user` does `select_related("organization")` so the whole list is one query.

PR #74, opened 2026-06-13
refactor(calendar): extract bundle calendar + bundle-event handling into CalendarBundleService

hugobessa:
## Summary
Phase 3. Moves the 7 bundle methods (`create_bundle_calendar`, `update_bundle_calendar`, `create_bundle_event`, `update_bundle_event`, `delete_bundle_event`, `_get_primary_calendar`, `_collect_bundle_attendees`) out of the `CalendarService` monolith into a new `CalendarBundleService` (`calendar_integration/services/calendar_bundle_service.py`). The facade's public `create_bundle_calendar`/`update_bundle_calendar` delegate to it, and the Phase-2 `EventServiceHost` bundle methods (`_create_bundle_event`/`_update_bundle_event`/`_delete_bundle_event`) now route here. `calendar_service.py` shrinks ~312 lines.

## Seam + fan-out wiring
Mirrors Phase 2: a `BundleServiceHost` Protocol declares what the bundle concern routes back to the facade (`create_event`/`update_event`/`delete_event` for child + primary representations, `get_availability_windows_in_range`, the shared write-adapter / permission / timezone helpers). The facade passes `host=self`. `_get_bundle_service()` rebuilds a context snapshot per call (cheap; no re-auth), shares `_calendar_cache` + `RecurrenceManager` by reference. The event↔bundle cycle is mediated by the facade (each `_get_*_service()` builds fresh) — no construction cycle, and children are validated non-BUNDLE so `host.create_event` takes the normal path (no recursive fan-out).

## Behavior preservation
Byte-for-byte. Reviewer diffed all 7 bodies against the original: child-reconciliation set math in `update_bundle_calendar` (`desired_ids`/`existing_ids` add/remove), `_get_primary_calendar` selection, and `_collect_bundle_attendees` dedup are identical; `@transaction.atomic()` preserved on `create_bundle_calendar`/`update_bundle_calendar`; host-routing for child events matches the original `self.create_event` call graph. Two dropped `if not self.organization: raise ValueError(...)` blocks were dead code (the `is_initialized_or_authenticated_calendar_service` guard already raises `CalendarServiceOrganizationNotSetError` first).

## Mypy
No new context-field `union-attr` (the Phase-2 narrowing pattern was reused). The 3 remaining `union-attr` (`bundle_event.calendar.id` etc., `Any | None` model relations) were moved verbatim from the facade; project net 141→138.

## Deferred to Phase 7
The facade pass-throughs keep `@transaction.atomic()` while delegating to the equally-atomic service methods (a harmless nested savepoint). Dropping the now-redundant facade decorators is part of Phase 7's facade-shrink (same as Phase 2's delegations).

## No feature flag
Pure refactor.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 3.

## Review history
Layer-3 adversarial review: 0 blockers, 0 should-fix. NITs (nested atomic → Phase 7; an intentionally-corrected docstring) left as noted.

## Test plan
- New `calendar_integration/tests/services/test_calendar_bundle_service.py` (29 tests) — asserts real DB bundle state (relationship add/remove, primary selection, representations) + fan-out create/update/delete.
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1608 passed, 0 failed**.

### DIFF
--- a/calendar_integration/services/calendar_bundle_service.py
+++ b/calendar_integration/services/calendar_bundle_service.py
@@ -1,0 +1,545 @@
+"""Bundle calendar CRUD and bundle-event fan-out logic.
+
+``CalendarBundleService`` owns the bundle concern extracted from the
+``CalendarService`` facade. It is a plain class (not a DI-container provider):
+the facade constructs it, after authentication, feeding it the shared
+:class:`CalendarServiceContext` so it never re-authenticates or re-builds a
+calendar adapter (the perf guardrail). Everything it needs arrives via the
+constructor:
+
+- ``context`` — the immutable auth snapshot (organization, user_or_token,
+  account, calendar_adapter, permission_service, side_effects_service). Read
+  through ``self._context``; the auth guards in ``type_guards.py`` inspect the
+  same ``organization`` / ``account`` / ``calendar_adapter`` attributes the
+  context exposes so behavior is byte-for-byte identical to the former methods.
+- ``host`` — the :class:`BundleServiceHost` (in Phase 3 the facade itself).
+  The bundle concern routes three things back through it:
+
+  - **availability** (``get_availability_windows_in_range``) — not extracted
+    until Phase 4, stays on the facade.
+  - **event CRUD** (``create_event`` / ``update_event`` / ``delete_event``) —
+    extracted in Phase 2 but still reached via the facade's public surface so
+    the call graph the existing test suite asserts on is preserved.
+  - **timezone conversion** (``convert_naive_utc_datetime_to_timezone``) —
+    extracted as a util in Phase 0 but delegated through the host for
+    byte-for-byte consistency with the original call pattern in blocked-time
+    updates.
+
+Routing through the host keeps single implementations and behaviour
+byte-for-byte; later phases swap concerns in without touching this service.
+
+The facade's ``_create_bundle_event`` / ``_update_bundle_event`` /
+``_delete_bundle_event`` methods become one-line delegations to this service.
+``CalendarEventService`` already calls those methods through the
+``EventServiceHost`` protocol — that call path is unchanged.
+
+Recursion guard
+---------------
+``CalendarEventService.create_event`` on a BUNDLE calendar calls
+``host._create_bundle_event`` → ``CalendarBundleService.create_bundle_event``
+→ calls ``host.create_event`` per child calendar. Children are never BUNDLE
+calendars (enforced by ``create_bundle_calendar`` / ``update_bundle_calendar``),
+so the child create_event calls take the normal (non-bundle) path. No loop.
+"""
+
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING, Protocol, cast
+
+from django.db import transaction
+
+from calendar_integration.constants import CalendarProvider, CalendarType
+from calendar_integration.models import (
+    BlockedTime,
+    Calendar,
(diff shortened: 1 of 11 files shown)


calendar_integration/services/calendar_bundle_service.py
```diff
@@ -0,0 +1,545 @@
...
+
+class BundleServiceHost(Protocol):
+    """The collaborator surface the bundle concern routes back to the facade for.
+
+    Three concerns are not extracted in Phase 3 and stay on the facade:
+
+    - **availability** (``get_availability_windows_in_range``) — extracted in Phase 4;
+    - **event CRUD** (``create_event`` / ``update_event`` / ``delete_event``) —
+      the public facade methods; reaching them through the host keeps the call
+      graph the existing test suite patches via the facade;
```
hugobessa, line 90: `BundleServiceHost` mirrors Phase 2's `EventServiceHost`: the bundle concern routes child + primary event creation back through the facade's public `create_event`/`update_event`/`delete_event` (so the fan-out goes through the same path as a normal event create), plus the shared availability / write-adapter / permission / timezone helpers. The facade passes `host=self`. A distinct protocol (not extending EventServiceHost) keeps each concern's surface minimal.

calendar_integration/services/calendar_service.py
```diff
@@ -407,6 +407,20 @@ def _get_event_service(self) -> CalendarEventService:
...
+        """Return the bundle sub-service bound to the facade's current auth context.
+
+        The bundle service routes availability (Phase 4), event CRUD, and the
+        shared write-adapter / permission helpers back through the facade
+        (``host=self``). The context is rebuilt from the live facade attributes each
+        call (a cheap dataclass construction — no network / adapter rebuild), so it
+        never goes stale across re-authentication or direct attribute mutation.
+        """
+        return CalendarBundleService(
+            context=self._build_context_snapshot(),
```
hugobessa, line 420: Same per-call-rebuild pattern as `_get_event_service()`. The event↔bundle relationship is a cycle by nature (an event on a bundle calendar fans out to children; a bundle creates child events) — it's kept safe because the facade mediates both directions, each `_get_*_service()` builds a fresh instance per call (no shared mutable construction state), and children are always non-BUNDLE calendars so `host.create_event` never re-enters bundle fan-out.

calendar_integration/services/calendar_bundle_service.py
```diff
@@ -0,0 +1,545 @@
...
+        # Remove dropped children
+        for rel in existing_relationships:
+            if rel.child_calendar_fk_id not in desired_ids:
+                rel.delete()
+
+        # Add new children
+        for calendar in child_calendars_list:
+            if calendar.id not in existing_ids:
+                is_primary = primary_calendar is not None and calendar.id == primary_calendar.id
+                ChildrenCalendarRelationship.objects.create(
```
hugobessa, line 276: The bundle invariant most worth not breaking: child reconciliation in `update_bundle_calendar`. `desired_ids`/`existing_ids` set math is byte-for-byte from the original — drop a child when its relationship id is `not in desired_ids`, add one when `calendar.id not in existing_ids`. A flipped set difference here silently corrupts bundle membership and could pass an undertested suite, so it was diffed explicitly against the original.

PR #75, opened 2026-06-13
refactor(calendar): extract availability + blocked/available times into AvailabilityService

hugobessa:
## Summary
Phase 4. Moves the 17 availability/blocked-time/available-time methods (the availability + unavailability window queries, `create_blocked_time`/`create_available_time`, the `*_expanded` reads, `bulk_create_availability_windows`, `batch_modify_available_times`, and the 8 recurring blocked/available exception + bulk-mod + modify/cancel methods) plus the interval-math privates (`_subtract_busy_intervals`, `_remove_available_time_windows_that_overlap_with_blocked_times_and_events`) into a new `AvailabilityService` (`calendar_integration/services/availability_service.py`). The facade delegates via `self._get_availability_service()`. `calendar_service.py` drops to **2557 lines** (from 4726 originally — nearly halved across Phases 0–4).

Availability + blocked-times live in one service because `get_availability_windows_in_range` is literally "all time minus blocked-times minus events" — they share the interval math and recurrence expansion.

## Seam
Mirrors Phases 2/3: an `AvailabilityServiceHost` Protocol; the facade passes `host=self`; the service reads events as busy intervals via `host.get_calendar_events_expanded(...)`. The `EventServiceHost`/`BundleServiceHost` `get_availability_windows_in_range` now routes through the facade to `AvailabilityService` (no infinite loop — the chain terminates at `get_calendar_events_expanded`, which never calls back into availability). Two thin facade private delegations are retained for existing tests/internal sync callers: `_subtract_busy_intervals` (as `@staticmethod` — a test calls `CalendarService._subtract_busy_intervals(...)` unbound) and `_remove_available_time_windows_that_overlap_with_blocked_times_and_events` (instance).

## Behavior preservation + performance
Byte-for-byte. Reviewer diffed the interval math and all 8 recurring methods against the original and confirmed byte-identical (modulo the established `self.X`→`context.X`/`host.X` substitutions). No added queries in loops; `select_related`/`bulk_create`/sorting/algorithmic complexity preserved — the plan's explicit availability perf guardrail holds.

## Mypy
3 new `attr-defined` (a `_remove_...` cast to `BaseCalendarService`, which doesn't declare `organization`) were fixed by casting to `InitializedOrAuthenticatedCalendarService` (declares `organization`; early-return-on-falsy-org semantics preserved). The 7 remaining errors (`Any | None` calendar/model-relation `arg-type`/`union-attr`) were verified verbatim from the original facade. Project mypy net unchanged (138, 0 new).

## Implementation note
This phase was completed across three agent sessions (two interrupted by API socket errors mid-extraction); the final result was independently Layer-3 reviewed precisely because of that, with the interval math + all 8 recurring methods diffed line-by-line against the original.

## No feature flag
Pure refactor.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 4.

## Review history
Layer-3 adversarial review: 0 blockers, 0 should-fix. NITs: a behavior-identical cast-style inconsistency in `_remove_...`; `AvailabilityService` not yet exported from `services/__init__.py` (deferred to Phase 7 per plan); one test uses a `MagicMock(spec=CalendarEvent)` for the host-injected event path (real-event path stays covered by the facade suite).

## Test plan
- New `calendar_integration/tests/services/test_availability_service.py` (11 tests) — window subtraction (full range / blocked-split / event-blocked), create blocked+available, recurring blocked-time expansion + exception, and the static interval helper.
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1619 passed, 0 failed**.

### DIFF
--- a/calendar_integration/services/availability_service.py
+++ b/calendar_integration/services/availability_service.py
@@ -1,0 +1,1095 @@
+"""Available-times, blocked-times, and availability/unavailability window reads.
+
+``AvailabilityService`` owns the availability concern extracted from the
+``CalendarService`` facade. It is a plain class (not a DI-container provider):
+the facade constructs it, after authentication, feeding it the shared
+:class:`CalendarServiceContext` so it never re-authenticates or re-builds a
+calendar adapter (the perf guardrail). Everything it needs arrives via the
+constructor:
+
+- ``context`` — the immutable auth snapshot (organization, user_or_token,
+  account, calendar_adapter, permission_service, side_effects_service). Read
+  through ``self._context``; the auth guards in ``type_guards.py`` inspect the
+  same ``organization`` / ``account`` / ``calendar_adapter`` attributes the
+  context exposes, so behavior is byte-for-byte identical to the former methods.
+- ``recurrence_manager`` — the stateless :class:`RecurrenceManager` the facade
+  also holds; the recurring blocked/available methods delegate to it exactly as
+  the former facade methods did.
+- ``host`` — the :class:`AvailabilityServiceHost` (in Phase 4 the facade
+  itself). The availability concern routes three things back through it:
+
+  - **event reads** (``get_calendar_events_expanded``) — the event concern,
+    extracted in Phase 2; reaching it via the host keeps a single
+    implementation and the call graph the existing test suite asserts on.
+  - **blocked-time creation** (``bulk_create_manual_blocked_times``) — stays
+    resident on the facade (other facade flows reference it and it is not part
+    of the availability concern's extracted surface), so ``create_blocked_time``
+    routes through the host.
+  - **recurrence-rule creation** (``_create_recurrence_rule_if_needed``) — a
+    shared facade helper used by both availability and (facade-resident)
+    blocked-time bulk creation; routed through the host so it has one
+    implementation and stays byte-for-byte.
+
+Routing through the host keeps single implementations and behavior
+byte-for-byte; later phases swap concerns in without touching this service.
+
+The interval-subtraction math (``_subtract_busy_intervals``) and the
+recurrence-expansion reads (``get_available_times_expanded`` /
+``get_blocked_times_expanded``) are moved verbatim — no added queries inside
+loops, no changed query structure, no algorithmic-complexity change.
+"""
+
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING, Any, Protocol, cast
+
+from django.db import transaction
+from django.db.models import Q
+
+from calendar_integration.constants import CalendarType
+from calendar_integration.models import (
+    AvailableTime,
+    AvailableTimeBulkModification,
+    AvailableTimeRecurrenceException,
+    BlockedTime,
+    BlockedTimeBulkModification,
+    BlockedTimeRecurrenceException,
+    Calendar,
+    CalendarEvent,
+    RecurrenceRule,
+    RecurringMixin,
+)
+from calendar_integration.services.calendar_service_utils import (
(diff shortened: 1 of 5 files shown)


calendar_integration/services/availability_service.py
```diff
@@ -0,0 +1,1095 @@
...
+    def _subtract_busy_intervals(
+        window_start: datetime.datetime,
+        window_end: datetime.datetime,
+        busy_intervals: Iterable[tuple[datetime.datetime, datetime.datetime]],
+    ) -> list[tuple[datetime.datetime, datetime.datetime]]:
+        """Return the parts of [window_start, window_end] not covered by any busy interval.
+
+        Busy intervals may be unsorted, overlapping, or extend beyond the window; they
+        are clipped to the window and merged on the fly. A window fully covered by busy
+        time yields an empty list.
```
hugobessa, line 210: The availability perf guardrail lives here. `_subtract_busy_intervals` (and the overlap-removal + window-range methods above/below) were moved byte-for-byte — same clip/sort/merge, no added queries in loops, `select_related`/`bulk_create` preserved. The reviewer diffed these against the original line-by-line because a subtle change here is a silent availability-correctness or O(n²) regression. Kept as a `@staticmethod` (the facade also exposes a `@staticmethod` delegation since a unit test calls `CalendarService._subtract_busy_intervals(...)` unbound on the class).

calendar_integration/services/availability_service.py
```diff
@@ -0,0 +1,1095 @@
...
+    ) -> Iterable[AvailableTimeWindow]:
+        """
+        Retrieve availability windows for a calendar within a specified date range.
+        :param calendar_id: ID of the calendar to retrieve availability for.
+        :param start_datetime: Start date for the availability search.
+        :param end_datetime: End date for the availability search.
+        :return: Iterable of AvalableTimeWindow instances.
+        """
+        if not is_initialized_or_authenticated_calendar_service(
+            cast("BaseCalendarService", self._context)
```
hugobessa, line 340: `get_availability_windows_in_range` reads events as busy intervals via `self._host.get_calendar_events_expanded(...)`. This is the one cross-service read in availability, and it's loop-safe: availability → host (facade) → `CalendarEventService.get_calendar_events_expanded`, which never calls back into availability. The reverse direction (an event create asking the host for availability) terminates at the same `get_calendar_events_expanded`, so there is no cycle. This is the original call graph, just routed through the host.

calendar_integration/services/availability_service.py
```diff
@@ -0,0 +1,1095 @@
...
+        calendar_id: int,
+        blocked_times: Iterable[BlockedTime],
+        events: Iterable[CalendarEvent],
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+    ):
+        """
+        Removes AvailableTime windows that overlap with BlockedTime or CalendarEvent instances.
+        """
+        context = cast("InitializedOrAuthenticatedCalendarService", self._context)
```
hugobessa, line 161: `_remove_...` casts to `InitializedOrAuthenticatedCalendarService` (which declares `organization`) rather than the `cast("BaseCalendarService", ...)` + TypeGuard pattern the other methods use — because the original facade method did a bare `if not self.organization: return` early-return (no TypeGuard), and `BaseCalendarService` doesn't declare `organization`. This preserves the early return exactly while keeping mypy clean. The lone stylistic exception, called out so it isn't "normalized" into a guard later (that would change the return-vs-raise behavior).

PR #76, opened 2026-06-13
refactor(calendar): extract sync + import state machine into CalendarSyncService

hugobessa:
## Summary
Phase 5 — the most coupled extraction. Moves the import/sync state machine (the 6 public methods `request_calendars_import`, `import_account_calendars`, `request_calendar_sync`, `sync_events`, `request_organization_calendar_resources_import`, `import_organization_calendar_resources`, plus the diff/merge internals `_execute_calendar_sync`, `_process_events_for_sync`, `_process_existing_event`, `_process_new_event`, `_process_event_attendees`, `_process_existing_blocked_time`, `_apply_sync_changes`, `_handle_deletions_for_full_sync`, `_link_orphaned_recurring_instances`, `_get_existing_calendar_data`, `_execute_organization_calendar_resources_import`) into a new `CalendarSyncService` (`calendar_integration/services/calendar_sync_service.py`). Facade delegates via `self._get_sync_service()`. `calendar_service.py` drops to **1985 lines** (down from 4726 originally — a 58% reduction across Phases 0–5).

## Seam + how sync materializes events
Mirrors Phases 2–4: a `SyncServiceHost` Protocol; the facade passes `host=self`; `_get_sync_service()` rebuilds context per call, shares `_calendar_cache`. Sync reads from the provider via `context.calendar_adapter.get_events(...)` / `get_available_calendar_resources(...)` and writes `BlockedTime`/`CalendarEvent`/`RecurrenceRule` rows via the same bulk operations. It routes back to the facade host for the availability overlap-removal (Phase 4), `_grant_calendar_owner_permissions`, and `request_calendar_sync` / `_execute_organization_calendar_resources_import` (so existing facade-patch tests still intercept).

## Behavior preservation
Byte-for-byte. Reviewer diffed the diff/merge machinery line-by-line and confirmed byte-identical — every create/update/delete decision, every `bulk_create`/`bulk_update`/`delete` (same field lists, e.g. `["start_time_tz_unaware","end_time_tz_unaware","reason","external_id"]`), every query and ordering, and the orphan-linking logic unchanged. External callers (Celery `calendar_sync_tasks.py`, `organizations/services.py`) are unaffected — facade public signatures unchanged; the DI container + tasks untouched.

## Host self-routing is non-recursive (the one genuinely new wiring)
Some sync methods route an inner call through `self._host` so existing patch-based tests intercept. Verified safe: each host round-trip lands on a *different* terminating method (e.g. `import_account_calendars → host.request_calendar_sync → fresh service.request_calendar_sync` which only creates a `CalendarSync` + queues a task — never calls host). No method routes back to itself; the fresh service is value-identical to the unpatched original `self.<x>` call.

## Mypy
7 errors in the new file, each verbatim from the original facade (`account.id`/`account.user` union-attr, etc.); the facade shed those same 10 sync-body errors. No new category.

## Deferred to Phase 7
Redundant facade-level `@transaction.atomic()` on the `import_account_calendars` delegation; stale `_process_events_for_sync` / `_link_orphaned_recurring_instances` declarations on the `AuthenticatedCalendarService` protocol (facade no longer implements them). A direct orphan-link unit test is a noted follow-up — the orphan path stays covered by the `test_calendar_sync_tasks.py` integration gate.

## No feature flag
Pure refactor.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 5.

## Review history
Layer-3 adversarial review: 0 blockers, 0 should-fix. Implemented in two commits (extraction checkpoint, then tests) to survive interruption. NITs deferred to Phase 7 as above.

## Test plan
- New `calendar_integration/tests/services/test_calendar_sync_service.py` (3 tests) — a full sync diff/merge cycle + import path against a fake adapter.
- Integration safety net: `calendar_integration/tests/tasks/test_calendar_sync_tasks.py` (unchanged) exercises the full facade→sync path.
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1622 passed, 0 failed**.

### DIFF
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -2,11 +2,11 @@ import datetime
 import json
 import logging
 from collections.abc import Callable, Iterable
-from typing import Annotated, Literal, TypedDict
+from typing import Annotated, TypedDict
 
 from django.conf import settings
 from django.db import transaction
-from django.db.models import Q, QuerySet
+from django.db.models import QuerySet
 from django.http import HttpRequest
 from django.urls import reverse
 
@@ -14,7 +14,6 @@ from allauth.socialaccount.models import SocialAccount, SocialToken
 from dependency_injector.wiring import Provide, inject
 
 from calendar_integration.constants import (
-    CalendarOrganizationResourceImportStatus,
     CalendarProvider,
     CalendarSyncStatus,
     CalendarSyncTriggerSource,
@@ -39,7 +38,6 @@ from calendar_integration.models import (
     CalendarWebhookSubscription,
     EventAttendance,
     EventExternalAttendance,
-    ExternalAttendee,
     GoogleCalendarServiceAccount,
     RecurrenceRule,
 )
@@ -77,6 +75,7 @@ from calendar_integration.services.calendar_service_utils import (
     serialize_event_internal_attendee as _serialize_event_internal_attendee_util,
 )
 from calendar_integration.services.calendar_side_effects_service import CalendarSideEffectsService
+from calendar_integration.services.calendar_sync_service import CalendarSyncService
 from calendar_integration.services.dataclasses import (
     ApplicationCalendarData,
     AvailableTimeWindow,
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_sync_service.py
```diff
@@ -0,0 +1,992 @@
...
+    def _process_existing_event(
+        self,
+        event: CalendarEventAdapterOutputData,
+        existing_event: CalendarEvent,
+        changes: EventsSyncChanges,
+        update_events: bool,
+    ):
+        """Process an existing calendar event."""
+        if not update_events:
+            return
```
hugobessa, line 660: The sync diff/merge is the highest-risk code in the whole refactor — a flipped condition or a changed `bulk_update` field list here silently corrupts synced calendars, and the suite might not catch an undertested edge. So `_process_existing_event` / `_process_new_event` / `_apply_sync_changes` / `_handle_deletions_for_full_sync` / `_link_orphaned_recurring_instances` were moved byte-for-byte and diffed line-by-line against the original (same field lists, same query/ordering, same create/update/delete decisions). Only the standard `self.X`→`context.X`/`host.X` substitutions differ.

calendar_integration/services/calendar_sync_service.py
```diff
@@ -0,0 +1,992 @@
...
+class SyncServiceHost(Protocol):
+    """The collaborator surface the sync concern routes back to the facade for.
+
+    Two concerns are not part of the sync concern's extracted surface and stay on
+    the facade:
+
+    - **available-time pruning**
+      (``_remove_available_time_windows_that_overlap_with_blocked_times_and_events``)
+      — the availability concern (Phase 4); reached through the host to keep one
+      implementation and the call graph the existing test suite patches via the facade;
```
hugobessa, line 100: `SyncServiceHost` routes a few inner calls (`request_calendar_sync`, `_execute_organization_calendar_resources_import`) back through the facade so the existing patch-based tests keep intercepting them. This is the one new wiring in the phase and it's self-routing, so it was checked for recursion: every host round-trip lands on a DIFFERENT terminating method (e.g. `import_account_calendars → host.request_calendar_sync → fresh service.request_calendar_sync`, which only creates a `CalendarSync` + queues a Celery task and never calls the host). No method routes back to itself.

calendar_integration/services/calendar_service.py
```diff
@@ -432,39 +431,29 @@ def _get_availability_service(self) -> AvailabilityService:
...
+        """Return the sync sub-service bound to the facade's current auth context.
+
+        The sync service shares the facade-owned calendar cache and routes
+        available-time pruning (Phase 4) and the shared owner-permission helper back
+        through the facade (``host=self``). The context is rebuilt from the live facade
+        attributes each call (a cheap dataclass construction — no network / adapter
+        rebuild), so it never goes stale across re-authentication or direct attribute
+        mutation.
+        """
+        return CalendarSyncService(
```
hugobessa, line 444: `_get_sync_service()` is the 4th `_get_*_service()` (event/bundle/availability/sync) — same per-call rebuild, shared `_calendar_cache`, `host=self`. Many sync PRIVATE methods are kept as facade delegations (not just the public ones) because `test_calendar_service.py` calls them directly on a facade instance; the two with no facade caller (`_process_events_for_sync`, `_link_orphaned_recurring_instances`) were not kept. After this phase the facade is 1985 lines (was 4726) — auth, calendar CRUD, and webhooks (Phase 6) remain.

PR #77, opened 2026-06-13
refactor(calendar): extract webhook lifecycle into CalendarWebhookService

hugobessa:
## Summary
Phase 6 — the last concern extraction. Moves the 8 webhook methods (`request_webhook_triggered_sync`, `create_calendar_webhook_subscription`, `process_webhook_notification`, `handle_webhook`, `list_webhook_subscriptions`, `delete_webhook_subscription`, `refresh_webhook_subscription`, `get_webhook_health_status`) + the `WebhookHealthStatus` TypedDict into a new `CalendarWebhookService` (`calendar_integration/services/calendar_webhook_service.py`). Facade delegates via `self._get_webhook_service()`. `calendar_service.py` drops to **1618 lines** (down from 4726 originally — a 66% reduction across Phases 0–6).

## Seam + webhook→sync routing
Mirrors Phases 2–5: a `WebhookServiceHost` Protocol; `host=self`; `_get_webhook_service()` rebuilds context per call, shares `_calendar_cache`. Webhook-triggered sync routes back through the host: `process_webhook_notification → host.request_webhook_triggered_sync → host.request_calendar_sync → CalendarSyncService` (which never calls webhooks). Verified self-routing terminates — no method routes back to itself.

## handle_webhook split (the one non-trivial deviation)
`handle_webhook` is the HTTP entrypoint. The facade RETAINS the org-extraction half (resolver_match guard, `organization_id` kwarg, `Organization.objects.get`, exact ValueError messages + `from exc` chaining) and `self.organization = organization` — because that mutation must land on the facade so `_build_context_snapshot()` includes the org when the sub-service is constructed (the downstream `request_webhook_triggered_sync → host.request_calendar_sync` path needs it). The service half does header parsing + dispatch. Reviewer confirmed the facade half is byte-identical to the original's first half and nothing was dropped or duplicated.

## Behavior preservation
Byte-for-byte. Reviewer diffed `process_webhook_notification` (security-sensitive: adapter validation / signature / token check, `WebhookIgnoredError`/`ServiceNotAuthenticatedError` handling, `IncomingWebhookProcessingStatus` transitions) — all intact. The external-id lookup uses the Phase-0 org-scoped shared cache. No tenancy leak. Removed facade imports verified truly unused.

## One regression caught + fixed in-phase
The extraction initially weakened `process_webhook_notification`'s first auth guard from raise-on-org-unset to warn-and-continue (`raise_error=False`). Restored to the original immediate-raise semantics (`CalendarServiceOrganizationNotSetError`) and added a regression test asserting it short-circuits before any lookup/validation.

## Mypy
137→136 (0 new; the duplicated facade `WebhookHealthStatus` TypedDict that's now single-sourced cleared one).

## No feature flag
Pure refactor.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 6.

## Review history
Layer-3 adversarial review: 0 blockers, 1 should-fix (the guard regression above — fixed in-phase with a covering test), NITs verbatim-preserved (a duplicated `expires_at = None` from the original). After the fix: full suite 1634 passed, 0 failed.

## Test plan
- New `calendar_integration/tests/services/test_calendar_webhook_service.py` (12 tests) — subscription create/refresh/delete, notification parse, health status, and the org-unset immediate-raise regression.
- Integration safety net: `test_google_calendar_webhooks.py`, `test_microsoft_calendar_webhooks.py`, `test_incoming_webhook_*`, webhook mgmt-command tests (all unchanged).
- Outer gate: `ruff check` clean, `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1634 passed, 0 failed**.

### DIFF
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -1,30 +1,23 @@
 import datetime
-import json
 import logging
 from collections.abc import Callable, Iterable
-from typing import Annotated, TypedDict
+from typing import Annotated
 
-from django.conf import settings
 from django.db import transaction
 from django.db.models import QuerySet
 from django.http import HttpRequest
-from django.urls import reverse
 
 from allauth.socialaccount.models import SocialAccount, SocialToken
 from dependency_injector.wiring import Provide, inject
 
 from calendar_integration.constants import (
     CalendarProvider,
-    CalendarSyncStatus,
     CalendarSyncTriggerSource,
     CalendarType,
     CalendarVisibility,
-    IncomingWebhookProcessingStatus,
 )
 from calendar_integration.exceptions import (
     InvalidCalendarTokenError,
-    ServiceNotAuthenticatedError,
-    WebhookIgnoredError,
 )
 from calendar_integration.models import (
     AvailableTime,
@@ -76,6 +69,10 @@ from calendar_integration.services.calendar_service_utils import (
 )
 from calendar_integration.services.calendar_side_effects_service import CalendarSideEffectsService
 from calendar_integration.services.calendar_sync_service import CalendarSyncService
+from calendar_integration.services.calendar_webhook_service import (
+    CalendarWebhookService,
+    WebhookHealthStatus,
+)
 from calendar_integration.services.dataclasses import (
     ApplicationCalendarData,
     AvailableTimeWindow,
@@ -101,16 +98,6 @@ from calendar_integration.services.type_guards import (
 from organizations.models import Organization
 from public_api.models import SystemUser
 from users.models import User
-
-
-class WebhookHealthStatus(TypedDict):
-    total_subscriptions: int
-    active_subscriptions: int
-    expired_subscriptions: int
-    expiring_soon_subscriptions: int
-    recent_events_count: int
-    failed_events_count: int
-    success_rate: float
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_service.py
```diff
@@ -1690,111 +1546,29 @@ def process_webhook_notification(
...
-        """
-        Handle Google Calendar webhook processing with organization context.
+        """Handle calendar webhook processing with organization context.
+
+        Extracts the organization from the HTTP request and writes it to
+        ``self.organization`` so that :meth:`_build_context_snapshot` picks it
+        up when the webhook sub-service calls back through the host
+        (``request_webhook_triggered_sync`` → ``request_calendar_sync``).
+        The header parsing and notification dispatching are delegated to
+        :class:`CalendarWebhookService`.
```
hugobessa, line 1568: `handle_webhook` is the one method NOT reduced to a pure 1-line delegation. The facade keeps the org-extraction half (byte-identical to the original) and the `self.organization = organization` mutation, then delegates the rest. That mutation has to land on the facade: `_build_context_snapshot()` reads `self.organization`, so the org must be set BEFORE `_get_webhook_service()` builds the sub-service — otherwise the downstream `request_webhook_triggered_sync → host.request_calendar_sync` wouldn't see the org. The header parsing + dispatch half moved to the service.

calendar_integration/services/calendar_webhook_service.py
```diff
@@ -0,0 +1,676 @@
...
+    def process_webhook_notification(
+        self,
+        provider: str,
+        calendar_external_id: str,
+        headers: dict[str, str],
+        payload: dict | str | None = None,
+        validation_token: str | None = None,
+    ) -> CalendarWebhookEvent | None:
+        """Process incoming webhook notification using adapter validation.
+
```
hugobessa, line 382: Security-sensitive: this validates the incoming notification via the adapter. Moved byte-for-byte (validation / signature / token dispatch, the WebhookIgnoredError / ServiceNotAuthenticatedError handling, and all IncomingWebhookProcessingStatus transitions unchanged). One regression was caught in review + fixed here: the first auth guard had been weakened to `raise_error=False` (warn-and-continue); it's restored to the original immediate-raise on org-unset, with a regression test asserting it short-circuits before any calendar lookup or validation runs.

calendar_integration/services/calendar_webhook_service.py
```diff
@@ -0,0 +1,676 @@
...
+class WebhookServiceHost(Protocol):
+    """The collaborator surface the webhook concern routes back to the facade for.
+
+    Concerns not part of the webhook surface that stay on the facade (or the facade
+    re-delegates from here):
+
+    - **sync triggering** (``request_calendar_sync``) — the sync concern (Phase 5);
+      reached through the host to keep one implementation and the call graph the
+      existing test suite patches on the facade.
```
hugobessa, line 116: Last of the 6 sub-services (event/bundle/availability/sync/webhook). After this phase the facade is 1618 lines (was 4726, −66%). `WebhookServiceHost` routes webhook-triggered sync back through the facade to the sync service; the chain terminates at the sync service, which never calls webhooks. Phase 7 shrinks the facade to delegation-only + confirms the DI wiring.

PR #78, opened 2026-06-14
refactor(calendar): shrink CalendarService facade to delegation-only and finalize wiring

hugobessa:
## Summary
Phase 7 — the final cleanup phase of the calendar-service refactor. Pure cleanup, no behavior change:
1. **Dropped 5 redundant facade `@transaction.atomic()` decorators** on now-pure-delegation methods (`import_account_calendars`, `create_bundle_calendar`, `update_bundle_calendar`, `create_event`, `update_event`) — the target sub-service method owns the single atomic block, so this is byte-for-behavior-identical (one atomic level, as in the original). Kept atomic on the 3 facade methods that still hold real ORM logic (`create_virtual_calendar`, `get_default_calendar_for_user`, `bulk_create_manual_blocked_times`).
2. **Removed 2 stale protocol declarations** (`_process_events_for_sync`, `_link_orphaned_recurring_instances`) from `AuthenticatedCalendarService` — they now live only on `CalendarSyncService` and the facade never implemented them.
3. **Populated `calendar_integration/services/__init__.py`** with the package's public surface: the `CalendarService` facade + the 6 sub-services + `CalendarServiceContext` + `RecurrenceManager`.
4. **Confirmed `di_core/containers.py` unchanged** — the `calendar_service` provider still injects only `calendar_side_effects_service` + `calendar_permission_service`; `calendar_group_service` still injects `calendar_service`. No sub-service providers added (the facade constructs them).
5. **Module docstring** on the facade documenting its delegation-only role.

## Where the refactor landed
`calendar_service.py` went from **4726 lines → ~1644 lines** (a single 4,726-line `CalendarService` class became a thin facade + 6 focused sub-services + a stateless `RecurrenceManager` + a shared utils module + a `CalendarServiceContext`). Every external call site (views, serializers, GraphQL mutations, Celery tasks, `organizations/services.py`, `public_api`, `CalendarGroupService`) is unchanged — they still inject `CalendarService` and call the same methods. The DI container is untouched.

## Behavior preservation
Byte-for-byte. Reviewer confirmed each of the 5 atomic removals is a pure delegation whose sub-service target owns its own atomic (single-transaction guarantee preserved), the protocol removals are grep-confirmed unused, and `__init__.py` introduces no import cycle. Full suite green with no assertion edits.

## No feature flag
Pure refactor. No flag was introduced across any phase → no flag-removal phase needed.

## Plan reference
Plan `ai-plans/2026-06-13-CALENDAR_SERVICE_REFACTOR_IMPLEMENTATION_PLAN.md`, Phase 7 (final).

## Review history
Layer-3 adversarial review: 0 blockers, 0 should-fix. NIT: `CalendarPermissionService`/`CalendarSideEffectsService` (pre-existing separate services, not part of this refactor's sub-services) aren't in the export list — the module docstring scopes the exports to the facade-constructed sub-services, so left as-is.

## Test plan
- No new tests (cleanup phase). Full existing suite green with NO assertion edits.
- Outer gate: `ruff check` clean, `mypy .` 136 (unchanged baseline), `check --deploy` (5 pre-existing security warnings only), `pytest -n auto` → **1634 passed, 0 failed**.


calendar_integration/services/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+from calendar_integration.services.availability_service import AvailabilityService
+from calendar_integration.services.calendar_bundle_service import CalendarBundleService
+from calendar_integration.services.calendar_event_service import CalendarEventService
+from calendar_integration.services.calendar_service import CalendarService
+from calendar_integration.services.calendar_service_context import CalendarServiceContext
+from calendar_integration.services.calendar_sync_service import CalendarSyncService
+from calendar_integration.services.calendar_webhook_service import CalendarWebhookService
+from calendar_integration.services.recurrence_manager import RecurrenceManager
+
```
hugobessa, line 21: The package's public surface after the refactor: the `CalendarService` facade (the DI-injected entry point, unchanged for every existing caller) plus the 6 sub-services it constructs after authentication, exported so they can be unit-tested in isolation. Import-clean — the facade already imported the sub-services top-level and no sub-service imports the facade at module level (context refs are under TYPE_CHECKING), so no cycle.

PR #79, opened 2026-06-14
feat(organizations): allow accepting invitations into additional organizations

hugobessa:
## Summary
Phase 4 of the **Multi-Organization Membership** plan. A user already in one org can now accept an
invitation into another org and gain a second membership. The provisioning guards that Phase 1
deliberately preserved as "refuse any second membership" are relaxed to **per-org** — refuse only a
duplicate in the SAME org.

## What changed
- `accept_invitation`: dropped the blanket "user has any membership → refuse" guard; now refuses only when the user already has a membership in the matched invitation's org. Cross-org accept creates a second membership.
- `provision_tenant_for_user` (signup/social adapter path): the **pending-invitation** branch goes per-org (join a new org via invite even when already a member elsewhere). The **`organization_name` create-new-org** branch KEEPS its single-membership guard — auto-creating an additional org on signup is **Phase 5's** concern (the `OrganizationManagementPermission` relaxation), not this phase.
- `OrganizationInvitation.email`: relaxed from global `unique=True` to `unique(email, organization)` (migration `0007`, reversible) so org A and org B can both hold a pending invite for the same email.
- `UserAlreadyHasMembershipError.default_detail` reworded to "already a member of this organization" (it now means same-org, not any-org).

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 4 + Guiding Decisions (multi-org join) + Open Questions (invitation email uniqueness, now resolved to per-org). Use-case 6.

## Review history
Layer-3 review: no blockers. Confirmed the per-org guard correctly still blocks re-accept into an org where the user has an INACTIVE membership (admin-reactivation invariant), the `organization_name` branch stays Phase-5-gated, and the migration is reversible. Four SHOULD-FIXes applied: corrected the now-misleading error message; added the missing social-signup cross-org-invite integration test (through the adapter, two memberships); added the inactive-membership-blocks-re-accept coverage; added `invite_user_to_organization` duplicate/reset + same-email-different-org coverage.

## Test plan
- `uv run pytest organizations/tests/test_services.py organizations/tests/test_views.py accounts/tests/test_social_gated_onboarding.py -vs`
- Asserts: org-A member accepts org-B invite → two memberships (incl. via social signup adapter); same-org re-accept → 400; inactive membership in inviting org still blocks re-accept; both orgs invite same email → two coexisting rows; `invite_user_to_organization` second same-org call resets (count 1, no dup); `provision_tenant_for_user` organization_name + existing membership still refused (Phase-5 boundary); migration reverse applies clean.
- Outer gate fully green: ruff, format, `makemigrations --check`, `check --deploy`, `pytest -n auto` (1604 passed, 0 failures). mypy 139 (test-only `@inject` false positive matching an existing pattern; baseline was 138).


organizations/services.py
```diff
@@ -271,28 +271,33 @@ def accept_invitation(self, token: str, user: User) -> OrganizationMembership:
...
             email__iexact=user.email, expires_at__gt=now
         )
         for invitation in invitations:
             if verify_long_lived_token(token, invitation.token_hash):
+                # Per-org guard (Phase 4): refuse only a duplicate in the SAME org.
+                # A user already in a different org is allowed to join this one.
+                if user.organization_memberships.filter(
+                    organization=invitation.organization
+                ).exists():
+                    raise UserAlreadyHasMembershipError()
```
hugobessa, line 300: The per-org relaxation. Phase 1 preserved the OneToOne-era "refuse any second membership" here; Phase 4 narrows it to the matched invitation's org. Note the guard query intentionally has NO is_active filter — an INACTIVE membership in the inviting org still blocks re-accept, preserving the invariant that the only un-disable path is admin reactivation, not re-accepting an invite. The composite unique(user, organization) is the DB backstop for the concurrent-accept race.

organizations/services.py
```diff
@@ -367,6 +378,11 @@ def provision_tenant_for_user(
             return membership
 
         if organization_name:
+            # DESIGN: auto-creating a second org for an existing member is Phase 5's
+            # concern (relaxing OrganizationManagementPermission on POST /organizations/).
+            # This signup-path branch keeps the original single-membership guard.
+            if user.organization_memberships.exists():
+                raise UserAlreadyHasMembershipError()
```
hugobessa, line 385: Deliberate phase boundary: the create-NEW-org branch keeps the blanket single-membership guard. Letting an existing member create an additional org is Phase 5 (relaxing OrganizationManagementPermission on POST /organizations/). Relaxing it here too would smear two use-cases across phases — this branch is the signup auto-provision path, not the explicit create-org API.

organizations/models.py
```diff
@@ -253,6 +258,14 @@ class OrganizationInvitation(BaseModel):
         blank=True,
     )
 
+    class Meta:
+        constraints: ClassVar = [
+            models.UniqueConstraint(
+                fields=["email", "organization"],
+                name="uniq_invitation_email_organization",
+            ),
```
hugobessa, line 266: email goes from global unique to unique(email, organization) so the same person can hold pending invites from multiple orgs at once — the prerequisite for cross-org invite accept. A duplicate invite to the SAME org is still rejected. invite_user_to_organization already get_or_creates on (email, organization), so it slots in without change; migration 0007 is reversible (reverse restores the global unique).

PR #80, opened 2026-06-14
feat(organizations): allow members to create additional organizations

hugobessa:
## Summary
Phase 5 (final) of the **Multi-Organization Membership** plan. An authenticated user who already
belongs to one or more orgs can now `POST /organizations/` and become ADMIN of the new one — the
onboarding gate previously restricted create to membership-less users.

## What changed
- `OrganizationManagementPermission.has_permission` is action-aware: `create` is allowed for any authenticated user; all other actions keep the exact prior membership-less gate. `has_object_permission` unchanged.
- `OrganizationViewSet.active_org_optional_actions = ("mine", "create")` — `create` is exempt from the strict header resolution so a multi-org member POSTing without a header isn't 400'd in `initial()` (pre-write) or by the post-write re-resolve.
- `OrganizationViewSet.create` is overridden: after `perform_create`, it looks up the just-created membership and stashes it as the active org so the response refetch scopes to the NEW org (the generic post-write re-resolve would leave a multi-org caller at gated/None and 500 the refetch). This also resolves the Phase-2c carry-forward concern: a create can no longer 400/403 after the write commits.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 5. Use-case 7. No feature flag in this plan, so this is the final phase (no flag-removal phase).

## Review history
Layer-3 review: no blockers. Confirmed unauthenticated POST still 401, no privilege escalation (creator is admin of the NEW org only), other actions' gating byte-for-byte unchanged, and the refetch reliably scopes to the new org (no `DoesNotExist`/500 path). Four fixes applied: corrected a false `create()` docstring that had leaked into the public `schema.yml` (it claimed the re-resolve "is still called" — it is skipped); mirrored the base mixin's `get_return_queryset` refetch branch; refreshed the stale `get_permissions` docstring; replaced an overstated "handles None gracefully" comment with a `logger.error` guard.

## Test plan
- `uv run pytest organizations/tests/test_views.py -k "CreateAdditional or create" -vs`
- Asserts: org-A member POSTs /organizations/ (no header) → 201, ADMIN of new org, 2 memberships, both in `mine/`, header for new org scopes to it; a 2-existing-orgs member → 201 (NOT 400) proving create is exempt and the post-write re-resolve can't 400; gated first-org create → still 201/ADMIN/1 membership; unauthenticated → 401.
- Outer gate fully green: ruff, format, `makemigrations --check`, `check --deploy`, `backend-schema-local` hook, `pytest -n auto` (1610 passed, 0 failures).


organizations/permissions.py
```diff
@@ -23,10 +23,18 @@ def has_permission(self, request, view):
         except AttributeError:
             return True
 
+        # Phase 5: any authenticated user may create an additional organisation
+        # (they become its admin via a fresh membership). Gating create to
+        # membership-less users would block the "create additional org" use-case.
+        # All other actions still require the user to have no active membership
+        # (the onboarding gate).
+        if view.action == "create":
+            return bool(user and user.is_authenticated)
```
hugobessa, line 32: Action-aware relaxation: only `create` opens to any authenticated user; every other action keeps the exact membership-less onboarding gate (the early return is before the unchanged `membership is None` path). IsAuthenticated still runs first in permission_classes, so an anonymous POST is still 401. The creator becomes admin of the NEW org only — has_object_permission is untouched and still excludes create, so no object-level access changes.

organizations/views.py
```diff
@@ -70,9 +70,17 @@ class OrganizationViewSet(NoListVintaScheduleModelViewSet):
...
+    #:
+    #: The ``create`` action is also exempt so that a member with existing
+    #: memberships can POST /organizations/ without a header (Phase 5). Without
+    #: this exemption, the multi-org 400 would fire in ``initial()`` before
+    #: ``perform_create`` runs, and the post-create re-resolve in
+    #: ``CreateModelMixin.create`` would again raise 400 (the user now has one
+    #: more membership than before the write).
+    #:
+    #: All other actions keep the standard header enforcement (400 / 403).
+    active_org_optional_actions = ("mine", "create")
```
hugobessa, line 83: `create` joins `mine` in the per-action opt-out. Without it, a member with 2+ memberships POSTing without a header would 400 twice: once in initial() (pre-write, multi-org-no-header) and again in the post-perform_create re-resolve (they now have one MORE membership). Exempting create lets the write through; the create() override below then stashes the new org for the response.

organizations/views.py
```diff
@@ -105,6 +115,68 @@ def get_queryset(self):
...
+        entirely.  For the ``create`` action — exempted via
+        ``active_org_optional_actions = ("mine", "create")`` — that re-resolve
+        would leave a multi-org caller with no ``X-Organization-Id`` header
+        resolved to ``None``, making ``get_queryset`` return nothing and causing
+        the re-fetch to raise ``DoesNotExist`` / 500.
+
+        Instead, after ``perform_create`` we look up the just-created membership
+        directly and stash it on the request so the re-fetch (via
+        ``get_queryset``) is scoped to the new organization.
+        """
```
hugobessa, line 134: Why the override exists: the base CreateModelMixin re-resolves the active org after the write to scope the response refetch. For create (now opt-out), that re-resolve yields None for a multi-org caller → org-scoped get_queryset returns nothing → DoesNotExist/500, even though the org WAS created. Instead we look up the just-created membership directly and stash it, so the refetch scopes to the new org. This is also what makes a post-write 400/403 impossible (the Phase-2c carry-forward concern).

PR #82, opened 2026-06-14
feat(openapi): document X-Organization-Id header on tenant-scoped routes

hugobessa:
## Summary
Phase 6 (appended via amend-plan) of the **Multi-Organization Membership** plan. Makes the
`X-Organization-Id` tenant-selection contract visible in the OpenAPI schema. The Phase 2a resolver
reads the header in `TenantScopedViewMixin.initial()` — which drf-spectacular never introspects — so
`schema.yml` previously declared **zero** header parameters and API consumers had no way to discover
the header or its 400/403 contract.

## What changed
- New `common/openapi.py` — `TenantScopedAutoSchema(AutoSchema)` overriding `get_override_parameters()`. For every operation whose view is a `TenantScopedViewMixin` subclass — and which is NOT opted out via `active_org_resolution_optional` or `active_org_optional_actions` — it appends an `X-Organization-Id` header `OpenApiParameter` (HEADER, `required=False`) whose description states the full contract (optional for single-membership callers; required for 2+ memberships → 400 if omitted; 403 if it names a non-member org).
- `vinta_schedule_api/settings/base.py` — `REST_FRAMEWORK["DEFAULT_SCHEMA_CLASS"]` → `"common.openapi.TenantScopedAutoSchema"`.
- `schema.yml` regenerated: `in: header` parameters went 0 → 164.

## Plan reference
Plan `ai-plans/2026-06-13-MULTI_ORG_MEMBERSHIP_IMPLEMENTATION_PLAN.md` — Phase 6 (see the **Amendments** section). Append-only amendment: no existing phase rewritten, no branch force-pushed.

## Review history
Layer-3 review: no blockers, no should-fixes. Verified the override faithfully mirrors `TenantScopedViewMixin._is_active_org_resolution_optional`, preserves per-view `@extend_schema` params, and that the global `DEFAULT_SCHEMA_CLASS` swap is inert for non-tenant views (the `isinstance` guard returns early — confirmed `public_api`, `accounts`, `users`, `payments`, `AcceptInvitationView` all still lack the header). Two NITs applied: dropped an unneeded `django_db` marker; added a positive `/organizations/current/` assertion.

## Test plan
- `uv run pytest common/tests/test_openapi.py -vs`
- Asserts (via in-process `SchemaGenerator().get_schema()`): a tenant-scoped op (`/calendar/`) and `/organizations/current/` DECLARE the header (`in: header`, required not True); the opt-out ops `/organizations/mine/` (GET) and `/organizations/` (POST) and a non-tenant op do NOT.
- Outer gate fully green: ruff, format, mypy (0 new), `makemigrations --check`, `check --deploy`, `backend-schema-local` hook, `pytest -n auto` (1616 passed, 0 failures).


common/openapi.py
```diff
@@ -0,0 +1,80 @@
...
+        params.append(
+            OpenApiParameter(
+                name=ACTIVE_ORG_HEADER,
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.HEADER,
+                required=False,
+                description=_HEADER_DESCRIPTION,
+            )
+        )
+        return params
```
hugobessa, line 80: Centralized doc injection: one AutoSchema subclass auto-covers every current + future TenantScopedViewMixin route, so no per-view @extend_schema annotation is needed. The opt-out check mirrors TenantScopedViewMixin._is_active_org_resolution_optional exactly (view-level flag then per-action tuple) so the schema can never advertise the header on a route that actually waives it (mine, create). `super().get_override_parameters()` is preserved so per-view params aren't dropped. required=False because single-membership callers may omit it.

vinta_schedule_api/settings/base.py
```diff
@@ -142,7 +142,7 @@ def base_dir_join(*args):
         "rest_framework.permissions.IsAuthenticated",
     ],
     "DEFAULT_FILTER_BACKENDS": ("django_filters.rest_framework.DjangoFilterBackend",),
-    "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
+    "DEFAULT_SCHEMA_CLASS": "common.openapi.TenantScopedAutoSchema",
```
hugobessa, line 145: Global DEFAULT_SCHEMA_CLASS swap. Safe blast radius: the new class only adds the header when isinstance(view, TenantScopedViewMixin); every non-tenant DRF view (public_api, accounts, users, payments, AcceptInvitationView) returns early from the isinstance guard and is byte-for-byte unchanged in the schema. Referenced lazily as a string so there's no import cycle at startup.

PR #85, opened 2026-06-17
feat(organizations,public_api): Phase 0 — reseller hierarchy, DB-only capability flag, gate helper

hugobessa:
## Summary
Foundation phase for White-Label API Provisioning. Pure additive scaffolding — no reachable behavior change. Existing orgs are byte-for-byte unchanged (new boolean defaults off, new FK defaults NULL).

- `Organization.parent` self-FK (`on_delete=PROTECT`) + `can_invite_organizations` boolean (`default=False`), plus `is_reseller()` and `get_branding_root()` (walks the parent chain to the nearest reseller ancestor, else `None`).
- `can_invite_organizations` is **DB / Django-admin only** — never exposed by any GraphQL input/output type or DRF serializer. Enforced by a guard test that introspects the built GraphQL schema and scans Organization serializers.
- `public_api/capabilities.py::assert_org_can_invite(acting_org)` — the single gate every reseller-bundle resolver (Phases 1–9) calls after its `ResourceAccess` scope check (gate + scope = defense in depth).
- `PublicAPIResources` gains `MEMBERSHIP`, `INVITATION`, `BRANDING`, `CHILD_ORG_ANALYTICS` (scopes reserved for later phases; not yet wired to any resolver).
- Two additive, lock-light migrations (organizations columns + ResourceAccess choices). Reversible, no backfill.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 0. Implements Data Model Changes 3.1, 3.3, 3.4.

## Test plan
- `uv run pytest organizations/tests/test_models.py public_api/tests/test_capabilities.py public_api/tests/test_schema_surface.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1888 passed) + `uv run python manage.py makemigrations --check`.

Notes for reviewers: the schema-surface guard is **blocker-class** per the plan's Risk notes — it introspects `public_api/schema.py::schema._schema.type_map` (all input + output types) with an anti-vacuity assertion, so it fails loudly if a future PR exposes the flag or if introspection silently breaks.


organizations/models.py
```diff
@@ -86,10 +86,49 @@ class Organization(BaseModel):
...
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
+        related_name="child_organizations",
+        help_text=(
+            "The parent organization if this is a child org. "
+            "A reseller with live children cannot be deleted."
+        ),
+    )
+    can_invite_organizations = models.BooleanField(
```
hugobessa, line 100: DB-only capability switch (Goals #1, Guiding Decisions "The capability switch"). `default=False` so every existing org is unchanged. This field must never reach a GraphQL input/output type or DRF serializer — `public_api/tests/test_schema_surface.py` guards that invariant and the plan's Risk notes treat any future API exposure as a blocker.

organizations/models.py
```diff
@@ -86,10 +86,49 @@ class Organization(BaseModel):
     should_sync_rooms = models.BooleanField(
         default=False, help_text="Whether to sync rooms for this organization."
     )
+    parent = models.ForeignKey(
+        "self",
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
```
hugobessa, line 93: `on_delete=PROTECT` is deliberate (Data Model 3.1): a reseller with live child orgs must not be deletable out from under them — deletion fails loudly instead.

organizations/models.py
```diff
@@ -86,10 +86,49 @@ class Organization(BaseModel):
...
+        return self.can_invite_organizations
+
+    def get_branding_root(self) -> Organization | None:
+        """
+        Walk up the parent chain to the nearest ancestor with can_invite_organizations=True.
+
+        Returns the reseller ancestor (which has branding), or None if no such ancestor exists.
+        The None case means this org (or its entire lineage) has no reseller, so vinta defaults apply.
+        """
+        seen: set[int] = set()
```
hugobessa, line 123: Visited-PK guard. PROTECT does not prevent a parent cycle (A.parent=B; B.parent=A), so a naive `while org is not None` walk could hang when no org in the cycle is a reseller. `get_branding_root` is called on every branding resolution starting in Phase 6 — covered by a regression test that builds a 2-org cycle and asserts termination.

public_api/capabilities.py
```diff
@@ -0,0 +1,25 @@
...
+    scope. The DB flag is the operator's switch; the scope is the reseller's lease-privilege control
+    over its own tokens.
+
+    Args:
+        acting_org: The Organization instance performing the action.
+
+    Raises:
+        PermissionDenied: if acting_org.can_invite_organizations is False.
+    """
+    if not acting_org.can_invite_organizations:
```
hugobessa, line 22: The single gate chokepoint (Guiding Decisions "Gate is necessary, scope is also necessary"). Every reseller-bundle resolver calls this AFTER its `OrganizationResourceAccess` scope check — the DB flag is the operator's switch, the scope is the reseller's least-privilege control over its own tokens. Raises DRF `PermissionDenied` to match the public_api permission layer.

public_api/tests/test_schema_surface.py
```diff
@@ -0,0 +1,123 @@
...
+def collect_all_graphql_field_names() -> set[str]:
+    """Introspect the fully-built Strawberry schema's graphql-core type map.
+
+    Enumerates every input and output type and all their fields to ensure
+    comprehensive field discovery across all GraphQL types.
+    """
+    from public_api.schema import schema
+
+    names: set[str] = set()
+    for gql_type in schema._schema.type_map.values():
```
hugobessa, line 20: Blocker-class guard. Introspects the fully-built schema's `type_map` (every input AND output type), not just top-level resolver names — the flag would leak as a field on the Organization type or a Create*Input, not as a top-level query. The `assert field_names` anti-vacuity check (and the sibling one on the serializer scan) makes the guard fail loudly if introspection ever silently returns nothing.

public_api/constants.py
```diff
@@ -30,3 +30,7 @@ class PublicAPIResources(TextChoices):
     ORGANIZATION = "organization", "Organization"
     CALENDAR_GROUP = "calendar_group", "Calendar Group"
     SYSTEM_USER = "system_user", "System User"
+    MEMBERSHIP = "membership", "Membership"
```
hugobessa, line 33: Scopes reserved for Phases 1–9; not yet mapped to any resolver. The matching `ResourceAccess.resource_name` choices migration and `schema.yml` regen ship in this PR so the enum and the persisted choices stay in lockstep.

PR #86, opened 2026-06-17
feat(public_api): gated createOrganization mutation for reseller child provisioning

hugobessa:
## Summary
Adds the first reseller-bundle mutation: a reseller org creates a child organization through the public GraphQL API.

- `createOrganization(input: { name })` → `{ organization { id name } }`.
- Child is created with `parent = acting_org` and `can_invite_organizations = False` (hardcoded — never read from input; children are always plain orgs, the flag is never delegated).
- **Dual gate**: `OrganizationResourceAccess('ORGANIZATION')` permission class (scope, pre-resolver) AND `assert_org_can_invite(acting_org)` (DB flag, in-resolver). A flag-off org is refused even with the ORGANIZATION scope; a token without the scope is refused.
- Acting org resolved from `request.public_api_organization` (token-pinned / validated `X-Public-Api-Organization-Id`) — cannot be spoofed by input.
- Duplicate sibling name rejected by a `UniqueConstraint(parent, name)` (race-proof DB backstop) plus a friendly resolver-level message; Postgres NULL-distinct keeps top-level (parent=NULL) names unconstrained.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 1, API Design 4.1. **Stacked on Phase 0** (PR #85).

## Test plan
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1896 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes: security-invariant tests assert the SPECIFIC failure (gate message vs scope message) and that no org/membership was created, so a silent gate/scope regression can't pass via the wrong error. The flag-injection test attempts to pass `canInviteOrganizations: true` and fails if the field is ever added to the input type.


public_api/mutations.py
```diff
@@ -101,3 +105,49 @@ def delete_system_user(
...
+        4. Returns the created organization's id and name.
+
+        The token's OrganizationResourceAccess must include the ORGANIZATION resource.
+        """
+        acting_org = info.context.request.public_api_organization
+        if not acting_org:
+            raise GraphQLError("Organization not found")
+
+        # Gate: check the org can invite before proceeding
+        assert_org_can_invite(acting_org)
```
hugobessa, line 131: Gate runs AFTER the scope check (Guiding Decisions "Gate is necessary, scope is also necessary"): `OrganizationResourceAccess('ORGANIZATION')` is a permission class that fires pre-resolver; `assert_org_can_invite` fires here, in-resolver. So a flag-off org WITH the ORGANIZATION scope is still refused. DRF `PermissionDenied` surfaces cleanly through Strawberry as a GraphQL error (HTTP 200, `errors` array) — no 500.

public_api/mutations.py
```diff
@@ -101,3 +105,49 @@ def delete_system_user(
...
+            raise GraphQLError(
+                f"An organization with name '{input.name}' already exists under this parent."
+            )
+
+        # Create the child org with parent=acting_org and can_invite_organizations=False
+        try:
+            child_org = Organization.objects.create(
+                name=input.name,
+                parent=acting_org,
+                can_invite_organizations=False,
```
hugobessa, line 144: `can_invite_organizations=False` is hardcoded and never read from input (Non-goals: no flag delegation). The child is always a plain org; only a Vinta admin can flip the flag via the DB/admin. `CreateOrganizationInput` exposes only `name`, and the Phase 0 schema-surface guard covers this new input type automatically.

public_api/mutations.py
```diff
@@ -101,3 +105,49 @@ def delete_system_user(
...
+            )
+        except IntegrityError as e:
```
hugobessa, line 146: IntegrityError backstop for the duplicate-name race: the `.exists()` pre-check gives a friendly message, but two concurrent requests could both pass it, so the `UniqueConstraint(parent, name)` is the race-proof guarantee and this except re-raises it as the same user-facing GraphQLError.

organizations/migrations/0009_organization_uniq_org_name_per_parent.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ("organizations", "0008_organization_can_invite_organizations_and_more"),
+    ]
+
+    operations = [
+        migrations.AddConstraint(
+            model_name="organization",
+            constraint=models.UniqueConstraint(
+                fields=("parent", "name"), name="uniq_org_name_per_parent"
```
hugobessa, line 16: Unique on (parent, name). Postgres treats NULLs as distinct, so multiple top-level orgs (parent=NULL) may share a name — the constraint only bites siblings under the same non-null parent, which is exactly the plan's "duplicate name under parent" rule. Additive, reversible, single AddConstraint.

public_api/permissions.py
```diff
@@ -40,6 +40,7 @@ class OrganizationResourceAccess(BasePermission):
         "calendarGroupBookableSlots": PublicAPIResources.CALENDAR_GROUP,
         "calendarGroupEvents": PublicAPIResources.CALENDAR_GROUP,
         "deleteSystemUser": PublicAPIResources.SYSTEM_USER,
+        "createOrganization": PublicAPIResources.ORGANIZATION,
```
hugobessa, line 43: Maps the `createOrganization` field to the ORGANIZATION resource so OrganizationResourceAccess denies tokens lacking that scope. Per the create-graphql-public-query convention.

PR #87, opened 2026-06-17
feat(public_api): gated passwordless createUser mutation

hugobessa:
## Summary
Second reseller-bundle mutation: provision a passwordless end-user by email.

- `createUser(input: { email, firstName, lastName })` → `{ user { id email firstName lastName } }`.
- New user: `set_unusable_password()` + an allauth `EmailAddress(verified=False)` — passwordless and email-unverified. The user only becomes usable after the themed Google login (Google verifies the email) + the existing social-adapter invite auto-join. Vinta's verification posture is preserved, not bypassed.
- **Idempotent on email**: a repeat call returns the existing user without re-mutating it (no second EmailAddress, no password reset). A `Profile` is guaranteed on both paths so the idempotent return never crashes on a profile-less pre-existing user.
- **Dual gate**: `OrganizationResourceAccess('USER')` + in-resolver `assert_org_can_invite(acting_org)`. No `OrganizationMembership` is created here (membership arrives via Phase 3's invitation).

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 2, API Design 4.2. **Stacked on Phase 1** (PR #86).

## Test plan
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1904 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes: review caught a BLOCKER — the original idempotent return dereferenced `user.profile` while Profile has no auto-create signal, so an email mapping to a profile-less user would 500. Fixed by guaranteeing the Profile before return; a regression test exercises that exact path. A behavioral test now asserts `authenticate(...)` rejects the provisioned user (not just `has_usable_password() is False`).

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -1,10 +1,12 @@
 from dataclasses import dataclass
 from typing import Annotated, cast
 
+from django.contrib.auth import get_user_model
 from django.db import IntegrityError
 from django.utils import timezone
 
 import strawberry
+from allauth.account.models import EmailAddress
 from dependency_injector.wiring import Provide, inject
 from graphql import GraphQLError
 
@@ -14,7 +16,15 @@ from public_api.capabilities import assert_org_can_invite
 from public_api.models import SystemUser
 from public_api.permissions import IsAuthenticated, OrganizationResourceAccess
 from public_api.services import PublicAPIAuthService
-from public_api.types import CreateOrganizationInput, CreateOrganizationResult, OrganizationResult
+from public_api.types import (
+    CreateOrganizationInput,
+    CreateOrganizationResult,
+    CreateUserInput,
+    CreateUserResult,
+    OrganizationResult,
+    UserResult,
+)
+from users.models import Profile
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


public_api/mutations.py
```diff
@@ -151,3 +161,67 @@ def create_organization(
...
+        assert_org_can_invite(acting_org)
+
+        # Idempotent on email: get-or-create pattern
+        user, created = user_model.objects.get_or_create(
+            email=input.email,
+            defaults={},
+        )
+
+        # Guarantee a Profile exists (handles both newly created and profile-less existing users)
+        profile, _ = Profile.objects.get_or_create(user=user)
```
hugobessa, line 198: Profile is guaranteed on BOTH the created and the existing-user path (moved out of `if created`). This is the BLOCKER fix: Profile has no post_save signal, so the idempotent return previously crashed (`Profile.RelatedObjectDoesNotExist`) for an email mapping to a profile-less user. For an existing user this backfills an empty Profile (safe no-op); names/password/EmailAddress are only touched on the created path so idempotency never re-mutates.

public_api/mutations.py
```diff
@@ -151,3 +161,67 @@ def create_organization(
...
+
+        acting_org = info.context.request.public_api_organization
+        if not acting_org:
+            raise GraphQLError("Organization not found")
+
+        # Gate: check the org can invite before proceeding
+        assert_org_can_invite(acting_org)
+
+        # Idempotent on email: get-or-create pattern
+        user, created = user_model.objects.get_or_create(
```
hugobessa, line 192: Idempotent on email via get_or_create (API Design 4.2). User.email is unique, so the get_or_create IntegrityError fallback makes concurrent same-email calls safe. A repeat call returns the existing user and adds no second EmailAddress.

public_api/mutations.py
```diff
@@ -151,3 +161,67 @@ def create_organization(
...
+            email=input.email,
+            defaults={},
+        )
+
+        # Guarantee a Profile exists (handles both newly created and profile-less existing users)
+        profile, _ = Profile.objects.get_or_create(user=user)
+
+        if created:
+            # Only set unusable password and email if the user is newly created
+            user.set_unusable_password()
```
hugobessa, line 202: Passwordless + unverified (Guiding Decisions "User account semantics"): `set_unusable_password()` plus an allauth `EmailAddress(verified=False)`. The user is only usable after the themed Google login verifies the email + the social-adapter invite auto-join — verification is preserved, not bypassed. A behavioral test asserts `authenticate()` returns None.

public_api/permissions.py
```diff
@@ -41,6 +41,7 @@ class OrganizationResourceAccess(BasePermission):
         "calendarGroupEvents": PublicAPIResources.CALENDAR_GROUP,
         "deleteSystemUser": PublicAPIResources.SYSTEM_USER,
         "createOrganization": PublicAPIResources.ORGANIZATION,
+        "createUser": PublicAPIResources.USER,
```
hugobessa, line 44: Maps `createUser` → USER resource. Combined with the in-resolver `assert_org_can_invite`, a flag-off org WITH the USER scope is still refused and a token without USER scope is denied.

hugobessa:
Closing: Phase 2 (`createUser` standalone GraphQL mutation) has been removed from the plan. The invitation flow (`createInvitation` → `OrganizationService.invite_user_to_organization`) already creates-or-finds the user, so a standalone passwordless mutation was redundant. The branch `plan/whitelabel-api-provisioning/phase-2` is abandoned; the rest of the stack (PR #88 onward) has been rebased onto phase-1 to drop the createUser commits. See the plan's Amendments section.

PR #88, opened 2026-06-17
feat(public_api): gated createInvitation mutation with role propagation and subtree isolation

hugobessa:
## Summary
Third reseller-bundle mutation: invite a provisioned user into a child org; the user auto-joins on first themed login.

- `createInvitation(input: { userEmail, organizationId, role = MEMBER, sendEmail = true })` → `{ invitation { id email expiresAt }, token, inviteUrl }`. Phase 3 implements only the `sendEmail=true` (email) path; `token`/`inviteUrl` are null. The `sendEmail=false` raw-token branch is Phase 4.
- **Tenant isolation** (`assert_target_in_subtree`): `organizationId` must be the acting org or a descendant — walks the target's `parent` chain up to the acting org (cycle-guarded). Reused by Phases 5 & 9. Rejects cross-reseller-tree targets.
- **Dual gate**: `OrganizationResourceAccess('INVITATION')` + in-resolver `assert_org_can_invite`.
- `OrganizationInvitation` gains a `role` field (default MEMBER); `invited_by` is now nullable + `SET_NULL` (the public-API caller is a SystemUser, not a Django User). Role propagates on accept through BOTH paths — `accept_invitation` (token/REST) and `provision_tenant_for_user` (social-login auto-join).
- Already-active-member of the target → `UserAlreadyHasMembershipError`.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 3, API Design 4.3. **Stacked on Phase 2** (PR #87).

**Deviation (acknowledged):** Plan 4.3 names resources `INVITATION + MEMBERSHIP`. `OrganizationResourceAccess.FIELD_TO_RESOURCE_MAPPING` is one-resource-per-field by design; re-architecting it to AND-require multiple resources is out of Phase 3 scope. The field is gated on `INVITATION` (plus the `can_invite_organizations` DB flag); MEMBERSHIP is conceptually implied (the invitation creates a membership on accept). Flagging so plan and code don't diverge silently — revisit if multi-resource gating is wanted later.

## Test plan
- `uv run pytest public_api/tests/test_mutations.py public_api/tests/test_provisioning_flow.py public_api/tests/test_capabilities.py organizations/tests/test_services.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1923 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes: review caught a BLOCKER — `accept_invitation` (token/REST path) created the membership without `role=`, silently downgrading ADMIN invites to MEMBER (only the social-login path had been updated). Fixed + tested on both paths. Subtree isolation hardened with direct unit tests (grandchild accept, cross-reseller-tree reject, parent-cycle terminates). Implemented across two socket-interrupted agent runs; the orchestrator finished the OrgRole enum (`strawberry.Enum` → `enum.Enum`) + lint.


public_api/capabilities.py
```diff
@@ -23,3 +24,40 @@ def assert_org_can_invite(acting_org: Organization) -> None:
         raise PermissionDenied(
             "This organization does not have permission to invite or create other organizations."
         )
+
+
+def assert_target_in_subtree(acting_org: Organization, target_org: Organization) -> None:
```
hugobessa, line 29: Tenant-isolation guard (Risk notes: highest-risk area). Walks the target org's `parent` chain upward looking for the acting org, with a visited-set cycle guard so pathological parent loops terminate. Rejects any org outside the acting reseller's subtree — including another reseller's children. Reused by Phases 5 (createSystemUserToken) and 9 (childOrganizations). Direct unit tests cover self/child/grandchild accept, cross-reseller-tree reject, and cycle termination.

organizations/services.py
```diff
@@ -301,7 +309,9 @@ def accept_invitation(self, token: str, user: User) -> OrganizationMembership:
                 try:
                     with transaction.atomic():
                         membership = OrganizationMembership.objects.create(
-                            user=user, organization=invitation.organization
+                            user=user,
+                            organization=invitation.organization,
+                            role=invitation.role,
```
hugobessa, line 314: BLOCKER fix: the token/REST `accept_invitation` path previously created the membership without `role=`, silently downgrading an ADMIN invitation to MEMBER (only the social-login `provision_tenant_for_user` path had been updated). Both accept paths now propagate `invitation.role`, tested with an ADMIN-accept regression.

public_api/mutations.py
```diff
@@ -225,3 +235,86 @@ def create_user(
                 last_name=profile.last_name or None,
             )
         )
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def create_invitation(
```
hugobessa, line 171: Phase 3 implements only the sendEmail=true (email) path — token/inviteUrl returned null. The sendEmail=false raw-token branch is Phase 4. Reuses `OrganizationService.invite_user_to_organization` (DI-injected) with `invited_by=None` (the caller is a SystemUser, not a Django User) and `role=input.role.to_model_role()`. Gate (`assert_org_can_invite`) + subtree guard run before any write.

organizations/models.py
```diff
@@ -291,8 +291,19 @@ class OrganizationInvitation(BaseModel):
     )
     invited_by = models.ForeignKey(
```
hugobessa, line 292: `invited_by` is now nullable + `on_delete=SET_NULL`: a public-API invitation has no human inviter (the caller is a SystemUser). Existing human-invite callers still pass `invited_by` explicitly; the only behavior change is that deleting an inviter user now nulls the FK instead of CASCADE- deleting the invitation — benign/desirable.

public_api/permissions.py
```diff
@@ -42,6 +42,10 @@ class OrganizationResourceAccess(BasePermission):
         "deleteSystemUser": PublicAPIResources.SYSTEM_USER,
         "createOrganization": PublicAPIResources.ORGANIZATION,
         "createUser": PublicAPIResources.USER,
+        # createInvitation requires INVITATION scope. MEMBERSHIP is conceptually also implied
```
hugobessa, line 44: Acknowledged deviation from plan 4.3 (INVITATION + MEMBERSHIP): the permission mechanism is one-resource-per-field, so the field is gated on INVITATION (plus the DB flag); MEMBERSHIP is implied. Re-architecting multi-resource gating is out of Phase 3 scope — see PR description.

PR #89, opened 2026-06-17
feat(public_api): self-managed invitations — suppress email and return raw token once

hugobessa:
## Summary
Completes `createInvitation` with the `sendEmail=false` branch so a reseller can render the invite link in its own UI.

- `sendEmail=false` (gated by `can_invite_organizations` + INVITATION scope + subtree guard): no email is sent; returns the raw invitation token + a constructed `inviteUrl` **once**. The reseller delivers the link itself.
- `sendEmail=true` (default): unchanged Phase 3 behavior — email sent, `token`/`inviteUrl` null.
- **Security**: only `token_hash` (SHA-256) is ever persisted. The raw token is surfaced via a transient `invitation._raw_token` Python attribute set after `save()` and never re-saved — no plaintext column exists. The returned token validates end-to-end through `accept_invitation` (proven by test).
- `inviteUrl` is built from the same `HEADLESS_FRONTEND_URLS["account_accept_invitation"]` template the branded email uses (present in staging/production; test-only addition in settings/test.py). Phase 6+ may refine it from the reseller's `return_url_allowlist` once `OrganizationBranding` exists.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 4, API Design 4.3. **Stacked on Phase 3** (PR #88).

## Test plan
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1929 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes (security-sensitive phase, reviewed adversarially): no BLOCKER — all critical invariants confirmed (no plaintext persistence; returned token validates via accept_invitation; email truly suppressed on false with both existing callers unchanged at default True; the flag/scope/subtree gate still applies on the false path; true-path returns null token even though `_raw_token` is attached). Accepted low-value test-convention SHOULD-FIX (tests use `OrganizationService()` directly vs DI) as-is. Pre-existing cross-org token-scan in `accept_invitation` noted for a future hardening pass (out of scope).

### DIFF
--- a/organizations/services.py
+++ b/organizations/services.py
@@ -206,10 +206,17 @@ class OrganizationService:
         organization: Organization,
         invited_by: User | None = None,
         role: str = OrganizationRole.MEMBER,
+        send_email: bool = True,
     ) -> OrganizationInvitation:
         """
         Invite a user to join the organization. if the invitation already exists, resets the token
         and the expiration date.
+
+        The raw token is attached to the returned invitation as ``invitation._raw_token`` (a
+        transient, non-persisted attribute) so callers that need it (e.g. the public-API
+        self-managed-invitation path) can surface it once without it ever being stored in
+        plaintext. Only ``token_hash`` is persisted.
+
         :param email: Email of the user to invite.
         :param organization: Organization the user is being invited to.
         :param invited_by: User who is sending the invitation. May be None when the invitation is
@@ -217,6 +224,10 @@ class OrganizationService:
             corresponding Django User.
         :param role: Role the invited user should receive on accepting the invitation. Defaults to
             MEMBER. Use OrganizationRole.ADMIN for admin invitations.
+        :param send_email: When True (default) the invitation email is dispatched via the
+            notification service. When False the email is suppressed — the caller is responsible
+            for delivering the invite link using the raw token attached to the returned instance
+            as ``_raw_token``.
         """
         token = generate_long_lived_token()
         token_hash = hash_long_lived_token(token)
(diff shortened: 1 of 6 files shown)


organizations/services.py
```diff
@@ -252,27 +263,32 @@ def invite_user_to_organization(
...
-                        "invitation_url": (getattr(settings, "HEADLESS_FRONTEND_URLS", {}))
-                        .get("account_accept_invitation", "")
-                        .format(token=token),
-                    }
-                ),
-                subject_template="organizations/emails/organization_invitation.subject.txt",
-                preheader_template="organizations/emails/organization_invitation.pre_header.txt",
+        # Attach the raw token as a transient attribute so the caller can surface it once.
+        # It is never persisted — only token_hash is stored in the DB row.
+        invitation._raw_token = token  # type: ignore[attr-defined]
```
hugobessa, line 268: Raw token surfaced WITHOUT persisting plaintext: `_raw_token` is a transient Python attribute set AFTER save() and never re-saved; only `token_hash` (SHA-256) is stored on the row. Mirrors the SystemUser once-returned-token pattern. The caller reads it exactly once on the sendEmail=false path; on the email path it's attached but never returned.

organizations/services.py
```diff
@@ -252,27 +263,32 @@ def invite_user_to_organization(
...
+
+        if send_email:
```
hugobessa, line 270: Email suppression: the `if send_email:` guard wraps the entire transaction.on_commit notification dispatch. `send_email` defaults True, so both existing human-invite callers (serializers.py, views.py — which omit the arg) email byte-for-byte as before. A test asserts on_commit is NOT called on the false path.

public_api/mutations.py
```diff
@@ -296,25 +297,40 @@ def create_invitation(
...
+            send_email=input.send_email,
         )
 
-        # Phase 4: sendEmail=false returns raw token. For now (Phase 3), sendEmail is always true
-        # and the invitation email is sent by invite_user_to_organization above.
-        # token and invite_url remain null in the sendEmail=true path.
+        raw_token: str | None = None
+        invite_url: str | None = None
+
+        if not input.send_email:
```
hugobessa, line 248: sendEmail=false branch. The gate (`assert_org_can_invite`), INVITATION scope, and subtree guard all run earlier and independently of send_email, so a flag-off org cannot mint a raw token. inviteUrl is built from the same HEADLESS_FRONTEND_URLS template the email uses; if that key is unset the token is still returned (reseller builds its own URL). Token returned once, never retained.

PR #90, opened 2026-06-17
feat(public_api): gated createSystemUserToken delegation mutation

hugobessa:
## Summary
Fifth reseller-bundle mutation: a reseller mints delegated Public API tokens pinned to its own subtree.

- `createSystemUserToken(input: { organizationId, integrationName, resources: [String!]! })` → `{ systemUserId, token }` (plaintext token once).
- `organizationId` must be the acting org or a descendant (`assert_target_in_subtree`). The minted `SystemUser` is pinned to that target org.
- `resources` may include `ORGANIZATION` to delegate child-org creation — but a minted token can **never** set `can_invite_organizations` (the flag is DB/admin-only; the resolver never touches it). Children created via a minted token are always plain orgs.
- **Dual gate**: `OrganizationResourceAccess('SYSTEM_USER')` + in-resolver `assert_org_can_invite`. Resource list validated (non-empty, all valid `PublicAPIResources`) and deduped.
- Mint + ResourceAccess persistence wrapped in `transaction.atomic()` (mirrors the REST `SystemUserTokenCreate` serializer) so a duplicate `integration_name` rolls back cleanly with no orphan SystemUser.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 5, API Design 4.4. **Stacked on Phase 4** (PR #89).

## Test plan
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1944 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes: BLOCKER — mint + bulk_create were not atomic (the REST path is), so a duplicate integration_name poisoned the request transaction and could orphan a SystemUser; FIXED with `transaction.atomic()` + a duplicate-integration_name regression asserting no orphan rows. SHOULD-FIX — added a dedup-resources test and rewrote the toothless cross-reseller-tree integration test to actually exercise the subtree guard via createSystemUserToken. No-flag-delegation invariant confirmed (resolver never writes the flag; child via minted token is flag-false; minting doesn't flip the target's flag).

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -15,7 +15,8 @@ from organizations.exceptions import UserAlreadyHasMembershipError
 from organizations.models import Organization, OrganizationMembership
 from organizations.services import OrganizationService
 from public_api.capabilities import assert_org_can_invite, assert_target_in_subtree
-from public_api.models import SystemUser
+from public_api.constants import PublicAPIResources
+from public_api.models import ResourceAccess, SystemUser
 from public_api.permissions import IsAuthenticated, OrganizationResourceAccess
 from public_api.services import PublicAPIAuthService
 from public_api.types import (
@@ -23,6 +24,8 @@ from public_api.types import (
     CreateInvitationResult,
     CreateOrganizationInput,
     CreateOrganizationResult,
+    CreateSystemUserTokenInput,
+    CreateSystemUserTokenResult,
     InvitationResult,
     OrganizationResult,
 )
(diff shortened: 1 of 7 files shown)


public_api/mutations.py
```diff
@@ -334,3 +337,79 @@ def create_invitation(
...
+        invalid = [r for r in input.resources if r not in valid_values]
+        if invalid:
+            raise GraphQLError(
+                f"Invalid resource(s): {', '.join(invalid)}. "
+                f"Valid values are: {', '.join(sorted(valid_values))}."
+            )
+
+        # Mint the system user and persist ResourceAccess rows (mirrors REST create)
+        try:
+            with transaction.atomic():
```
hugobessa, line 327: BLOCKER fix: mint + ResourceAccess bulk_create wrapped in transaction.atomic() (an inner savepoint), mirroring the REST SystemUserTokenCreate serializer. Under ATOMIC_REQUESTS the realistic IntegrityError is the unique SystemUser.integration_name INSERT; without the savepoint it poisoned the request transaction and could leave an orphan SystemUser. A regression test asserts a duplicate integration_name yields no orphan rows.

public_api/mutations.py
```diff
@@ -334,3 +337,79 @@ def create_invitation(
             token=raw_token,
             invite_url=invite_url,
         )
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def create_system_user_token(
```
hugobessa, line 274: Token delegation, NOT flag delegation (Guiding Decisions). resources may include ORGANIZATION so the minted token can create child orgs on the reseller's behalf, but the resolver never touches `can_invite_organizations` (DB/admin-only) — children minted tokens create are always flag-false. Even a token pinned to the reseller org itself can only create plain children / mint further subtree tokens; it can never confer the DB flag.

public_api/mutations.py
```diff
@@ -334,3 +337,79 @@ def create_invitation(
...
+        assert_org_can_invite(acting_org)
+
+        # Resolve the target organization
+        try:
+            target_org = Organization.objects.get(id=int(input.organization_id))
+        except (Organization.DoesNotExist, ValueError) as e:
+            raise GraphQLError(f"Organization with id '{input.organization_id}' not found.") from e
+
+        # Tenant-isolation guard: target must be the acting org or a descendant
+        assert_target_in_subtree(acting_org, target_org)
```
hugobessa, line 311: Reuses the Phase 3 `assert_target_in_subtree` guard: the minted SystemUser is pinned to a target org that must be the acting org or a descendant, so a reseller cannot mint a token into another reseller's tree. Applied before any mint.

public_api/permissions.py
```diff
@@ -46,6 +46,7 @@ class OrganizationResourceAccess(BasePermission):
         # (the invitation will create a membership on accept), but the permission mechanism
         # supports one resource per field; INVITATION is the primary gating resource.
         "createInvitation": PublicAPIResources.INVITATION,
+        "createSystemUserToken": PublicAPIResources.SYSTEM_USER,
```
hugobessa, line 48: Maps createSystemUserToken → SYSTEM_USER. Combined with the in-resolver assert_org_can_invite, a flag-off org with the SYSTEM_USER scope is still refused.

PR #91, opened 2026-06-17
feat(organizations,public_api): OrganizationBranding model + gated updateBranding mutation

hugobessa:
## Summary
Branding storage + the reseller's GraphQL write surface.

- New `OrganizationBranding` model (OneToOne on the reseller org, `related_name="branding"`): `app_name`, `logo_url`, `primary_color`, `secondary_color`, `support_email`, `return_url_allowlist` (ArrayField). Migration 0011 (plain additive CreateModel).
- `resolve_branding(org) -> OrganizationBranding | None` = `getattr(org.get_branding_root(), "branding", None)` — walks to the nearest reseller ancestor (reuses the Phase 0 cycle-guarded walk); None ⇒ vinta default. Consumed by Phase 7 (public read), Phase 8 (emails), Phase 10a (REST).
- `updateBranding` mutation (resource BRANDING, gated by `assert_org_can_invite`): upserts on the **acting** org only (`update_or_create(organization=acting_org, ...)` — input has no org id, so a reseller can't brand another tree). Validates hex colors (`#RRGGBB`/`#RRGGBBAA`) and `return_url_allowlist` entries via Django `URLValidator` (http/https), and `app_name` length.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 6, Data Model 3.2, API Design 4.5. **Stacked on Phase 5** (PR #90). (This PR also carries the plan amendment commit that inserted Phase 10a.)

## Test plan
- `uv run pytest organizations/tests/test_branding.py public_api/tests/test_mutations.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1958 passed) + `uv run python manage.py makemigrations --check`.

Reviewer notes: the implementing agent's outer-gate claim was FALSE (it ran only its 13 new tests). The orchestrator's full-suite run caught a catastrophic accidental deletion — the agent dropped the trailing `return super().save(*args, **kwargs)` from `OrganizationModel.save()` while appending the new model, silently breaking persistence for every OrganizationModel subclass (584 failures). Restored + amended; full suite now 1958 passed. Layer 3 then audited the append boundaries (clean) and the SHOULD-FIX pass replaced a weak `startswith` URL check with `URLValidator` (the allowlist feeds Phase 7's OAuth return-URL surface), added `app_name` length/blank guards, hoisted the hex regex, and added a reseller-A-vs-B isolation regression.


organizations/models.py
```diff
@@ -414,3 +415,84 @@ def save(self, *args, **kwargs):
                     setattr(self, f"{field_name}_fk", foreign_object_field_value)
 
         return super().save(*args, **kwargs)
```
hugobessa, line 417: NB for reviewers: the trailing `return super().save(*args, **kwargs)` of `OrganizationModel.save()` sits immediately above the new OrganizationBranding class. The first implementer pass accidentally deleted this line while appending the model, which silently stopped EVERY OrganizationModel subclass from persisting (584-test failure caught by the orchestrator's full-suite run, not the agent's scoped run). Restored here; a full `pytest -n auto` (1958 passed) is the guard.

organizations/models.py
```diff
@@ -414,3 +415,84 @@ def save(self, *args, **kwargs):
...
+
+    class Meta:
+        verbose_name = "Organization Branding"
+        verbose_name_plural = "Organization Brandings"
+
+    def __str__(self):
+        return f"Branding for {self.organization.name}"
+
+
+def resolve_branding(org: Organization) -> OrganizationBranding | None:
```
hugobessa, line 479: resolve_branding reuses the Phase 0 cycle-guarded get_branding_root: returns the nearest reseller ancestor's branding row for any child, None when no reseller ancestor exists or the reseller has no row (⇒ vinta default). Single resolution entry point for Phase 7 (public read), Phase 8 (emails), Phase 10a (REST).

public_api/mutations.py
```diff
@@ -413,3 +424,83 @@ def create_system_user_token(
...
+        for url in allowlist:
+            try:
+                _url_validator(url)
+            except DjangoValidationError as e:
+                raise GraphQLError(
+                    f"Invalid URL in return_url_allowlist: '{url}'. Must be a valid http(s) URL."
+                ) from e
+
+        # Upsert branding on the acting org (always acts on acting org, never another org)
+        branding, _ = OrganizationBranding.objects.update_or_create(
```
hugobessa, line 417: Upsert keyed to acting_org only — UpdateBrandingInput has no org id, so a reseller can never brand another org's tree (regression test: reseller B's update leaves reseller A's row untouched). Gate (assert_org_can_invite) + BRANDING scope run first.

public_api/mutations.py
```diff
@@ -413,3 +424,83 @@ def create_system_user_token(
             system_user_id=strawberry.ID(str(system_user.id)),
             token=plaintext_token,
         )
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def update_branding(
```
hugobessa, line 361: return_url_allowlist entries are validated with Django URLValidator (not a startswith check): update_or_create bypasses the model's URLField validators, so this resolver is the only guard, and these URLs become Phase 7's OAuth return-address allowlist — a security surface. Hex colors and app_name length are validated here too.

public_api/permissions.py
```diff
@@ -47,6 +47,7 @@ class OrganizationResourceAccess(BasePermission):
         # supports one resource per field; INVITATION is the primary gating resource.
         "createInvitation": PublicAPIResources.INVITATION,
         "createSystemUserToken": PublicAPIResources.SYSTEM_USER,
+        "updateBranding": PublicAPIResources.BRANDING,
```
hugobessa, line 49: Maps updateBranding → BRANDING; combined with the in-resolver assert_org_can_invite, a non-reseller (flag-off) org with the BRANDING scope is still refused.

PR #92, opened 2026-06-17
feat(public_api): unauthenticated brandingForTenant public read query

hugobessa:
## Summary
The frontend can fetch resolved, secret-free branding for a tenant before any session exists (themes the OAuth interstitials, Phase 10b).

- `brandingForTenant(tenantId)` → `PublicBrandingResult { appName, logoUrl, primaryColor, secondaryColor }` — **unauthenticated** (no permission classes), parent-walked via `resolve_branding`.
- **No enumeration oracle**: a nonexistent tenant id, a non-integer id, and an existing-but-unbranded tenant all return the byte-identical vinta-default sentinel (single-sourced via `_vinta_default_branding()`). No distinct error/status/message.
- **No secrets**: `PublicBrandingResult` exposes only app name/logo/colors — never `support_email`, `return_url_allowlist`, the branding row id, or org names.
- **Rate-limited per-IP**: extended `OrganizationRateLimiter` so anonymous requests are keyed `anon:{client_ip}` (X-Forwarded-For aware); the authenticated path is unchanged (still keyed by org id, same bucket).

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 7, API Design 4.6. **Stacked on Phase 6** (PR #91).

## Test plan
- `uv run pytest public_api/tests/test_queries.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1965 passed, orchestrator-verified) + `uv run python manage.py makemigrations --check`.

Reviewer notes: no BLOCKER — both high-risk invariants confirmed (no enumeration oracle: all three return paths converge on the identical sentinel; authenticated rate-limiting byte-for-byte unchanged). **Residual coverage gap (accepted)**: the anon-IP rate-limit path can't be exhaustion-tested because test settings zero out the limits (limiter is None in tests, so `try_acquire` isn't reached); the keying/exhaustion is correct-by-inspection and the endpoint is low-impact (public branding with a vinta-default fallback). Follow-up: rate-limit test-infra to drive the in-memory limiter with a positive rate. XFF-spoof / shared-NAT bucket collapse accepted for v1 per the plan's per-IP throttle decision (documented inline; trusted-proxy-count IP derivation deferred pending real proxy topology).

### DIFF
--- a/public_api/extensions.py
+++ b/public_api/extensions.py
@@ -109,6 +109,15 @@ class OrganizationRateLimiter(SchemaExtension):
             rate_limit_key = str(organization_id)
         else:
             # Get client IP from headers (respects X-Forwarded-For behind proxy)
+            # NAT/XFF Tradeoff (v1):
+            # - Clients behind shared NAT or a CDN that doesn't forward per-client XFF
+            #   will share one anon rate-limit bucket (reduces granularity but acceptable
+            #   for public branding endpoint with vinta-default fallback).
+            # - XFF is client-controlled and spoofable, so anon-limit evasion/forgery
+            #   is accepted in v1 (impact is low: the endpoint returns only public branding
+            #   with a vinta-default fallback, so unauthorized access doesn't expose secrets).
+            # - Trusted-proxy-count-aware IP derivation is deferred (requires ops input
+            #   on real proxy topology).
             client_ip = request.headers.get("X-Forwarded-For", "").split(",")[
                 0
             ].strip() or request.META.get("REMOTE_ADDR", "")
--- a/public_api/extensions.py
+++ b/public_api/extensions.py
@@ -15,14 +15,16 @@ from common.redis import ResilientLimiter
 
 
 class OrganizationRateLimiter(SchemaExtension):
-    """
-    Uses redis and the leaky bucket algorithm to limit the number of requests a organization can make
-    within a specified period.
-    This is useful for preventing abuse and ensuring fair usage of resources across organizations.
+    """Rate-limit organization and anonymous requests.
 
-    Redis is optional: a process-wide circuit breaker guards every Redis call and,
-    when Redis is unconfigured or down, the limiter falls back to an in-process
-    bucket so the public API keeps serving requests.
+    Uses redis and the leaky bucket algorithm to limit the number of requests
+    an organization or IP can make within a specified period.
+    This is useful for preventing abuse and ensuring fair usage of resources
+    across organizations.
+
+    Redis is optional: a process-wide circuit breaker guards every Redis call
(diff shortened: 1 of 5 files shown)


public_api/queries.py
```diff
@@ -518,3 +534,42 @@ def calendar_group_events(
             group_id=group_id, start=start_datetime, end=end_datetime
         )
         return cast(list[CalendarEventGraphQLType], list(events))
+
+    @strawberry.field()
+    def branding_for_tenant(self, tenant_id: strawberry.ID) -> PublicBrandingResult:
```
hugobessa, line 539: Intentionally UNAUTHENTICATED (bare @strawberry.field(), unlike every other field which carries IsAuthenticated + OrganizationResourceAccess). No-enumeration-oracle: a missing id, a non-int id, and an existing-but-unbranded org all return the identical `_vinta_default_branding()` sentinel — no different error/status/message. Returns only public branding (no secrets, no allowlist, no id).

public_api/queries.py
```diff
@@ -79,6 +80,21 @@ def _get_org(info: strawberry.Info):
     return org
 
 
+def _vinta_default_branding() -> PublicBrandingResult:
```
hugobessa, line 83: The vinta-default sentinel is single-sourced here so the no-oracle guarantee (identical response for missing vs unbranded) can't drift between two literals in future edits.

public_api/extensions.py
```diff
@@ -80,30 +82,54 @@ def get_default_rates() -> Iterable[Rate]:
...
+            #   for public branding endpoint with vinta-default fallback).
+            # - XFF is client-controlled and spoofable, so anon-limit evasion/forgery
+            #   is accepted in v1 (impact is low: the endpoint returns only public branding
+            #   with a vinta-default fallback, so unauthorized access doesn't expose secrets).
+            # - Trusted-proxy-count-aware IP derivation is deferred (requires ops input
+            #   on real proxy topology).
+            client_ip = request.headers.get("X-Forwarded-For", "").split(",")[
+                0
+            ].strip() or request.META.get("REMOTE_ADDR", "")
+            rate_limit_key = f"anon:{client_ip}"
```
hugobessa, line 124: Anonymous requests are now rate-limited by `anon:{client_ip}`; the authenticated path is unchanged (still keyed by org id, same bucket, same `limiter is None` early-return). Accepted v1 tradeoffs are documented inline — shared NAT/CDN clients share one anon bucket, and X-Forwarded-For is client-controlled/spoofable (low impact: the endpoint returns only public branding with a vinta default). Trusted-proxy-count-aware IP derivation is deferred pending the real proxy topology.

PR #93, opened 2026-06-17
feat(accounts,organizations): reseller-branded invitation emails with byte-for-byte vinta fallback

hugobessa:
## Summary
The invitation email a child org's user receives reads as the reseller's product; non-reseller emails are unchanged.

- Invitation templates (body/subject/preheader) replace the hardcoded "Vinta Schedule" with `{{ app_name }}`, sourced from `resolve_branding(invitation.organization)` and defaulting to exactly "Vinta Schedule" when no reseller ancestor/branding exists — so the non-reseller email renders byte-for-byte as before.
- The invitation context handles public-API invites (`invited_by=None`, the reseller path) gracefully: it falls back to the organization name as the inviter rather than raising, so the branded `sendEmail=true` invite actually generates its email.
- `invited_by_name` preserves the original `f"{first_name} {last_name}"` (no `.strip()`) so an inviter with a blank last name renders identically to today.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 8. **Stacked on Phase 7** (PR #92). This is the only phase touching an existing user-facing surface; the backwards-compat guarantee is render-tested.

## Test plan
- `uv run pytest accounts/tests/test_email_branding.py organizations/tests/ -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1979 passed, orchestrator-verified) + `uv run python manage.py makemigrations --check`.

Reviewer notes: the implementer's first pass had three defects the orchestrator + reviewer caught and fixed — (1) confirmation templates wrongly substituted `{{ app_name }}` for the literal domain URL `https://vinta_schedule.com.br/` (byte-for-byte break) → reverted, those account emails stay vinta default (no reliable org context at confirmation time); (2) the invitation context raised for `invited_by=None`, killing the marquee public-API branded-invite path → made graceful; (3) a `.strip()` on `invited_by_name` changed trailing-space behavior for blank last names → reverted, now render-tested with an empty-last-name case. Dead per-email branding queries added to accounts/notification_contexts.py were removed. Tests now RENDER templates (not just inspect the context dict).

**Deferred (documented TODO at the send site):** From/reply-to = `support_email` — vintasend's `DjangoEmailNotificationAdapter` has no per-notification `from_email` hook, so wiring it requires extending the adapter; out of Phase 8 scope. `support_email` is carried in the context for footer/use now. Follow-up needed.

### DIFF
--- a/organizations/notification_contexts.py
+++ b/organizations/notification_contexts.py
@@ -5,7 +5,15 @@ from django.core.exceptions import ObjectDoesNotExist
 from vintasend.exceptions import NotificationContextGenerationError
 from vintasend.services.notification_service import register_context
 
-from organizations.models import OrganizationInvitation
+from organizations.models import OrganizationInvitation, resolve_branding
+
+
+# Vinta Schedule default branding values (used when no reseller branding is configured)
+VINTA_DEFAULT_APP_NAME = "Vinta Schedule"
+VINTA_DEFAULT_LOGO_URL = ""
+VINTA_DEFAULT_PRIMARY_COLOR = ""
+VINTA_DEFAULT_SECONDARY_COLOR = ""
+VINTA_DEFAULT_SUPPORT_EMAIL = ""
 
 
 @register_context("organization_invitation_context")
@@ -14,6 +22,11 @@ def organization_invitation_context(
 ) -> dict[str, Any]:
     """
     Provides a context for organization invitation-related notifications.
+
+    Injects resolved branding (app_name, logo_url, primary_color, secondary_color,
+    support_email) for the invitation's organization. When the organization has a
+    reseller ancestor with a branding row, that branding is used. Otherwise, vinta
+    defaults are applied (byte-for-byte backwards-compat guarantee).
     """
     try:
         invitation = OrganizationInvitation.objects.get(id=organization_invitation_id)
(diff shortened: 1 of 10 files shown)


organizations/notification_contexts.py
```diff
@@ -14,17 +22,51 @@ def organization_invitation_context(
...
+        # Compute invited_by_name in the normal path without strip() for byte-for-byte
+        # compatibility with phase-7 (preserves trailing space when last_name is empty).
+        invited_by_name = f"{first_name} {last_name}"
+
+    # Resolve branding: walks to the nearest reseller ancestor and uses its branding row.
+    # If no reseller ancestor or no branding row, returns None → vinta defaults apply.
+    branding_row = resolve_branding(invitation.organization)
+
+    # Build the branding context dict with resolved values or vinta defaults.
+    branding_context = {
```
hugobessa, line 57: Branding injected via resolve_branding(invitation.organization). Every field defaults to the EXACT current vinta value (app_name "Vinta Schedule", others empty), so a non-reseller invitation email renders byte-for-byte as pre-phase-8 — the one user-facing surface this plan touches. Render-based tests assert the default output equals the baseline and the branded output shows the reseller app name with no vinta-domain leak.

organizations/notification_contexts.py
```diff
@@ -14,17 +22,51 @@ def organization_invitation_context(
...
         invitation = OrganizationInvitation.objects.get(id=organization_invitation_id)
     except OrganizationInvitation.DoesNotExist as e:
         raise NotificationContextGenerationError("Invalid organization invitation ID") from e
 
-    try:
-        first_name = invitation.invited_by.profile.first_name
-        last_name = invitation.invited_by.profile.last_name
-    except ObjectDoesNotExist as e:  # noqa: BLE001
-        raise NotificationContextGenerationError("Failed to retrieve inviter's profile") from e
+    if invitation.invited_by is None:
```
hugobessa, line 36: Public-API invites (createInvitation, Phases 3-5) are created with invited_by=None (the caller is a SystemUser, not a Django User). The context falls back to the organization name as the inviter so the branded sendEmail=true email — the marquee path — generates and renders cleanly, instead of raising NotificationContextGenerationError (which the first pass did, silently breaking it).

organizations/notification_contexts.py
```diff
@@ -14,17 +22,51 @@ def organization_invitation_context(
...
+        invited_by_name = invitation.organization.name
+    else:
+        try:
+            first_name = invitation.invited_by.profile.first_name
+            last_name = invitation.invited_by.profile.last_name
+        except ObjectDoesNotExist as e:  # noqa: BLE001
+            raise NotificationContextGenerationError("Failed to retrieve inviter's profile") from e
+        # Compute invited_by_name in the normal path without strip() for byte-for-byte
+        # compatibility with phase-7 (preserves trailing space when last_name is empty).
+        invited_by_name = f"{first_name} {last_name}"
```
hugobessa, line 50: invited_by_name keeps the original `f"{first_name} {last_name}"` with NO .strip(): last_name is blank=True, so an inviter with an empty last name must still render "John " (trailing space) to stay byte-for-byte. A .strip() in an earlier pass silently changed this; a render test with an empty last name now locks the exact substring.

templates/organizations/emails/organization_invitation.body.html
```diff
@@ -11,7 +11,7 @@ <h1>Confirmation Signup</h1>
   {% blocktrans %}Dear user,{% endblocktrans %}
 {% endif %}
 </p>
-{% blocktrans with invited_by_name=invitation.invited_by_name organization_name=invitation.organization_name %}You're receiving this email because you were invited by {{ invited_by_name }} to join the organization {{ organization_name }} in Vinta Schedule.
+{% blocktrans with invited_by_name=invitation.invited_by_name organization_name=invitation.organization_name app_name=branding.app_name %}You're receiving this email because you were invited by {{ invited_by_name }} to join the organization {{ organization_name }} in {{ app_name }}.
```
hugobessa, line 14: "in Vinta Schedule" → "in {{ app_name }}", default "Vinta Schedule" — byte-for-byte identical for non-reseller invites, reseller app name for branded subtrees. NB: the confirmation/signup account emails were deliberately left vinta-default (reverted) — no reliable org context at send time, and an app name is not a substitute for the domain URL they reference.

PR #94, opened 2026-06-17
feat(public_api): gated childOrganizations analytics query

hugobessa:
## Summary
The reseller's analytics surface: list its direct child orgs with point-in-time aggregate counts.

- `childOrganizations(offset, limit)` → `[{ id, name, createdAt, membershipCount, calendarCount, eventCount, calendarGroupCount }]`.
- Direct children only (`parent = acting_org`); only the acting reseller's own children — no cross-reseller leak.
- Counts via four independent correlated `Subquery` annotations (one per metric), avoiding the multi-relation `Count()` join fan-out that would multiply counts. `Subquery(...) or 0` yields 0 for empty metrics.
- **Dual gate**: `OrganizationResourceAccess('CHILD_ORG_ANALYTICS')` + in-resolver `assert_org_can_invite`. Paginated offset/limit, ordered by pk.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 9, API Design 4.7, "Child analytics shape". **Stacked on Phase 8** (PR #93). Last backend GraphQL phase; Phase 10a (REST branding) is next.

## Test plan
- `uv run pytest public_api/tests/test_queries.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (1988 passed, orchestrator-verified) + `uv run python manage.py makemigrations --check`.

Reviewer notes: no BLOCKER / no SHOULD-FIX. Fan-out avoidance and cross-reseller isolation both confirmed; tests seed distinct per-metric values (3/2/5/1) so a fan-out bug fails, and seed a second reseller's children to assert exclusion. `original_manager` use in the count subqueries is correct (the org-scoped default manager enforces tenancy via Python-side hooks that never fire inside a compiled Subquery; the correlation filter `organization_id=OuterRef("pk")` is provably a child of the acting org). Counts are of stored rows (no soft-delete on these models; recurring occurrences are not expanded) — the intended point-in-time semantics. NIT (not blocking): no direct `createdAt` assertion; `membershipCount` counts active+inactive per the plan's unqualified "memberships".

### DIFF
--- a/public_api/permissions.py
+++ b/public_api/permissions.py
@@ -47,6 +47,7 @@ class OrganizationResourceAccess(BasePermission):
         "createInvitation": PublicAPIResources.INVITATION,
         "createSystemUserToken": PublicAPIResources.SYSTEM_USER,
         "updateBranding": PublicAPIResources.BRANDING,
+        "childOrganizations": PublicAPIResources.CHILD_ORG_ANALYTICS,
     }
 
     def has_permission(self, source, info: Info, **kwargs) -> bool:  # type: ignore
--- a/public_api/queries.py
+++ b/public_api/queries.py
@@ -2,7 +2,8 @@ import datetime
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Annotated, cast
 
-from django.db.models import Value
+from django.db.models import Count as DjangoCount
+from django.db.models import OuterRef, Subquery, Value
 from django.db.models.functions import Concat
 
 import strawberry
@@ -33,12 +34,13 @@ from calendar_integration.models import (
     CalendarEvent,
     CalendarGroup,
 )
-from organizations.models import Organization, resolve_branding
+from organizations.models import Organization, OrganizationMembership, resolve_branding
+from public_api.capabilities import assert_org_can_invite
 from public_api.permissions import (
     IsAuthenticated,
     OrganizationResourceAccess,
 )
-from public_api.types import PublicApiHttpRequest, PublicBrandingResult
+from public_api.types import ChildOrganizationMetrics, PublicApiHttpRequest, PublicBrandingResult
 from users.graphql import UserGraphQLType
 from users.models import User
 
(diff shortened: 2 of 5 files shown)


public_api/queries.py
```diff
@@ -535,6 +537,84 @@ def calendar_group_events(
...
+
+        Gate: acting org must have can_invite_organizations=True (assert_org_can_invite)
+        AND the token must carry CHILD_ORG_ANALYTICS scope (OrganizationResourceAccess).
+        """
+        org = _get_org(info)
+        assert_org_can_invite(org)
+
+        # Subquery-based counts to avoid join fan-out when multiple aggregates
+        # are applied over different relations in a single queryset.
+        membership_sq = (
```
hugobessa, line 568: Four INDEPENDENT correlated Subquery counts (membership/calendar/event/group), each grouped on organization_id=OuterRef("pk"). This is the fan-out fix: combining multiple Count() over different relations in one annotate() cross-joins and multiplies the counts. Tests seed distinct per-metric values (3/2/5/1) so any fan-out regression fails. Uses `.original_manager` to bypass the org-scoped default manager (which would scope to the ACTING org and return 0 for children) — safe because the correlation pk is provably a child of the acting org.

public_api/queries.py
```diff
@@ -535,6 +537,84 @@ def calendar_group_events(
...
+        )
+        group_sq = (
+            CalendarGroup.original_manager.filter(organization_id=OuterRef("pk"))
+            .values("organization_id")
+            .annotate(cnt=DjangoCount("id"))
+            .values("cnt")
+        )
+
+        qs = (
+            Organization.objects.filter(parent=org)
```
hugobessa, line 594: Tenant isolation (Risk notes: highest-risk): outer queryset is `parent = acting_org` — DIRECT children only, so another reseller's children (and grandchildren) are excluded. A second-reseller seed test asserts no cross-reseller leak.

public_api/queries.py
```diff
@@ -535,6 +537,84 @@ def calendar_group_events(
...
+        says "memberships" with no active-only qualifier, so all rows are counted.
+
+        "Children" means DIRECT children (parent = acting_org). The plan says
+        "its child organizations"/"its children" — literal parent FK match.
+
+        Gate: acting org must have can_invite_organizations=True (assert_org_can_invite)
+        AND the token must carry CHILD_ORG_ANALYTICS scope (OrganizationResourceAccess).
+        """
+        org = _get_org(info)
+        assert_org_can_invite(org)
```
hugobessa, line 564: Gate (assert_org_can_invite) runs in-resolver after the OrganizationResourceAccess('CHILD_ORG_ANALYTICS') scope check — a flag-off org WITH the scope is still refused; a token without the scope is denied.

public_api/permissions.py
```diff
@@ -48,6 +48,7 @@ class OrganizationResourceAccess(BasePermission):
         "createInvitation": PublicAPIResources.INVITATION,
         "createSystemUserToken": PublicAPIResources.SYSTEM_USER,
         "updateBranding": PublicAPIResources.BRANDING,
+        "childOrganizations": PublicAPIResources.CHILD_ORG_ANALYTICS,
```
hugobessa, line 50: Maps childOrganizations → CHILD_ORG_ANALYTICS. Final scope wiring of the bundle (ORGANIZATION, USER, INVITATION, SYSTEM_USER, BRANDING, CHILD_ORG_ANALYTICS all now mapped).

PR #95, opened 2026-06-17
feat(organizations): reseller-admin REST branding endpoint (/branding/)

hugobessa:
## Summary
First-party REST surface (added per requester) so the frontend branding console (Phase 11b) can read/edit a reseller org's branding over REST — distinct from Phase 6's public GraphQL `updateBranding` (the reseller machine-integration surface). Both write the same `OrganizationBranding` row that Phase 8 emails and Phase 7 `brandingForTenant` resolve.

- `GET/PUT/PATCH /branding/` (singleton, at root — consistent with `/organizations/`, `/invitations/`). DELETE → 405.
- **Reseller-admin gated**: `IsOrganizationAdmin` (existing) AND the acting org must be a reseller (`can_invite_organizations=True`) — non-reseller admin → 403, non-admin member → 403, unauthenticated → 401. The reseller check runs on GET/PUT/PATCH.
- **Tenant-scoped**: subclasses `TenantScopedViewMixin`, so `X-Organization-Id` selects the acting org (multi-org admin → header drives selection; multi-org + no header → 400; non-member org → 403). Operates ONLY on the acting org's row — upsert keys on `membership.organization_id`, never input; cannot target another org.
- `OrganizationBrandingSerializer` round-trips `app_name, logo_url, primary_color, secondary_color, support_email, return_url_allowlist` (the first-party owner-edit surface reads back what it wrote — unlike the secret-minimized public GraphQL/`brandingForTenant`). Validates hex colors + allowlist URLs. NEVER exposes `can_invite_organizations`; `organization` is not writable. GET when unset → 404.

## Plan reference
ai-plans/2026-06-16-WHITELABEL_API_PROVISIONING_IMPLEMENTATION_PLAN.md — Phase 10a (inserted 2026-06-17). **Stacked on Phase 9** (PR #94). Depends on Phase 6's OrganizationBranding model.

**Deviation (accepted):** the plan suggested a DRF ViewSet registered in routes.py; the implementation uses an `APIView` singleton (a viewset's list/detail shape doesn't fit a single-row resource cleanly) mounted via `routes.extra_patterns` at `/branding/`. Functionally equivalent, tenant-scoped via the same mixin every viewset uses.

## Test plan
- `uv run pytest organizations/tests/test_branding_rest.py -vs`
- Outer gate: `uv run python manage.py check --deploy` + `uv run pytest -n auto` (2007 passed, orchestrator-verified) + `uv run python manage.py makemigrations --check`.

Reviewer notes: review caught a BLOCKER — the first cut was a bare APIView WITHOUT `TenantScopedViewMixin`, so it ignored `X-Organization-Id` and `get_active_organization_membership` fell back to the caller's OLDEST membership (wrong-org read/write for multi-org admins; the multi-org→400 guard never fired). Fixed by adding the mixin (APIView-safe — the mixin reads `self.action` via `getattr`); multi-org tests now prove header-driven selection + the absent-header 400. Earlier passes also fixed the `/api/branding/` → `/branding/` path inconsistency and DELETE 403 → 405.


organizations/views.py
```diff
@@ -851,3 +854,119 @@ def create(self, request, *args, **kwargs):
             },
             status=status.HTTP_201_CREATED,
         )
+
+
+@extend_schema(tags=["Branding"])
+class OrganizationBrandingView(TenantScopedViewMixin, views.APIView):
```
hugobessa, line 860: BLOCKER fix: subclasses TenantScopedViewMixin so X-Organization-Id resolves the acting org (the same mechanism every org viewset uses). The first cut was a bare APIView, so get_active_organization_membership fell back to the caller's OLDEST membership and ignored the header — a multi-org reseller-admin could read/write the wrong org's branding and the multi-org→400 guard never fired. The mixin is APIView-safe (it reads self.action via getattr); multi-org tests prove header-driven selection + the absent-header 400.

organizations/views.py
```diff
@@ -851,3 +854,119 @@ def create(self, request, *args, **kwargs):
...
+    uses ``getattr(self, "action", None)``, so the absent ``self.action`` attribute
+    (ViewSetMixin is not in the MRO) is safe — ``None`` is never in the default
+    ``active_org_optional_actions = ()`` tuple, so full 400/403 enforcement runs
+    on every HTTP method.
+    """
+
+    permission_classes = (IsOrganizationAdmin,)
+    serializer_class = OrganizationBrandingSerializer
+
+    def _check_reseller_status(self):
```
hugobessa, line 889: Reseller-admin gate, both layers: IsOrganizationAdmin (existing permission) + this reseller check (acting org must have can_invite_organizations=True). Runs on GET/PUT/PATCH, so a non-reseller admin is 403 on every verb. DELETE is 405 (method not allowed — no override).

organizations/views.py
```diff
@@ -851,3 +854,119 @@ def create(self, request, *args, **kwargs):
...
+        self._check_reseller_status()
+        membership = get_active_organization_membership(request.user)
+        if membership is None:
+            raise PermissionDenied("No active organization membership.")
+
+        serializer = OrganizationBrandingSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        # Create or update (upsert) the branding row for the acting org
+        instance, created = OrganizationBranding.objects.update_or_create(
```
hugobessa, line 942: Upsert keyed on membership.organization_id (the header-resolved acting org), never input — so the endpoint can only ever create/replace the acting org's OWN branding row; a multi-org test asserts writing org B leaves org A untouched.

organizations/serializers.py
```diff
@@ -326,3 +331,54 @@ def create(self, validated_data: dict):
...
+
+
+def _validate_hex_color(value: str) -> str:
+    """Validate a hex color string: #RRGGBB or #RRGGBBAA. Returns the value unchanged."""
+    if value and not re.match(r"^#[0-9A-Fa-f]{6}([0-9A-Fa-f]{2})?$", value):
+        raise serializers.ValidationError("Invalid color format. Expected #RRGGBB or #RRGGBBAA.")
+    return value
+
+
+class OrganizationBrandingSerializer(serializers.ModelSerializer):
```
hugobessa, line 343: REST serializer round-trips support_email + return_url_allowlist (the first-party owner editing its own row reads back what it wrote) — intentionally MORE than the secret-minimized public GraphQL BrandingResult / brandingForTenant. Never exposes can_invite_organizations; organization is not writable. Hex-color + allowlist-URL validation via the shared _validate_hex_color helper + URLValidator.

organizations/routes.py
```diff
@@ -30,3 +33,8 @@
         "basename": "OrganizationInvitations",
     },
 ]
+
+# Non-viewset routes (APIViews) — URL patterns to register with Django URL conf
+extra_patterns = [
+    path("branding/", OrganizationBrandingView.as_view(), name="branding"),
```
hugobessa, line 39: Mounted at /branding/ (root, via extra_patterns) to match the other org resources (/organizations/, /invitations/, /organization-members/) — corrected from an initial /api/branding/.

PR #97, opened 2026-06-18
feat(calendar): single-use booking-code lifecycle foundation (Phase 0)

hugobessa:
## Summary

Foundation phase for **single-use scheduling codes** (Public GraphQL API for the Building
Blocks / Medplum patient portal). No user-visible GraphQL field yet — this phase ships the
model, atomic consume path, standalone-code service methods, the new permission resource, and
the shared result types that Phases 1–6c consume.

What's in this phase:

- `CalendarManagementToken` gains nullable, additive columns: `expires_at`,
  `minted_by_system_user` (FK → `public_api.SystemUser`, `SET_NULL`), `consumed_source_ip`, and a
  `calendar_group` `OrganizationForeignKey` (so group booking codes can bind to a `CalendarGroup`).
- `CalendarManagementTokenManager`: `active()` (not used / not revoked / not expired),
  atomic `consume(token, source_ip)` (reentrant `transaction.atomic()` + `select_for_update()`
  + re-check → first-write-wins), and read-only `get_token_error_code()`.
- `CalendarPermissionService`: `create_booking_token(...) → (token, plaintext_code)`,
  `validate_code(code, organization_id) → token`, `consume_code(token, source_ip)`.
- New exceptions: `TokenExpiredError`, `TokenAlreadyUsedError`, `TokenRevokedError`.
- `PublicAPIResources.CALENDAR_BOOKING_CODE` + the six mint field names + `revokeBookingCode`
  mapped in `FIELD_TO_RESOURCE_MAPPING` (org-token gate for minting). With-code patient fields are
  deliberately NOT mapped — they are unauthenticated.
- Strawberry `BookingCodeErrorCode` enum + `BookingCodeResult` + `CodeEventResult` result types
  (defined here, wired by later phases).
- Standard Django migrations: `calendar_integration/0020`, `public_api/0007` (choices change).

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 0.
- Spec: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_SPEC.md` — Decisions → State transitions
  & edge cases (lifecycle: reusable read, single-use on first successful write, atomic, optional
  expiry, revocable).

## Test plan

All gates run in-container (docker postgres) via `docker compose run --rm api uv run …`:

- `ruff check ./` — clean.
- `python manage.py makemigrations --check --dry-run` — no changes.
- `python manage.py check --deploy` — 5 pre-existing security warnings, 0 errors.
- `pytest -n auto` — **2032 passed**.
- New: `calendar_integration/tests/test_management_token_manager.py` (incl. a real two-thread
  concurrency test proving first-write-wins under row-lock contention),
  `calendar_integration/tests/test_calendar_permission_service_codes.py`.

Stacked on `plan/single-use-scheduling-codes/wt-base`.


calendar_integration/managers.py
```diff
@@ -224,3 +239,83 @@ class CalendarEventGroupSelectionManager(BaseOrganizationModelManager):
...
+
+            if locked.used_at is not None:
+                raise TokenAlreadyUsedError()
+
+            if locked.expires_at is not None and locked.expires_at <= now:
+                raise TokenExpiredError()
+
+            locked.used_at = now
+            locked.consumed_source_ip = source_ip
+            locked.save(update_fields=["used_at", "consumed_source_ip"])
```
hugobessa, line 301: Core single-use invariant. `consume()` wraps lock + re-check + save in a reentrant `transaction.atomic()` (a no-op when a request transaction already exists) so the `select_for_update()` row lock is reliable even for non-request callers (Celery tasks, management commands) where `ATOMIC_REQUESTS` is False. The re-check of revoked/used/expired AFTER acquiring the lock is what makes two concurrent writes first-write-wins: the loser re-reads `used_at IS NOT NULL` and raises `TokenAlreadyUsedError`. Callers MUST commit the booking action in the same transaction so a rollback also rolls back `used_at`.

calendar_integration/managers.py
```diff
@@ -224,3 +239,83 @@ class CalendarEventGroupSelectionManager(BaseOrganizationModelManager):
...
+
+        Returns None when the token is active (no error).
+
+        This method does NOT acquire a lock — it is safe to call for read-only
+        validation where atomicity is not required (e.g. resolvers for
+        availability reads).
+
+        Returns:
+            ``"REVOKED"`` / ``"ALREADY_USED"`` / ``"EXPIRED"`` or ``None``.
+        """
```
hugobessa, line 314: Lifecycle precedence (revoked > used > expired) and the expiry boundary (`expires_at > now` = active, `<= now` = expired) are kept identical across `active()`, `get_token_error_code()`, and the service's `validate_code`/`consume()`. Keep these three in sync if the lifecycle ever changes — later phases rely on all paths agreeing.

calendar_integration/services/calendar_permission_service.py
```diff
@@ -487,3 +498,157 @@ def create_external_attendee_schedule_token(
...
+        if token.revoked_at is not None:
+            raise TokenRevokedError()
+
+        if token.used_at is not None:
+            raise TokenAlreadyUsedError()
+
+        if token.expires_at is not None and token.expires_at <= timezone.now():
+            raise TokenExpiredError()
+
+        return token
```
hugobessa, line 634: `validate_code` is read-only (no lock) — safe for the code-gated availability reads in Phase 4, which must NOT consume the code. The write phases (5a/5b/6a/6b/6c) call `validate_code` then `consume_code` inside one transaction; the validate→consume TOCTOU is safe because `consume()` re-checks terminal state under `select_for_update()`. Malformed/unknown/wrong-org/bad-hash all collapse to a single `InvalidTokenError` so the surface never reveals which codes exist.

calendar_integration/services/calendar_permission_service.py
```diff
@@ -487,3 +498,157 @@ def create_external_attendee_schedule_token(
...
+
+        token.save()
+
+        for perm in permissions:
+            token.permissions.create(permission=perm, organization_id=organization_id)
+
+        # Encode as "<id>:<raw>" in base64, matching initialize_with_token's decode logic.
+        plaintext_code = base64.b64encode(f"{token.pk}:{token_str}".encode()).decode("utf-8")
+
+        return token, plaintext_code
```
hugobessa, line 577: Plaintext code is `base64("<token_id>:<raw>")` and returned exactly once here; only the SHA-256 hash is persisted (`token_hash`). The id-prefix lets `validate_code` look the row up before the constant-time hash compare. Phase 1+ mint mutations return this in `BookingCodeResult.code` once and never again.

calendar_integration/models.py
```diff
@@ -1641,6 +1647,17 @@ class CalendarManagementToken(OrganizationModel):
...
+        "CalendarGroup",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="management_tokens",
+        help_text=(
+            "If set, this token is scoped to a calendar group (for group booking codes). "
+            "Mutually exclusive with the ``calendar`` scope for booking codes."
+        ),
+    )
```
hugobessa, line 1660: `calendar_group` was added (no such FK existed) so group booking codes can bind to a `CalendarGroup`. It mirrors the existing `calendar`/`event` `OrganizationForeignKey` composite-FK multi-tenancy pattern. Booking codes set exactly one of `calendar` / `calendar_group`; reschedule/cancel codes additionally set `event`. Consumed by Phases 1, 2, 5b, 6b.

public_api/permissions.py
```diff
@@ -48,6 +48,14 @@ class OrganizationResourceAccess(BasePermission):
...
         "updateBranding": PublicAPIResources.BRANDING,
         "childOrganizations": PublicAPIResources.CHILD_ORG_ANALYTICS,
+        # Single-use booking-code mint / revoke mutations (Phase 0+)
+        "createCalendarBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "createCalendarGroupBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "createCalendarRescheduleBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "createCalendarGroupRescheduleBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "createCalendarCancellationBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "createCalendarGroupCancellationBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
+        "revokeBookingCode": PublicAPIResources.CALENDAR_BOOKING_CODE,
```
hugobessa, line 58: Minting + revoke are gated by the new `CALENDAR_BOOKING_CODE` resource on the org token. Note the with-code patient mutations (createCalendarEventWithCode, …) are deliberately ABSENT from this map — they are unauthenticated and authorized solely by the code itself (Phase 4–6c follow the `check_token` precedent: no permission_classes).

calendar_integration/graphql.py
```diff
@@ -384,3 +385,65 @@ class BookableSlotProposalGraphQLType:
...
+    id: int | None = None  # noqa: A003
+
+
+@strawberry.type
+class CodeEventResult:
+    """Result type for unauthenticated with-code write mutations (book / reschedule / cancel).
+
+    ``event`` is present only when the operation succeeded.
+    """
+
```
hugobessa, line 445: `BookingCodeErrorCode` / `BookingCodeResult` / `CodeEventResult` are defined here but not wired to any field in this phase — intentional scaffolding. Result shape matches the repo convention (`success` + `error_message` + payload) plus a machine-readable `error_code` enum so integrators branch on failure category without string-parsing.

PR #99, opened 2026-06-18
feat(calendar): mint single-use booking codes (Phase 1)

hugobessa:
## Summary

Adds the two org-token-gated mint mutations on `CalendarGroupMutations` (reaching the public
schema via `public_api.mutations.Mutation`):

- `createCalendarBookingCode(input: { organizationId, calendarId, expiresAt })` → `BookingCodeResult`
- `createCalendarGroupBookingCode(input: { organizationId, calendarGroupId, expiresAt })` → `BookingCodeResult`

Both are gated by `IsAuthenticated + OrganizationResourceAccess` → the `CALENDAR_BOOKING_CODE`
resource (mapping shipped in Phase 0). They delegate to Phase 0's
`CalendarPermissionService.create_booking_token(permissions=[CREATE])`, set `minted_by` from the
request system user, and return the plaintext code + opaque id once. Bundle calendars
(`calendar_type=BUNDLE`) mint through the calendar variant transparently.

Security: the organization is taken from the authenticated request (`public_api_organization`),
never trusted from the input. A `organizationId` that mismatches the token's org, or a
calendar/group not in the org, both return `success=False, errorCode=INVALID_CODE` with an
identical message — no cross-org existence is revealed.

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 1.
- Spec: Decisions → Use-cases, Use-case 1 (provider mints a booking code).

## Test plan

In-container (docker postgres) via `docker compose run --rm api uv run …`:

- `public_api/tests/test_booking_code_mutations.py` — 11 tests: calendar + group mint (asserting
  token scope, `CREATE` permission row, `minted_by` set), without-resource rejection (no token
  created), bundle calendar, `expiresAt` persisted (full-timestamp equality), cross-org calendar/
  group → INVALID_CODE, and organizationId-mismatch → INVALID_CODE (no token).
- `ruff check` clean · `makemigrations --check` no changes · `check --deploy` clean ·
  `pytest -n auto` **2043 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-0`.

### DIFF
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -405,7 +405,7 @@ def get_booking_code_mutation_dependencies(
 ) -> BookingCodeMutationDependencies:
     """Get booking-code mutation dependencies from DI container."""
     if calendar_permission_service is None:
-        raise GraphQLError("Missing required dependency: calendar_permission_service")
+        raise GraphQLError("Internal server error.")
     return BookingCodeMutationDependencies(
         calendar_permission_service=cast("CalendarPermissionService", calendar_permission_service),
     )
@@ -562,6 +562,13 @@ class CalendarGroupMutations:
                 error_message="Organization not found.",
             )
 
+        if input.organization_id != org.id:
+            return BookingCodeResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
+                error_message="Organization not found.",
+            )
+
         # Verify the calendar belongs to the authenticated org.
         try:
             Calendar.objects.filter_by_organization(org.id).get(id=input.calendar_id)
@@ -597,6 +604,13 @@ class CalendarGroupMutations:
         """
         org = info.context.request.public_api_organization
         if org is None:
+            return BookingCodeResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
+                error_message="Organization not found.",
+            )
+
+        if input.organization_id != org.id:
             return BookingCodeResult(
                 success=False,
                 error_code=BookingCodeErrorCode.INVALID_CODE,
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -10,11 +10,13 @@ from graphql import GraphQLError
 
 from calendar_integration.exceptions import CalendarGroupError
 from calendar_integration.graphql import (
+    BookingCodeErrorCode,
+    BookingCodeResult,
     CalendarEventGraphQLType,
     CalendarGroupGraphQLType,
     CalendarWebhookSubscriptionGraphQLType,
 )
-from calendar_integration.models import CalendarGroup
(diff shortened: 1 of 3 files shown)


calendar_integration/mutations.py
```diff
@@ -493,3 +541,99 @@ def create_calendar_group_event(
...
+        token, plaintext_code = deps.calendar_permission_service.create_booking_token(
+            organization_id=org.id,
+            permissions=[EventManagementPermissions.CREATE],
+            expires_at=input.expires_at,
+            minted_by=minted_by,
+            calendar_id=input.calendar_id,
+        )
+        return BookingCodeResult(success=True, code=plaintext_code, id=token.pk)
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
```
hugobessa, line 593: Org-token-gated mint (permission_classes → CALENDAR_BOOKING_CODE, mapped in Phase 0). Booking codes carry ONLY `EventManagementPermissions.CREATE` and never set `event_id` (that's for the reschedule/cancel codes in Phase 2). The calendar variant accepts a bundle calendar (`calendar_type=BUNDLE`) with no special branch — it's just a calendar id.

calendar_integration/mutations.py
```diff
@@ -493,3 +541,99 @@ def create_calendar_group_event(
...
+                error_message="Organization not found.",
+            )
+
+        if input.organization_id != org.id:
+            return BookingCodeResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
+                error_message="Organization not found.",
+            )
+
```
hugobessa, line 571: Confused-deputy guard. The authoritative org is `request.public_api_organization` (token-bound); the input's `organizationId` is validated against it and a mismatch returns INVALID_CODE with the SAME message as not-found, so the surface never reveals whether another org's calendar/group exists. `organizationId` is kept in the input to match the plan's API Design, but it is never trusted as the scope source.

calendar_integration/mutations.py
```diff
@@ -381,6 +385,50 @@ def _load_organization(organization_id: int) -> Organization | None:
...
+        "CalendarPermissionService | None", Provide["calendar_permission_service"]
+    ] = None,
+) -> BookingCodeMutationDependencies:
+    """Get booking-code mutation dependencies from DI container."""
+    if calendar_permission_service is None:
+        raise GraphQLError("Internal server error.")
+    return BookingCodeMutationDependencies(
+        calendar_permission_service=cast("CalendarPermissionService", calendar_permission_service),
+    )
+
```
hugobessa, line 412: DI resolution mirrors the existing `get_calendar_group_mutation_dependencies` pattern in this module (`@inject` + `Provide["calendar_permission_service"]`). `create_booking_token` does not depend on `initialize_with_token` state, so the shared service instance is safe to reuse across requests for minting.

PR #98, opened 2026-06-18
Phase 2a — createResourceCalendar Public GraphQL mutation

hugobessa:
## Summary
Adds the `createResourceCalendar` Public GraphQL mutation so admin integrators can create a manual resource (room/equipment) calendar. Thin wrapper over `CalendarService.create_resource_calendar`.

- `public_api/constants.py`: new `PublicAPIResources.CREATE_RESOURCE_CALENDAR`.
- `public_api/permissions.py`: `FIELD_TO_RESOURCE_MAPPING["createResourceCalendar"]`.
- `public_api/mutations.py`: `CreateResourceCalendarInput` + `CreateResourceCalendarResult` + `create_resource_calendar` mutation, guarded by `IsAuthenticated` + `OrganizationResourceAccess`, delegating via the Phase 0 `_get_org_and_init_calendar_service` helper.
- `public_api/migrations/0007_alter_resourceaccess_resource_name.py`: choices-only `AlterField` for the new resource value (no schema change).

This branch also carries a docs-only commit (`f651b66`) recording that the single-calendar **event** mutation phases (1a/1b/1c) were dropped — `CalendarEventService` deliberately blocks `SystemUser` callers from creating/updating events via the Public API.

## Plan reference
Plan `ai-plans/2026-06-17-PUBLIC_GRAPHQL_SERVICE_WRAPPERS_IMPLEMENTATION_PLAN.md` → Phase 2a. No feature flag (per-token resource grant gates exposure). Stacked on Phase 0 (PR #96).

## Test plan
Run inside docker compose (host Postgres unused):
```
COMPOSE_PROJECT_NAME=vinta-schedule docker compose run --rm api uv run python manage.py check --deploy
COMPOSE_PROJECT_NAME=vinta-schedule docker compose run --rm api uv run pytest -n auto
```
Outer gate green: ruff ✅, format ✅, check --deploy ✅, makemigrations --check ✅, pytest **2010 passed**. `TestCreateResourceCalendarMutation` covers happy path (DB row + resource calendar_type), permission-denied, unauthenticated, org-scoping, and the service-error (`success=False`) branch.

PR #100, opened 2026-06-18
Phase 2b — disableResourceCalendar mutation + disable_resource_calendar service method

hugobessa:
## Summary
Adds the `disableResourceCalendar` Public GraphQL mutation and the backing `CalendarService.disable_resource_calendar(calendar_id)` service method (the doc flagged this as a true gap — no service method existed; disable = `visibility = INACTIVE`).

- `calendar_integration/services/calendar_service.py`: `disable_resource_calendar` — org-scoped fetch, validates `calendar_type == RESOURCE` (raises `ValueError` otherwise), sets `visibility = CalendarVisibility.INACTIVE`, `save(update_fields=["visibility"])`.
- `public_api/mutations.py`: `disableResourceCalendar` mutation (`{success, errorMessage}`), guarded by `IsAuthenticated` + `OrganizationResourceAccess`, via the Phase 0 helper.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.DISABLE_RESOURCE_CALENDAR` + mapping.
- `public_api/migrations/0008_alter_resourceaccess_resource_name.py`: choices-only AlterField.

## Plan reference
Plan → Phase 2b. No feature flag. Stacked on Phase 2a (PR #98).

## Test plan
Docker compose gate (host Postgres unused). Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2018 passed**. Tests: service-level unit tests in `calendar_integration/tests/services/test_calendar_service.py` (INACTIVE set; cross-org/missing → `DoesNotExist`; non-resource → `ValueError`) + HTTP integration tests in `public_api/tests/test_mutations.py` (happy path, non-resource rejected, cross-org rejected, permission-denied).


calendar_integration/services/calendar_service.py
```diff
@@ -706,6 +706,33 @@ def create_resource_calendar(
...
+        if calendar.calendar_type != CalendarType.RESOURCE:
+            raise ValueError(
+                f"Calendar {calendar_id} is not a resource calendar "
+                f"(type={calendar.calendar_type})."
+            )
+
+        calendar.visibility = CalendarVisibility.INACTIVE
+        calendar.save(update_fields=["visibility"])
+
+        return calendar
```
hugobessa, line 734: New service method (true gap — the doc noted there was no `disable_*` method, only the `visibility` field). Disable = soft state change to `CalendarVisibility.INACTIVE`, never a row delete, so the calendar drops out of `calendars().only_listed()` but history is kept. Org-scoping is via `filter_by_organization` so a cross-org id raises `Calendar.DoesNotExist` (the resolver maps that to success=False). The `calendar_type == RESOURCE` guard keeps this mutation from disabling non-resource calendars (e.g. user/virtual/bundle).

calendar_integration/services/calendar_service.py
```diff
@@ -706,6 +706,33 @@ def create_resource_calendar(
...
+        Fetches the calendar with an org-scoped query, validates it is a resource calendar,
+        then sets ``visibility = INACTIVE`` and saves.
+
+        :param calendar_id: Primary key of the Calendar to disable.
+        :return: The updated Calendar instance.
+        :raises Calendar.DoesNotExist: If no calendar with this id exists within the org.
+        :raises ValueError: If the calendar is not of type RESOURCE.
+        """
+        if not is_initialized_or_authenticated_calendar_service(self):
+            raise
```
hugobessa, line 721: `if not is_initialized_or_authenticated_calendar_service(self): raise` is copied verbatim from the sibling methods in this file (e.g. batch_modify_available_times) — kept for consistency rather than introduced here. The helper itself raises when the org isn't set, so the bare `raise` is effectively unreachable in the default path. Left as-is intentionally; a repo-wide cleanup of this pattern is out of scope for this phase.

PR #101, opened 2026-06-18
feat(calendar): mint single-use reschedule & cancel codes (Phase 2)

hugobessa:
## Summary

Adds the four event-bound, org-token-gated mint mutations on `CalendarGroupMutations`:

- `createCalendarRescheduleBookingCode` / `createCalendarGroupRescheduleBookingCode` → permission `RESCHEDULE`
- `createCalendarCancellationBookingCode` / `createCalendarGroupCancellationBookingCode` → permission `CANCEL`

Each binds to one specific event (single-use, `event_id` set) and follows the Phase 1 pattern:
org from the authenticated request, `organizationId`-mismatch rejected, `minted_by` from the
request system user, returns `BookingCodeResult{code, id}`.

Scope validation at mint:
- The named calendar/group must belong to the org.
- The event must belong to the org AND to the named calendar (`calendar_fk_id`) or group
  (`calendar_group_fk_id`).
- **Calendar variants additionally require `calendar_group_fk_id IS NULL`** so a calendar-shaped
  reschedule/cancel code can never bind to a grouped event (which the group consume path in
  Phase 6 handles differently). Group variants require the event's `calendar_group_fk_id` to match.

All failure branches return an identical generic `"Not found."` with `INVALID_CODE`, so a prober
can't distinguish org / scope / event failures.

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 2.
- Spec: Decisions → Use-cases, Use-case 3 (provider mints a reschedule/cancel code).

## Test plan

In-container (docker postgres): `public_api/tests/test_reschedule_cancel_code_mutations.py` covers,
per mutation — happy path (token event-scoped + correct SINGLE permission + minted_by), event-in-
different-calendar/group → INVALID_CODE (no token), cross-org → INVALID_CODE, without-resource
rejection, organizationId-mismatch, a permission-swap guard (RESCHEDULE vs CANCEL not swapped), and
grouped-event-rejected for the calendar variants. `ruff` clean · `makemigrations --check` no changes ·
`check --deploy` clean · `pytest -n auto` **2066 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-1`.

### DIFF
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -16,7 +16,12 @@ from calendar_integration.graphql import (
     CalendarGroupGraphQLType,
     CalendarWebhookSubscriptionGraphQLType,
 )
-from calendar_integration.models import Calendar, CalendarGroup, EventManagementPermissions
+from calendar_integration.models import (
+    Calendar,
+    CalendarEvent,
+    CalendarGroup,
+    EventManagementPermissions,
+)
 from calendar_integration.services.dataclasses import (
     CalendarGroupEventInputData,
     CalendarGroupInputData,
@@ -429,6 +434,26 @@ class CreateGroupBookingCodeInput:
     expires_at: datetime.datetime | None = None
 
 
+@strawberry.input
+class CreateEventCodeInput:
+    """Input for minting a single-use reschedule or cancel code scoped to a calendar + event."""
+
+    organization_id: int
+    calendar_id: int
+    event_id: int
+    expires_at: datetime.datetime | None = None
+
+
+@strawberry.input
+class CreateGroupEventCodeInput:
+    """Input for minting a single-use reschedule or cancel code scoped to a calendar group + event."""
+
+    organization_id: int
+    calendar_group_id: int
+    event_id: int
+    expires_at: datetime.datetime | None = None
+
+
 @strawberry.type
 class CalendarGroupMutations:
     """GraphQL mutations for CalendarGroup CRUD and grouped event booking."""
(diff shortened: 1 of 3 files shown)


calendar_integration/mutations.py
```diff
@@ -637,3 +662,249 @@ def create_calendar_group_booking_code(
...
+        try:
+            CalendarEvent.objects.filter_by_organization(org.id).get(
+                id=input.event_id,
+                calendar_fk_id=input.calendar_id,
+                calendar_group_fk_id__isnull=True,
+            )
+        except CalendarEvent.DoesNotExist:
+            return BookingCodeResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
```
hugobessa, line 712: Event-scope binding for the calendar reschedule variant: the event must be in the org AND on the named calendar (`calendar_fk_id`) AND NOT grouped (`calendar_group_fk_id__isnull=True`). The last clause is deliberate — a grouped event's `calendar_fk` is its group's primary calendar, so without it a calendar-shaped code could bind to a group booking and the Phase 6 single-calendar consume path would mishandle it. Group bookings get codes via the group variants only.

calendar_integration/mutations.py
```diff
@@ -637,3 +662,249 @@ def create_calendar_group_booking_code(
...
+            organization_id=org.id,
+            permissions=[EventManagementPermissions.RESCHEDULE],
+            expires_at=input.expires_at,
+            minted_by=minted_by,
+            calendar_id=input.calendar_id,
+            event_id=input.event_id,
+        )
+        return BookingCodeResult(success=True, code=plaintext_code, id=token.pk)
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
```
hugobessa, line 728: Reschedule mutations mint exactly `[EventManagementPermissions.RESCHEDULE]`; the cancellation mutations below mint exactly `[CANCEL]`. One permission per code, `event_id` always set — these are single-event, single-action codes. A test (`TestPermissionNotSwapped`) guards against an accidental reschedule/cancel swap in this copy-paste-shaped block.

calendar_integration/mutations.py
```diff
@@ -429,6 +434,26 @@ class CreateGroupBookingCodeInput:
...
+
+@strawberry.input
+class CreateGroupEventCodeInput:
+    """Input for minting a single-use reschedule or cancel code scoped to a calendar group + event."""
+
+    organization_id: int
+    calendar_group_id: int
+    event_id: int
+    expires_at: datetime.datetime | None = None
+
```
hugobessa, line 455: Two shared input types: `CreateEventCodeInput` (calendar + event) for both calendar variants, `CreateGroupEventCodeInput` (group + event) for both group variants — the reschedule and cancel inputs are structurally identical, so they share a type rather than duplicating four.

PR #102, opened 2026-06-18
feat(calendar): revoke single-use booking codes (Phase 3)

hugobessa:
## Summary

Adds the org-token-gated `revokeBookingCode(input: { organizationId, id })` mutation plus a
`CalendarPermissionService.revoke_token(organization_id, token_id)` method. A provider/admin
invalidates an unused (or any) code by its opaque id; the code then fails all downstream use
(reads + writes) as `REVOKED`.

- Org-scoped fetch (`filter_by_organization`) — cannot revoke another org's token.
- Idempotent: revoking an already-revoked code returns success and does NOT move `revoked_at`.
- Unknown id / cross-org id / `organizationId` mismatch all return `success=False`,
  `errorCode=INVALID_CODE` with an identical generic `"Not found."` (no existence leak).
- Success returns `BookingCodeResult(success=True)` with NO `code` echoed back.

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 3.
- Spec: Decisions → Use-cases, Use-case 5 (provider revokes an unused code).

## Test plan

In-container (docker postgres): `public_api/tests/test_revoke_booking_code.py` — revoke sets
`revoked_at`; idempotent (timestamp unchanged); cross-org token untouched + INVALID_CODE; unknown
id INVALID_CODE; without-resource rejected (row untouched); organizationId mismatch INVALID_CODE.
All DB-verified. `ruff` clean · `makemigrations --check` no changes (Phase 0 added `revoked_at`) ·
`check --deploy` clean · `pytest -n auto` **2072 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-2`.

### DIFF
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -8,7 +8,7 @@ import strawberry
 from dependency_injector.wiring import Provide, inject
 from graphql import GraphQLError
 
-from calendar_integration.exceptions import CalendarGroupError
+from calendar_integration.exceptions import CalendarGroupError, InvalidTokenError
 from calendar_integration.graphql import (
     BookingCodeErrorCode,
     BookingCodeResult,
@@ -452,6 +452,14 @@ class CreateGroupEventCodeInput:
     calendar_group_id: int
     event_id: int
     expires_at: datetime.datetime | None = None
+
+
+@strawberry.input
+class RevokeBookingCodeInput:
+    """Input for revoking a single-use booking code."""
+
+    organization_id: int
+    id: int  # noqa: A002
 
 
 @strawberry.type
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_permission_service.py
```diff
@@ -652,3 +652,34 @@ def consume_code(self, token: CalendarManagementToken, source_ip: str) -> None:
...
+                id=token_id
+            )
+        except CalendarManagementToken.DoesNotExist as e:
+            raise InvalidTokenError("Token not found") from e
+
+        if token.revoked_at is None:
+            token.revoked_at = timezone.now()
+            token.save(update_fields=["revoked_at"])
+
+        return True
```
hugobessa, line 685: Revoke sets `revoked_at` — the SAME field Phase 0's `consume()` re-checks under SELECT FOR UPDATE (→ TokenRevokedError), `active()` filters on, and `get_token_error_code()` maps to REVOKED. So a revoke takes effect across every downstream path with no field drift. Idempotent: the `if token.revoked_at is None` guard means a repeat revoke is a no-op on the timestamp. Unlocked read-modify-write is safe here — consume's own lock + recheck serialises any revoke-vs-consume race.

calendar_integration/mutations.py
```diff
@@ -908,3 +916,42 @@ def create_calendar_group_cancellation_booking_code(
...
+        try:
+            deps.calendar_permission_service.revoke_token(organization_id=org.id, token_id=input.id)
+        except InvalidTokenError:
+            return BookingCodeResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
+                error_message="Not found.",
+            )
+
+        return BookingCodeResult(success=True)
```
hugobessa, line 957: Gated by CALENDAR_BOOKING_CODE (the `revokeBookingCode` mapping shipped in Phase 0; field name matches exactly so it is not silently ungated). Unknown/cross-org id (InvalidTokenError) and org mismatch both return the same generic INVALID_CODE "Not found." Success returns `BookingCodeResult(success=True)` with NO code — revoke must never echo a code back.

PR #103, opened 2026-06-18
Phase 2c — importResourceCalendars Public GraphQL mutation

hugobessa:
## Summary
Adds the `importResourceCalendars` mutation so admin integrators can trigger a Google Workspace resource (room/equipment) import for their organization.

Routes to the existing high-level `OrganizationService.request_rooms_sync(organization, requested_by, start_time, end_time)` — **not** the low-level `request_organization_calendar_resources_import`. `request_rooms_sync` resolves the org-level `GoogleCalendarServiceAccount`, authenticates the calendar service with it, raises `NoServiceAccountConfiguredError` when none is configured, and enqueues the async import. (The low-level import requires an authenticated service with a Google account, which a Public API SystemUser token does not carry — so delegating to the high-level method is the correct path.)

- `public_api/mutations.py`: `importResourceCalendars` mutation, `{success, errorMessage}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`, via the wired `organization_service` DI provider.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.IMPORT_RESOURCE_CALENDARS` + mapping.
- `public_api/migrations/0009_alter_resourceaccess_resource_name.py`: choices-only AlterField.

## Plan reference
Plan → Phase 2c (optional in the integration doc; included). No feature flag. Stacked on Phase 2b (PR #100).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2023 passed**. Tests: happy path (asserts `request_rooms_sync` called with the supplied org + time window), no-service-account → `success=False` (real `NoServiceAccountConfiguredError`), permission-denied. Google calls mocked — no network.


public_api/mutations.py
```diff
@@ -587,3 +606,48 @@ def disable_resource_calendar(
...
+                organization=org,
+                requested_by=None,
+                start_time=input.start_time,
+                end_time=input.end_time,
+            )
+        except NoServiceAccountConfiguredError:
+            return ImportResourceCalendarsResult(
+                success=False,
+                error_message="No Google service account configured for this organization.",
+            )
```
hugobessa, line 649: Delegates to the HIGH-LEVEL `OrganizationService.request_rooms_sync`, not the low-level `request_organization_calendar_resources_import`. The low-level call needs an *authenticated* calendar service (a Google service account in context), which a Public API SystemUser token has none of — it would 500. `request_rooms_sync` does the service-account lookup + authenticate and raises `NoServiceAccountConfiguredError` cleanly. `requested_by` is passed as `None` (the param is accepted but unused by the service; the caller here is a SystemUser, not a User).

public_api/mutations.py
```diff
@@ -587,3 +606,48 @@ def disable_resource_calendar(
...
+        except (CalendarIntegrationError, ValueError, DjangoValidationError) as e:
+            return ImportResourceCalendarsResult(success=False, error_message=str(e))
+
```
hugobessa, line 652: Catches `CalendarIntegrationError` in addition to `ValueError`/`ValidationError`/ `NoServiceAccountConfiguredError`: `request_rooms_sync` authenticates with the stored Google service account, which can raise auth subclasses of `CalendarIntegrationError` (e.g. malformed credentials). Converting those to `success=False` keeps a bad service account from surfacing as an unhandled 500.

PR #104, opened 2026-06-18
feat(public_api): add scoped_to_user FK and owner-derivation helper

hugobessa:
## Summary
Phase 0 (foundation) of the **Per-Owner-Scoped Public API Tokens** plan. Adds the storage and the
shared helper that every later phase consumes; ships **no reachable behavior change**. Existing
org-wide tokens (`scoped_to_user IS NULL`) are untouched, and no backfill is required.

## What changed
- `SystemUser.scoped_to_user`: new **nullable** FK to `users.User` (`on_delete=CASCADE`, indexed).
  `NULL` means "organization-wide token" — the legacy default — so every existing token keeps its
  current behavior. When set, the token may only read/write data on calendars that user owns.
- `public_api/scoping.py` → `scoped_calendar_ids(system_user, organization)`: derives the owner's
  calendar-id set. Returns **`None`** for an org-wide token (the unrestricted sentinel, keeping the
  null path a no-op) and a **set** (possibly empty) for a scoped token. Empty set ⇒ "scoped but
  owns nothing" ⇒ reads empty / writes not-found in later phases.
- `public_api/constants.py` → `PROVIDER_SCOPED_RESOURCES`: the fixed provider allow-list
  (`AVAILABLE_TIME`, `BLOCKED_TIME`, `CALENDAR_EVENT`, `AVAILABILITY_WINDOWS`, `UNAVAILABLE_WINDOWS`,
  `CALENDAR`) that Phases 2/3 will validate mint requests against. References enum members, not raw
  strings, so it can't silently drift if an enum value is renamed.
- `public_api/admin/system_user.py`: surfaces `scoped_to_user` read-only in the admin (owner is
  immutable per the plan's Guiding Decisions).

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 0 +
the Data Model Changes section. Shared scaffolding — no spec use-case yet.

## Review history
Layer-3 review found no BLOCKERs; two SHOULD-FIX + one NIT applied: a tautological exclusion test
was made behavioral (now asserts the owned calendar IS present AND the other-user calendar absent);
`PROVIDER_SCOPED_RESOURCES` moved below the enum and switched to member references; `.distinct()`
added to the ownership join. Branch rebased onto latest `main` (docker-compose skill updates).

## Test plan
- `public_api/tests/test_scoping.py` — asserts `scoped_calendar_ids` returns `None` for org-wide
  tokens; only the owner's calendar ids for scoped tokens; empty set for an owner with no calendars;
  and **never** another org's calendar even when the same user id owns calendars in two orgs
  (multi-tenant isolation).
- Outer gate green inside the worktree's api container: `check --deploy` (5 expected local
  `security.W` warnings only) + `pytest -n auto` → **2005 passed**.


public_api/scoping.py
```diff
@@ -0,0 +1,27 @@
...
+    Args:
+        system_user: The SystemUser (token) making the request.
+        organization: The organization context.
+
+    Returns:
+        None if the token is org-wide (scoped_to_user is None);
+        a set of calendar IDs (possibly empty) if scoped to a user.
+    """
+    if system_user.scoped_to_user_id is None:
+        return None
```
hugobessa: Load-bearing sentinel for the whole feature: `None` means "org-wide token, unrestricted" — NOT "no calendars". Every later phase's resolver/mutation treats `None` as a no-op (skip filtering) and a returned set (even empty) as "restrict to exactly these ids". Conflating empty-set with None here would silently widen a scoped token that happens to own nothing back to org-wide access. Keep these two return paths distinct.

public_api/scoping.py
```diff
@@ -0,0 +1,27 @@
...
+    return set(
+        Calendar.objects.filter_by_organization(organization.id)
+        .filter(ownerships__user_id=system_user.scoped_to_user_id)
```
hugobessa: Multi-tenant guard: `filter_by_organization(organization.id)` is applied BEFORE the ownership filter, so a user who owns calendars in two orgs can never leak org B's calendar ids into a token scoped within org A. `ownerships__user_id` is the `CalendarOwnership.user` edge named in the plan's Owner-edge decision; `.distinct()` collapses duplicate join rows (a user can hold multiple ownership rows). There's a dedicated test asserting the cross-org case.

public_api/constants.py
```diff
@@ -34,3 +34,15 @@ class PublicAPIResources(TextChoices):
...
+PROVIDER_SCOPED_RESOURCES: frozenset[str] = frozenset(
+    [
+        PublicAPIResources.AVAILABLE_TIME,
+        PublicAPIResources.BLOCKED_TIME,
+        PublicAPIResources.CALENDAR_EVENT,
+        PublicAPIResources.AVAILABILITY_WINDOWS,
+        PublicAPIResources.UNAVAILABLE_WINDOWS,
+        PublicAPIResources.CALENDAR,
+    ]
+)
```
hugobessa, line 66: This frozenset is the contract Phases 2 (createScopedSystemUser) and 3 (REST optional owner) validate mint requests against — a provider token may grant ONLY these six resources, so a caller can't over-grant. Defined with enum members (not raw strings) and placed below the enum so a rename of any `PublicAPIResources` value updates this set in lockstep instead of drifting.

public_api/models.py
```diff
@@ -19,6 +19,18 @@ class SystemUser(BaseModel):
...
+        related_name="scoped_system_users",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        db_index=True,
+        help_text=(
+            "When set, this token may only read/write data belonging to calendars "
+            "owned by this user. NULL = organization-wide token (legacy default)."
+        ),
+    )
```
hugobessa: Nullable by design: `NULL` is the backward-compat sentinel = today's org-wide token, which is why no data backfill is needed and existing tokens are byte-for-byte unchanged. `on_delete= CASCADE` because a scoped token is meaningless once its owner is deleted (it could never reach any data afterward). The field is intentionally immutable post-creation — surfaced read-only in admin and never accepted by the REST grant-edit path in Phase 3.

hugobessa:

PR #105, opened 2026-06-18
feat(public_api): scope read queries to token owner's calendars

hugobessa:
## Summary
Phase 1 of the **Per-Owner-Scoped Public API Tokens** plan. A scoped token's reads now return only
its owner's calendar-derived data; org-wide tokens (`scoped_to_user IS NULL`) are byte-for-byte
unchanged. Enforcement is data-gated — no feature flag — and lives in the read resolvers, not the
boolean permission class, because list queries without a `calendarId` are the leak-prone case.

## What changed (all in `public_api/queries.py`)
- Six read resolvers now consume the Phase 0 helper `scoped_calendar_ids(system_user, org)`:
  `calendars`, `calendar_events`, `blocked_times`, `available_times`, `availability_windows`,
  `unavailable_windows`. Helper returns `None` (org-wide → no-op) or a set (restrict to those ids;
  empty set → empty reads).
- `calendars`: the list is constrained with `id__in=allowed_ids` (the no-`calendarId` leak case).
- `_prepare_service_and_calendar`: a scoped token requesting a cross-owner `calendar_id` now raises
  the SAME `Calendar.DoesNotExist` a genuinely missing calendar raises — no existence leak. Covers
  the `calendarId`+range paths of `calendar_events`/`blocked_times`/`available_times` and both
  window resolvers.
- Single-id lookups (`event_id`/`blocked_time_id`/`available_time_id`) filter `calendar_fk__in=
  allowed_ids` → empty on cross-owner, not an error.
- **Post-expansion filter (leak fix):** the service expansion (`get_*_expanded`) can return
  bundle-representation rows living on OTHER providers' child calendars even when the requested
  calendar is owned. The expanded result lists are now filtered to `calendar_fk_id ∈ allowed_ids`
  (fail-closed on a `None` calendar ref).

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 1.
Spec Decisions → Use-cases: "Provider token manages its owner's availability and schedule" (reads)
and "Provider token attempts to reach another provider's data" (reads).

## Review history
Layer-3 review caught a **BLOCKER**: the `calendarId`-path expansion (`get_*_expanded`) leaked
bundle-representation rows on other owners' calendars despite the requested calendar being owned.
Fixed with the post-expansion `calendar_fk_id ∈ allowed_ids` filter; a behavioral test was added
that fails when the filter is removed (verified). Also de-vacuumed three cross-owner-`calendarId`
tests that previously passed via an unreached assertion, and dropped a dead client line. Nested
GraphQL field traversal (`bundle_representations`, `resources`, `recurring_instances` on
`CalendarEventGraphQLType`) is deliberately left to the Phase 5 adversarial sweep.

## Test plan
- `public_api/tests/test_queries.py` — scoped token sees only its owner's calendars/events/blocked/
  available/windows; cross-owner `calendarId` is indistinguishable from a nonexistent `calendarId`;
  cross-owner single-id returns empty; owner with no calendars reads empty everywhere; **org-wide
  token returns results identical to pre-change** (per resolver); post-expansion filter strips
  other-calendar rows.
- Outer gate green inside the worktree api container: `check --deploy` (5 expected local W-warnings)
  + `pytest -n auto` → **2033 passed**. `mypy` clean on the touched modules.

### DIFF
--- a/public_api/queries.py
+++ b/public_api/queries.py
@@ -50,6 +50,7 @@ from public_api.permissions import (
     IsAuthenticated,
     OrganizationResourceAccess,
 )
+from public_api.scoping import scoped_calendar_ids
 from public_api.types import (
     ChildOrganizationMetrics,
     PublicApiHttpRequest,
@@ -178,6 +179,15 @@ def _prepare_service_and_calendar(
     deps.calendar_service.initialize_without_provider(
         user_or_token=request.public_api_system_user, organization=org
     )
+
+    # Owner-scope check: when the token is scoped, reject calendars outside the owner's set.
+    # Match the same not-found path used for a genuinely missing calendar (no existence leak).
+    system_user = request.public_api_system_user
+    if system_user is not None:
+        allowed_ids = scoped_calendar_ids(system_user, org)
+        if allowed_ids is not None and calendar_id not in allowed_ids:
+            raise Calendar.DoesNotExist("Calendar matching query does not exist.")
+
     cal = Calendar.objects.filter_by_organization(org.id).get(id=calendar_id)
     return deps.calendar_service, cal
 
@@ -277,6 +287,16 @@ class Query:
             raise GraphQLError("Limit must be between 1 and 100")
 
         queryset = Calendar.objects.filter_by_organization(org.id).only_listed()
+
+        # Owner-scope enforcement: scoped tokens may only see their owner's calendars.
+        # None => org-wide token (no-op). A set (possibly empty) => constrain to those ids.
+        request: PublicApiHttpRequest = info.context.request
+        system_user = request.public_api_system_user
+        if system_user is not None:
+            allowed_ids = scoped_calendar_ids(system_user, org)
+            if allowed_ids is not None:
+                queryset = queryset.filter(id__in=allowed_ids)
+
         if calendar_id is not None:
             queryset = queryset.filter(id=calendar_id)
 
@@ -308,7 +328,16 @@ class Query:
         org = _get_org(info)
 
         if event_id is not None:
-            return CalendarEvent.objects.filter_by_organization(org.id).filter(id=event_id)
+            qs = CalendarEvent.objects.filter_by_organization(org.id).filter(id=event_id)
+            # Owner-scope: for scoped tokens, only return the event if its calendar is in the
+            # owner's set. Return empty (not an error) to avoid existence leaks.
+            request: PublicApiHttpRequest = info.context.request
+            system_user = request.public_api_system_user
+            if system_user is not None:
+                allowed_ids = scoped_calendar_ids(system_user, org)
+                if allowed_ids is not None:
+                    qs = qs.filter(calendar_fk__in=allowed_ids)
+            return qs  # type: ignore[return-value]
 
         if not calendar_id or not start_datetime or not end_datetime:
             raise GraphQLError(
@@ -322,6 +351,15 @@ class Query:
             start_datetime,
             end_datetime,
         )
+
+        request: PublicApiHttpRequest = info.context.request
+        allowed_ids = (
+            scoped_calendar_ids(request.public_api_system_user, org)
+            if request.public_api_system_user is not None
+            else None
+        )
+        if allowed_ids is not None:
+            events = [e for e in events if getattr(e, "calendar_fk_id", None) in allowed_ids]
 
         return cast(
             list[CalendarEventGraphQLType],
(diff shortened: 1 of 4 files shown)


public_api/queries.py
```diff
@@ -114,6 +115,15 @@ def _prepare_service_and_calendar(
...
         user_or_token=request.public_api_system_user, organization=org
     )
+
+    # Owner-scope check: when the token is scoped, reject calendars outside the owner's set.
+    # Match the same not-found path used for a genuinely missing calendar (no existence leak).
+    system_user = request.public_api_system_user
+    if system_user is not None:
+        allowed_ids = scoped_calendar_ids(system_user, org)
+        if allowed_ids is not None and calendar_id not in allowed_ids:
+            raise Calendar.DoesNotExist("Calendar matching query does not exist.")
```
hugobessa, line 189: Cross-owner reads must not reveal that another provider's object exists. A scoped token passing a calendar_id it doesn't own raises the SAME `Calendar.DoesNotExist` a genuinely missing calendar raises — identical error shape, no existence leak. This single guard covers the calendarId+range paths of calendar_events / blocked_times / available_times and both window resolvers, since they all route through here. Org-wide tokens (`scoped_to_user is None`) skip the guard entirely.

public_api/queries.py
```diff
@@ -148,6 +158,16 @@ def calendars(
...
         queryset = Calendar.objects.filter_by_organization(org.id).only_listed()
+
+        # Owner-scope enforcement: scoped tokens may only see their owner's calendars.
+        # None => org-wide token (no-op). A set (possibly empty) => constrain to those ids.
+        request: PublicApiHttpRequest = info.context.request
+        system_user = request.public_api_system_user
+        if system_user is not None:
+            allowed_ids = scoped_calendar_ids(system_user, org)
+            if allowed_ids is not None:
+                queryset = queryset.filter(id__in=allowed_ids)
```
hugobessa, line 298: The leak-prone case: `calendars` is the one resolver that lists across the org with no required calendar_id argument. Without this `id__in=allowed_ids` a scoped token would enumerate every provider's calendars. The org filter stays intact — owner scoping is ADDITIONAL to org scoping, never a replacement. An empty `allowed_ids` (owner owns nothing) correctly yields an empty list, not an unfiltered one.

public_api/queries.py
```diff
@@ -194,6 +223,15 @@ def calendar_events(
...
         )
 
+        request: PublicApiHttpRequest = info.context.request
+        allowed_ids = (
+            scoped_calendar_ids(request.public_api_system_user, org)
+            if request.public_api_system_user is not None
+            else None
+        )
+        if allowed_ids is not None:
+            events = [e for e in events if getattr(e, "calendar_fk_id", None) in allowed_ids]
```
hugobessa, line 362: BLOCKER fix from review. `_prepare_service_and_calendar` already rejected a cross-owner *requested* calendar, but `get_calendar_events_expanded` can still return bundle-representation rows that live on OTHER providers' child calendars (shared-bundle primaries). Those rows have a `calendar_fk` outside the owner's set, so they're stripped here. Fail-closed: a row with a missing/None calendar ref is excluded. A behavioral test asserts this and fails if the line is removed. Same filter is applied to the blocked-times and available-times expansions.

public_api/queries.py
```diff
@@ -179,7 +199,16 @@ def calendar_events(
...
-            return CalendarEvent.objects.filter_by_organization(org.id).filter(id=event_id)
+            qs = CalendarEvent.objects.filter_by_organization(org.id).filter(id=event_id)
+            # Owner-scope: for scoped tokens, only return the event if its calendar is in the
+            # owner's set. Return empty (not an error) to avoid existence leaks.
+            request: PublicApiHttpRequest = info.context.request
+            system_user = request.public_api_system_user
+            if system_user is not None:
+                allowed_ids = scoped_calendar_ids(system_user, org)
+                if allowed_ids is not None:
+                    qs = qs.filter(calendar_fk__in=allowed_ids)
```
hugobessa, line 339: Single-id lookups (event_id / blocked_time_id / available_time_id) filter `calendar_fk__in= allowed_ids` rather than raising — a scoped token asking for another owner's object id gets an empty result, again indistinguishable from "doesn't exist". Reads return empty; only the calendar-resolution path (above) raises, matching the pre-existing missing-calendar behavior.

hugobessa:

PR #106, opened 2026-06-18
feat(public_api): add createScopedSystemUser mutation

hugobessa:
## Summary
Phase 2 of the **Per-Owner-Scoped Public API Tokens** plan. Adds the GraphQL mutation the Medplum
bot uses to mint a provider-scoped token: it scopes the new token to one provider (`scoped_to_user`)
and grants only provider-permitted resources. The mutation is itself gated by the `SYSTEM_USER`
resource, so a provider token (which never holds `SYSTEM_USER`) cannot mint more tokens.

## What changed
- `public_api/mutations.py` → `createScopedSystemUser`: `[IsAuthenticated, OrganizationResourceAccess]`
  guarded (mapped to `SYSTEM_USER`). Validation order: resolve caller org → owner must be an
  **active member of the caller's org** → `available_resources` non-empty → every value a valid
  `PublicAPIResources` → every value ∈ `PROVIDER_SCOPED_RESOURCES` (no over-grant) → atomic create +
  `ResourceAccess` grants → duplicate `integration_name` rejected. Plaintext token returned **once**.
- `public_api/types.py`: `CreateScopedSystemUserInput` (`integration_name`, `scoped_to_user_id`,
  `available_resources`) + `CreateScopedSystemUserResult` (adds `scoped_to_user_id` + write-once `token`).
- `public_api/services.py`: `create_system_user` gains an optional `scoped_to_user` (default `None` →
  existing callers unchanged).
- `public_api/permissions.py`: `createScopedSystemUser → SYSTEM_USER` mapping.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 2 + the
API Design `createScopedSystemUser` subsection. Spec Decisions → Use-cases: "Bot mints a
provider-scoped token on provider creation". Non-goals respected: no patient shape, no REST change.

## Review history
Layer-3 review found no BLOCKERs. Applied: narrowed the `IntegrityError` handling so only an
`integration_name` uniqueness violation reports "already exists" (a `ResourceAccess` constraint no
longer mislabels); response `scoped_to_user_id` now sourced from the persisted row; type annotations
added on the service. Two security tests added on review: a **scoped provider token cannot mint**
(escalation proof) and an **inactive-membership owner is rejected**.

## Test plan
- `public_api/tests/test_mutations.py` — happy mint (token once; persisted `scoped_to_user` + grants
  correct); caller lacking `SYSTEM_USER` denied; scoped provider token denied (no self-escalation);
  owner not in caller's org / nonexistent / inactive-member rejected with no token created;
  over-grant rejected; duplicate `integration_name` rejected; empty resource list rejected.
- Outer gate (docker): `check --deploy` (5 expected W-warnings) + `pytest -n auto` → **2043 passed**;
  mypy clean on touched modules.

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -751,15 +751,15 @@ class Mutation(CalendarGroupMutations):
         if not org:
             raise GraphQLError("Organization not found")
 
-        # Validate owner: must exist and be an active member of the caller's org
-        user_model = get_user_model()
+        # Validate owner: resolve the active membership of the given user in the caller's org.
+        # This single query both validates active membership AND yields the value to store.
         try:
-            owner = user_model.objects.get(
-                id=input.scoped_to_user_id,
-                organization_memberships__organization=org,
-                organization_memberships__is_active=True,
+            membership = OrganizationMembership.objects.get(
+                user_id=input.scoped_to_user_id,
+                organization=org,
+                is_active=True,
             )
-        except user_model.DoesNotExist as e:
+        except OrganizationMembership.DoesNotExist as e:
             raise GraphQLError(
                 f"User with id '{input.scoped_to_user_id}' is not an active member of "
                 "the caller's organization."
@@ -792,7 +792,7 @@ class Mutation(CalendarGroupMutations):
                 system_user, plaintext_token = deps.public_api_auth_service.create_system_user(
                     integration_name=input.integration_name,
                     organization=org,
-                    scoped_to_user=owner,
+                    scoped_to_membership=membership,
                 )
                 # dict.fromkeys dedupes while preserving order; prevents constraint violations
                 ResourceAccess.objects.bulk_create(
@@ -817,15 +817,15 @@ class Mutation(CalendarGroupMutations):
             )
         )
 
-        # scoped_to_user_id is always set here — we passed owner to create_system_user above.
-        assert system_user.scoped_to_user_id is not None  # noqa: S101
+        # scoped_to_membership_fk_id is always set here — we passed membership to create_system_user above.
+        assert system_user.scoped_to_membership_fk_id is not None  # noqa: S101
 
         return CreateScopedSystemUserResult(
             id=system_user.id,
             integration_name=system_user.integration_name,
             is_active=system_user.is_active,
             available_resources=granted_resources,
-            scoped_to_user_id=system_user.scoped_to_user_id,
+            scoped_to_user_id=membership.user_id,
             token=plaintext_token,
         )
 
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -28,7 +28,7 @@ from organizations.exceptions import NoServiceAccountConfiguredError, UserAlread
 from organizations.models import Organization, OrganizationBranding, OrganizationMembership
 from organizations.services import OrganizationService
 from public_api.capabilities import assert_org_can_invite, assert_target_in_subtree
-from public_api.constants import PublicAPIResources
+from public_api.constants import PROVIDER_SCOPED_RESOURCES, PublicAPIResources
 from public_api.models import ResourceAccess, SystemUser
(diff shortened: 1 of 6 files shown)


public_api/permissions.py
```diff
@@ -46,6 +46,7 @@ class OrganizationResourceAccess(BasePermission):
         # supports one resource per field; INVITATION is the primary gating resource.
         "createInvitation": PublicAPIResources.INVITATION,
         "createSystemUserToken": PublicAPIResources.SYSTEM_USER,
+        "createScopedSystemUser": PublicAPIResources.SYSTEM_USER,
```
hugobessa, line 49: This single mapping is what prevents mint-time privilege escalation: createScopedSystemUser is gated by the SYSTEM_USER resource, and SYSTEM_USER is deliberately NOT in PROVIDER_SCOPED_RESOURCES (Phase 0). So a provider-scoped token can never satisfy this permission and can never mint another token — there's a dedicated test asserting exactly that.

public_api/mutations.py
```diff
@@ -357,6 +359,111 @@ def create_system_user_token(
...
+
+        # Validate owner: must exist and be an active member of the caller's org
+        user_model = get_user_model()
+        try:
+            owner = user_model.objects.get(
+                id=input.scoped_to_user_id,
+                organization_memberships__organization=org,
+                organization_memberships__is_active=True,
+            )
+        except user_model.DoesNotExist as e:
```
hugobessa: Owner resolution doubles as the cross-org guard: the lookup requires an ACTIVE membership in the CALLER's org (`org` comes from request.public_api_organization, not from input), so a user id from another org, a nonexistent id, or an inactive-membership user all fail here identically — a token can never be scoped cross-org. `OrganizationMembership` has a unique (user, org) constraint, so `.get()` can't raise MultipleObjectsReturned.

public_api/mutations.py
```diff
@@ -357,6 +359,111 @@ def create_system_user_token(
...
+                f"Invalid resource(s): {', '.join(invalid_resources)}. "
+                f"Valid values are: {', '.join(sorted(valid_values))}."
+            )
+
+        # Validate each resource is within the provider allow-list (no over-grant)
+        over_grant = [r for r in input.available_resources if r not in PROVIDER_SCOPED_RESOURCES]
+        if over_grant:
+            raise GraphQLError(
+                f"Resource(s) not permitted for provider-scoped tokens: {', '.join(over_grant)}. "
+                f"Allowed resources are: {', '.join(sorted(PROVIDER_SCOPED_RESOURCES))}."
```
hugobessa, line 786: The over-grant gate: even a legitimate SYSTEM_USER caller may only grant resources in the fixed PROVIDER_SCOPED_RESOURCES allow-list. Passing USER / SYSTEM_USER / anything outside the provider set is rejected before any row is written, so a minted provider token's blast radius is bounded by construction, not by caller discipline.

public_api/mutations.py
```diff
@@ -357,6 +359,111 @@ def create_system_user_token(
...
+                )
+        except IntegrityError as e:
+            # Only convert to "already exists" when the integration_name uniqueness constraint
+            # fired; a ResourceAccess constraint failure would have a different message and
+            # must not be silently mislabeled.
+            if "integration_name" in str(e).lower():
+                raise GraphQLError(
+                    f"A token with integration_name '{input.integration_name}' already exists."
+                ) from e
+            raise
```
hugobessa, line 812: Narrowed on review: only an integration_name uniqueness violation is converted to the friendly "already exists" error; any other IntegrityError (e.g. the ResourceAccess unique constraint) is re-raised rather than mislabeled. The whole create stays inside one transaction.atomic(), so a duplicate name leaves no half-created SystemUser.

hugobessa:

PR #107, opened 2026-06-18
feat(public_api): accept optional owner scope on REST token create

hugobessa:
## Summary
Phase 3 of the **Per-Owner-Scoped Public API Tokens** plan. The admin REST endpoint
`POST /public-api-tokens/` now accepts an optional `scoped_to_user`, minting a provider-scoped token
with the same rules as the GraphQL mutation. The no-owner path is held byte-for-byte. Owner is
creation-only, and a scoped token's grants can't be widened past the provider allow-list via update.

## What changed
- `SystemUserTokenCreateSerializer`: optional `scoped_to_user` (int). `validate()` — ONLY when it's
  non-null — resolves the owner (must be an ACTIVE member of the caller's org) and enforces
  `available_resources ⊆ PROVIDER_SCOPED_RESOURCES`; `create()` passes the owner to
  `create_system_user`. When absent/null, behavior is exactly as before (org-wide, any valid resource).
- Response + list serializers expose read-only **nullable** `scoped_to_user` (the FK id; null for
  org-wide tokens).
- `SystemUserTokenViewSet.update`/`partial_update`: a **scoped** token can no longer be granted
  resources outside `PROVIDER_SCOPED_RESOURCES` (closes a post-creation escalation path); org-wide
  tokens keep accepting any valid resource. Owner itself remains immutable (the update serializer has
  no owner field).
- `schema.yml` regenerated.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 3 + the
"REST POST /public-api-tokens/ — optional owner" API Design subsection. Spec Decisions → Use-cases:
"Admin mints a scoped token via REST". Non-goals respected: no patient shape, no GraphQL change.

## Review history
Layer-3 review found no BLOCKERs. Applied: a **post-creation escalation guard** (a scoped token
can't add `SYSTEM_USER`/non-provider resources via PUT/PATCH — undermined Goal 5 otherwise); marked
the response `scoped_to_user` `nullable` so the schema doesn't lie for org-wide tokens; added an
explicit-`null`-owner backward-compat test and a dead-default cleanup. Validation parity with the
Phase 2 GraphQL mutation confirmed.

Deferred (pre-existing, tracked separately): `SystemUserTokenViewSet` does not extend
`TenantScopedViewMixin`, so a multi-org admin without `X-Organization-Id` resolves to their oldest
membership. No cross-org escalation (owner is validated against the same caller-resolved org), so
this is a correctness follow-up on the endpoint, not introduced or worsened here.

## Test plan
- `public_api/tests/test_views.py` — scoped create (owner echoed + persisted, token once); no-owner
  AND explicit-null create unchanged (org-wide, `user` resource still accepted → allow-list didn't
  leak); over-grant-with-owner 400 (no row); out-of-org / nonexistent owner 400 (no row); PUT & PATCH
  can't change owner; **update can't add non-provider resources to a scoped token** (400, grants
  unchanged) while org-wide update still accepts any resource.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2059 passed**; mypy clean on touched
  modules; `makemigrations --check` clean; `schema.yml` regenerated.

### DIFF
--- a/public_api/serializers.py
+++ b/public_api/serializers.py
@@ -1,6 +1,5 @@
 from typing import Annotated
 
-from django.contrib.auth import get_user_model
 from django.db import IntegrityError, transaction
 
 from dependency_injector.wiring import Provide, inject
@@ -51,10 +50,14 @@ class SystemUserTokenCreateSerializer(serializers.Serializer):
 
     def validate(self, attrs: dict) -> dict:
         """Cross-field validation: when scoped_to_user is present and non-null, resolve the
-        owner User (must be an active member of the caller's org) and enforce the provider
+        active OrganizationMembership of that user in the caller's org and enforce the provider
         allow-list on available_resources.  When scoped_to_user is absent or null, no
         additional constraints are applied — the no-owner path is byte-for-byte identical
         to the pre-Phase-3 behaviour.
+
+        The input field ``scoped_to_user`` is a User id (external REST API contract).  Internally
+        the membership FK is resolved and stashed as ``_resolved_membership``; only the membership
+        is ever stored — the User id is derived from it on read.
         """
         scoped_to_user_id: int | None = attrs.get("scoped_to_user")
         if scoped_to_user_id is None:
@@ -62,19 +65,21 @@ class SystemUserTokenCreateSerializer(serializers.Serializer):
             return attrs
 
         request = self.context["request"]
-        membership: OrganizationMembership | None = get_active_organization_membership(request.user)
-        if membership is None:
+        caller_membership: OrganizationMembership | None = get_active_organization_membership(
+            request.user
+        )
+        if caller_membership is None:
             raise PermissionDenied("No active organisation membership.")
 
-        # Resolve the owner: must exist AND be an active member of the caller's org.
-        user_model = get_user_model()
+        # Resolve the owner: the target user must be an active member of the caller's org.
+        # This single query both validates active membership AND yields the value to store.
         try:
-            owner = user_model.objects.get(
-                id=scoped_to_user_id,
-                organization_memberships__organization=membership.organization,
-                organization_memberships__is_active=True,
+            resolved_membership = OrganizationMembership.objects.get(
+                user_id=scoped_to_user_id,
+                organization=caller_membership.organization,
+                is_active=True,
             )
-        except user_model.DoesNotExist as e:
+        except OrganizationMembership.DoesNotExist as e:
             raise serializers.ValidationError(
                 {
                     "scoped_to_user": [
@@ -98,8 +103,8 @@ class SystemUserTokenCreateSerializer(serializers.Serializer):
                 }
             )
 
-        # Stash the resolved User instance so create() can pass it to create_system_user.
-        attrs["_resolved_owner"] = owner
+        # Stash the resolved membership so create() can pass it to create_system_user.
+        attrs["_resolved_membership"] = resolved_membership
         return attrs
 
     def create(self, validated_data: dict) -> SystemUser:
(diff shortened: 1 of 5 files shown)


public_api/serializers.py
```diff
@@ -41,6 +49,59 @@ def __init__(
...
 
+    def validate(self, attrs: dict) -> dict:
+        """Cross-field validation: when scoped_to_user is present and non-null, resolve the
+        owner User (must be an active member of the caller's org) and enforce the provider
+        allow-list on available_resources.  When scoped_to_user is absent or null, no
+        additional constraints are applied — the no-owner path is byte-for-byte identical
+        to the pre-Phase-3 behaviour.
+        """
+        scoped_to_user_id: int | None = attrs.get("scoped_to_user")
+        if scoped_to_user_id is None:
```
hugobessa, line 63: The whole backward-compat contract hinges on this early return: when scoped_to_user is null the serializer applies ZERO new constraints, so existing org-wide callers (incl. ones granting non-provider resources like `user`) behave byte-for-byte as before. The allow-list + owner resolution only run on the explicit-owner path. There are tests for both omitting the field and sending it as null.

public_api/serializers.py
```diff
@@ -41,6 +49,59 @@ def __init__(
...
+                    "scoped_to_user": [
+                        f"User with id '{scoped_to_user_id}' is not an active member of "
+                        "the caller's organization."
+                    ]
+                }
+            ) from e
+
+        # Enforce provider allow-list: every requested resource must be in PROVIDER_SCOPED_RESOURCES.
+        available_resources: list[str] = attrs["available_resources"]
+        over_grant = [r for r in available_resources if r not in PROVIDER_SCOPED_RESOURCES]
```
hugobessa, line 94: Parity with the Phase 2 GraphQL mutation: a REST scoped mint may only grant resources in PROVIDER_SCOPED_RESOURCES, validated against the SAME caller-resolved org as the owner lookup above. REST is never more permissive than GraphQL.

public_api/views.py
```diff
@@ -142,6 +144,20 @@ def update(self, request: Request, *args, **kwargs) -> Response:
...
+        # Guard: scoped tokens may not be updated with resources outside PROVIDER_SCOPED_RESOURCES.
+        if system_user.scoped_to_user_id is not None:
+            over_grant = [r for r in desired_resources if r not in PROVIDER_SCOPED_RESOURCES]
+            if over_grant:
+                raise ValidationError(
+                    {
+                        "available_resources": [
+                            f"Resource(s) not permitted for provider-scoped tokens: "
+                            f"{', '.join(over_grant)}. "
+                            f"Allowed resources are: {', '.join(sorted(PROVIDER_SCOPED_RESOURCES))}."
```
hugobessa, line 161: Added on review — closes a post-creation escalation path. Without this, the existing grant-edit endpoint could PUT/PATCH a SCOPED token to include SYSTEM_USER (which would even let it mint tokens) or any other non-provider resource, escaping the ceiling enforced at mint time. Guard is scoped-only: org-wide tokens (`scoped_to_user_id is None`) keep accepting any valid resource, so existing admin token-editing is unchanged. Owner stays immutable independently (the update serializer has no owner field).

public_api/serializers.py
```diff
@@ -84,15 +149,26 @@ class SystemUserTokenResponseSerializer(serializers.ModelSerializer):
...
     Includes the write-once ``token`` field (sourced from the view) and
     ``available_resources`` (derived from the related ``ResourceAccess`` rows).
+    Exposes ``scoped_to_user`` as the FK id (null for org-wide tokens).
     Never exposes ``long_lived_token_hash``.
     """
 
     available_resources = serializers.SerializerMethodField()
     token = serializers.CharField(read_only=True)
+    scoped_to_user: serializers.PrimaryKeyRelatedField = serializers.PrimaryKeyRelatedField(
+        read_only=True, allow_null=True
```
hugobessa: `allow_null=True` so drf-spectacular emits `nullable: true` — org-wide tokens legitimately return a null owner, and the generated schema must not claim otherwise. Same on the list serializer.

hugobessa:

PR #108, opened 2026-06-18
feat(public_api): add owner-guarded createAvailableTime mutation

hugobessa:
## Summary
Phase 4a of the **Per-Owner-Scoped Public API Tokens** plan. Adds the first provider write mutation:
`createAvailableTime`. A scoped token can set its owner's recurring availability and specific
availability dates; it can never write to another provider's calendar. Org-wide tokens are unaffected.

## What changed
- `public_api/mutations.py` → `createAvailableTime`: `[IsAuthenticated, OrganizationResourceAccess]`
  guarded (mapped to `AVAILABLE_TIME`). Args `calendar_id`, `start_time`, `end_time`, `timezone`,
  optional `rrule_string` (omit → one-off date; provide → recurring). Owner guard + service init are
  delegated to the shared `prepare_service_and_calendar` helper, which raises `Calendar.DoesNotExist`
  for a cross-owner calendar (re-wrapped as an indistinguishable "does not exist" error — no leak).
  Service `ValueError` (e.g. a calendar that doesn't manage availability windows) is converted to a
  clean GraphQL error rather than a 500. Returns `AvailableTimeGraphQLType`.
- `public_api/helpers.py` (new): the owner-guard/service-init helper `prepare_service_and_calendar`
  (plus `get_org` + the query-deps accessor) was promoted out of `queries.py` so it's a proper shared
  module — `queries.py` and `mutations.py` (and 4b/4c) import it. No behavior change; pure move.
- `public_api/permissions.py`: `createAvailableTime → AVAILABLE_TIME` mapping.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 4a + the
availability and schedule" (availability write). Non-goals respected: only the available-time
mutation here (blocked-time = 4b, schedule-event = 4c); no new resource enums.

## Review history
Layer-3 review caught a **BLOCKER**: a `ValueError` from the service (calendar not managing
availability windows) surfaced as a 500 — now converted to a GraphQL error, with a test. Also
promoted the cross-module `_`-private helper into `public_api/helpers.py` (4b/4c reuse it), renamed
the `timezone` param to avoid shadowing `django.utils.timezone` while keeping the GraphQL field name
`timezone` via a Strawberry argument alias, and hardened the cross-owner test to assert no row is
written.

## Test plan
- `public_api/tests/test_mutations.py` — scoped token creates a one-off and a recurring (rrule)
  available time on its owned calendar; cross-owner `calendar_id` → not-found indistinguishable from a
  missing calendar AND no `AvailableTime` row written; org-wide token writes on any org calendar;
  calendar-not-managing-windows → clean error (not 500), no row; missing `AVAILABLE_TIME` resource denied.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2066 passed**; mypy clean on touched
  modules (3 pre-existing `no-redef` annotations in `queries.py` from Phase 1, tracked for the Phase 5
  sweep).

### DIFF
--- a/public_api/helpers.py
+++ b/public_api/helpers.py
@@ -1,0 +1,87 @@
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Annotated, cast
+
+import strawberry
+from dependency_injector.wiring import Provide, inject
+from graphql import GraphQLError
+
+from calendar_integration.models import Calendar
+from public_api.scoping import scoped_calendar_ids
+from public_api.types import PublicApiHttpRequest
+
+
+if TYPE_CHECKING:
+    from calendar_integration.services.calendar_group_service import CalendarGroupService
+    from calendar_integration.services.calendar_service import CalendarService
+
+
+@dataclass
+class QueryDependencies:
+    calendar_service: "CalendarService"
+    calendar_group_service: "CalendarGroupService"
+
+
+@inject
+def get_query_dependencies(
+    calendar_service: Annotated["CalendarService | None", Provide["calendar_service"]] = None,
+    calendar_group_service: Annotated[
+        "CalendarGroupService | None", Provide["calendar_group_service"]
+    ] = None,
+) -> QueryDependencies:
+    required_dependencies = [calendar_service, calendar_group_service]
+    if any(dep is None for dep in required_dependencies):
+        raise GraphQLError(
+            f"Missing required dependency {', '.join([str(dep) for dep in required_dependencies if dep is None])}"
+        )
+
+    return QueryDependencies(
+        calendar_service=cast("CalendarService", calendar_service),
+        calendar_group_service=cast("CalendarGroupService", calendar_group_service),
+    )
+
+
+def get_org(info: strawberry.Info):
+    org = info.context.request.public_api_organization
+    if not org:
+        raise GraphQLError("Organization not found in request context")
+    return org
+
+
+def prepare_service_and_calendar(
+    info: strawberry.Info,
+    calendar_id: int,
+    _deps: "QueryDependencies | None" = None,
+) -> tuple["CalendarService", Calendar]:
+    """Resolve org, initialize calendar_service, apply owner-scope guard, and fetch calendar.
+
+    Args:
+        info: Strawberry resolver info carrying the request context.
+        calendar_id: ID of the calendar to fetch.
+        _deps: Optional pre-resolved QueryDependencies (used by callers that already hold
+            a deps instance, e.g. to support test mocking via the caller's DI accessor).
+            When omitted, get_query_dependencies() is called internally.
+
+    Returns:
+        Tuple of (CalendarService, Calendar).
+
+    Raises:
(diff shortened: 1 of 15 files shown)


public_api/helpers.py
```diff
@@ -0,0 +1,87 @@
...
+        user_or_token=request.public_api_system_user, organization=org
+    )
+
+    # Owner-scope check: when the token is scoped, reject calendars outside the owner's set.
+    # Match the same not-found path used for a genuinely missing calendar (no existence leak).
+    system_user = request.public_api_system_user
+    if system_user is not None:
+        allowed_ids = scoped_calendar_ids(system_user, org)
+        if allowed_ids is not None and calendar_id not in allowed_ids:
+            raise Calendar.DoesNotExist("Calendar matching query does not exist.")
```
hugobessa, line 84: The single owner-guard choke point, now a shared helper so reads (Phase 1) and every provider write mutation (4a/4b/4c) enforce scope identically. A scoped token requesting a calendar outside its owner's set raises Calendar.DoesNotExist — the same not-found a genuinely missing calendar raises, so a write attempt against another provider never reveals the calendar exists.

public_api/mutations.py
```diff
@@ -543,3 +547,48 @@ def update_branding(
...
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def create_available_time(
+        self,
+        info: strawberry.Info,
+        calendar_id: int,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        timezone_str: Annotated[str, strawberry.argument(name="timezone")],
+        rrule_string: str | None = None,
+    ) -> AvailableTimeGraphQLType:
```
hugobessa, line 560: `rrule_string` is the one knob that covers BOTH availability shapes the spec calls for: omitted = a specific one-off availability date; an RFC-5545 RRULE = recurring availability. The Python parameter is `timezone_str` to avoid shadowing the module-level `django.utils.timezone`, but the GraphQL field stays `timezone` via a Strawberry argument alias so the public contract (and 4b/4c) are unaffected.

public_api/permissions.py
```diff
@@ -49,6 +49,7 @@ class OrganizationResourceAccess(BasePermission):
         "createScopedSystemUser": PublicAPIResources.SYSTEM_USER,
         "updateBranding": PublicAPIResources.BRANDING,
         "childOrganizations": PublicAPIResources.CHILD_ORG_ANALYTICS,
+        "createAvailableTime": PublicAPIResources.AVAILABLE_TIME,
```
hugobessa, line 52: A provider-scoped token legitimately holds AVAILABLE_TIME (it's in PROVIDER_SCOPED_RESOURCES), so this mapping lets it call the mutation — but the owner guard still confines every write to the token owner's own calendars. Resource grant + owner scope are independent layers.

hugobessa:

PR #109, opened 2026-06-18
feat(public_api): add owner-guarded createBlockedTime mutation

hugobessa:
## Summary
Phase 4b of the **Per-Owner-Scoped Public API Tokens** plan. Adds the second provider write mutation:
`createBlockedTime`. A scoped token can block time only on its owner's calendars; org-wide tokens are
unaffected. Structurally a mirror of Phase 4a's `createAvailableTime`.

## What changed
- `public_api/mutations.py` → `createBlockedTime`: `[IsAuthenticated, OrganizationResourceAccess]`
  guarded (mapped to `BLOCKED_TIME`). Args `calendar_id`, `start_time`, `end_time`, `timezone`,
  `reason=""`, optional `rrule_string`. Owner guard via the shared `prepare_service_and_calendar`
  (cross-owner → `Calendar.DoesNotExist` re-wrapped as an indistinguishable "does not exist" error).
  Wraps `CalendarService.create_blocked_time`; service `ValueError` → clean GraphQL error. Returns
  `BlockedTimeGraphQLType`.
- `public_api/permissions.py`: `createBlockedTime → BLOCKED_TIME` mapping.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 4b +
availability and schedule" (blocked-time write). Non-goals: only the blocked-time mutation here.

## Review history
Layer-3 review found no BLOCKERs. Added a **`reason` length guard** (a >255-char `reason` would raise
an uncaught `DataError` → 500; now a clean GraphQL error, with a test that no row is written),
mirrored Phase 4a's cross-owner-vs-missing **indistinguishability test**, asserted `reason` is
persisted, and replaced hardcoded test emails/ids with uuid-suffixed values. `create_blocked_time`
has no `manage_available_windows`-style guard (blocked times accept any calendar), so no analogous
"calendar can't accept" test was needed.

## Test plan
- `public_api/tests/test_mutations.py` — scoped token creates one-off (with `reason`, asserted
  persisted) and recurring (rrule) blocked times on its owned calendar; cross-owner `calendar_id` →
  not-found with message IDENTICAL to a genuinely-missing calendar AND no `BlockedTime` row written;
  org-wide token writes on any org calendar; reason >255 chars → clean error, no row; missing
  `BLOCKED_TIME` resource denied.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2073 passed**; mypy clean on touched modules.

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -14,7 +14,7 @@ import strawberry
 from dependency_injector.wiring import Provide, inject
 from graphql import GraphQLError
 
-from calendar_integration.graphql import AvailableTimeGraphQLType
+from calendar_integration.graphql import AvailableTimeGraphQLType, BlockedTimeGraphQLType
 from calendar_integration.models import Calendar
 from calendar_integration.mutations import CalendarGroupMutations
 from organizations.exceptions import UserAlreadyHasMembershipError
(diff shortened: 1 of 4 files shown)


public_api/mutations.py
```diff
@@ -592,3 +592,54 @@ def create_available_time(
...
+            The created BlockedTime as BlockedTimeGraphQLType.
+
+        Raises:
+            GraphQLError: Calendar not found / not in owner's scope; or service-level
+                ValueError (e.g. invalid rrule format).
+
+        The token's OrganizationResourceAccess must include the BLOCKED_TIME resource.
+        """
+        if len(reason) > 255:
+            raise GraphQLError("reason must be 255 characters or fewer.")
```
hugobessa, line 627: Blocked-time-specific guard added on review: `BlockedTime.reason` is CharField(max_length=255), so a longer reason would raise django DataError (NOT ValueError, so the service-call except below wouldn't catch it) and surface as a 500. This length check turns it into a clean GraphQL error before the row is attempted. `createAvailableTime` needs no equivalent — AvailableTime has no reason field.

public_api/mutations.py
```diff
@@ -592,3 +592,54 @@ def create_available_time(
...
+            blocked_time = calendar_service.create_blocked_time(
+                calendar=calendar,
+                start_time=start_time,
+                end_time=end_time,
+                timezone=timezone_str,
+                reason=reason,
+                rrule_string=rrule_string,
+            )
+        except ValueError as e:
+            raise GraphQLError(str(e)) from e
```
hugobessa, line 644: Same dual error mapping as 4a: cross-owner / missing calendar (Calendar.DoesNotExist from the shared owner guard) and service-level invalid input (ValueError, e.g. a malformed rrule) both become clean GraphQL errors. The cross-owner error message is asserted byte-identical to a genuinely-missing-calendar error so a scoped token can't probe other providers' calendars.

public_api/permissions.py
```diff
@@ -50,6 +50,7 @@ class OrganizationResourceAccess(BasePermission):
         "updateBranding": PublicAPIResources.BRANDING,
         "childOrganizations": PublicAPIResources.CHILD_ORG_ANALYTICS,
         "createAvailableTime": PublicAPIResources.AVAILABLE_TIME,
+        "createBlockedTime": PublicAPIResources.BLOCKED_TIME,
```
hugobessa, line 53: Maps the field to BLOCKED_TIME (in PROVIDER_SCOPED_RESOURCES), so a provider token can call it — but the owner guard still confines every write to the owner's own calendars. Resource grant and owner scope remain independent layers.

hugobessa:

PR #110, opened 2026-06-18
feat(public_api): add owner-guarded scheduleEvent mutation

hugobessa:
## Summary
Phase 4c of the **Per-Owner-Scoped Public API Tokens** plan — the last provider write mutation:
`scheduleEvent`. A provider-scoped token can schedule events on its OWNER'S calendars. This required
a sanctioned change to the calendar event service, which previously hard-blocked **all** Public API
(`SystemUser`) event creation.

## Why the service changed
`CalendarEventService.create_event` deliberately blocked `SystemUser` callers ("Events cannot be
created through the Public API") — event creation was routed only through single-use scheduling codes
or public-scheduling calendars. To let a provider schedule on their own calendar, this phase adds a
narrow, sanctioned allowance (approved as a scope expansion):
- A `SystemUser` may create events **only** when it is scoped to a user (`scoped_to_user`) who **owns
  the target calendar** (a `CalendarOwnership` row in the calendar's org). That bypasses
  `can_perform_scheduling` because the owner is inherently authorized for their own calendar.
- **Org-wide tokens (`scoped_to_user IS NULL`) remain blocked** for event creation — they still need
  single-use codes / public scheduling. (This is intentionally stricter than 4a/4b, where org-wide
  tokens can write availability/blocked times; events are higher-stakes.)
- Bundle calendars are explicitly rejected for provider tokens (a provider owns personal/resource
  calendars, not multi-provider bundles).

## What changed
- `calendar_integration/services/calendar_event_service.py`: owner-scoped `SystemUser` allowance in
  `create_event` (independent ownership verification — defense in depth); explicit BUNDLE rejection;
  fixed an `on_commit` side-effect that read the permission-service `token` unguarded (would
  `AttributeError` with no token) — now `getattr`-guarded and the audit **actor** falls back to the
  `SystemUser` so owner-scoped events aren't actor-less.
- `public_api/mutations.py` → `scheduleEvent`: `[IsAuthenticated, OrganizationResourceAccess]`
  guarded (mapped to `CALENDAR_EVENT`). Owner guard via the shared `prepare_service_and_calendar`;
  pre-validates that internal attendee `user_id`s are active members of the caller's org (prevents
  out-of-org attendees and a misleading DB `IntegrityError`); translates external attendees; title
  length guard; catches `ValueError` / `PermissionDenied` / `NoAvailableTimeWindowsError` → clean
  GraphQL errors. Returns `CalendarEventGraphQLType`.
- `public_api/permissions.py`: `scheduleEvent → CALENDAR_EVENT`.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 4c +
availability and schedule" (schedule-event write). Non-goals upheld: single-use scheduling codes are
NOT implemented; the service allowance is a deliberate, minimal alternative that keeps org-wide event
creation routed through the (separately planned) codes mechanism.

## Review history
First implementation used an unsound `initialize_without_provider(None)` workaround — it only worked
on `accepts_public_scheduling=True` calendars (not the provider default) and crashed a post-commit
side-effect; the issue was caught in review and escalated to a product decision (expand scope vs
defer). The reworked owner-scoped path was then reviewed and a **BLOCKER** (bundle-calendar recursion
producing a confusing failure) was fixed with an explicit bundle guard + test; the audit actor and
internal-attendee validation were hardened; test setups were made honest (a realistic
managed-availability-window success test; the indistinguishability test now uses a populated
owner-calendar set).

## Test plan
- `public_api/tests/test_mutations.py` — scoped owner schedules on a default (non-public) calendar with
  a covering availability window → success; recurring; attendees persisted; out-of-org attendee →
  clean error, no row; cross-owner calendar → not-found identical to a missing calendar + no row;
  **org-wide token → denied, no row**; bundle calendar → clean error, no row; no availability window →
  clean error; title-too-long → clean error; missing `CALENDAR_EVENT` resource → denied.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2085 passed**; `calendar_integration/`
  regression suite **1143 passed**; mypy clean on touched files.

### DIFF
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -44,6 +44,7 @@ from calendar_integration.models import (
     Calendar,
     CalendarEvent,
     CalendarManagementToken,
+    CalendarOwnership,
     EventAttendance,
     EventBulkModification,
     EventExternalAttendance,
@@ -237,6 +238,7 @@ class CalendarEventService:
 
         calendar = self._get_calendar_by_id(calendar_id)
 
+        owner_scoped_authorized = False
         if isinstance(context.user_or_token, User):
             context.calendar_permission_service.initialize_with_user(
                 context.user_or_token,
@@ -244,15 +246,39 @@ class CalendarEventService:
                 calendar_id=calendar_id,
             )
         elif isinstance(context.user_or_token, SystemUser):
-            raise PermissionDenied("Events cannot be created through the Public API.")
+            # A provider-scoped Public API token may schedule events ONLY on calendars
+            # owned by the user it is scoped to. Org-wide tokens (scoped_to_user is None)
+            # remain blocked — event creation otherwise requires single-use scheduling
+            # codes / public-scheduling calendars.
+            owner_id = context.user_or_token.scoped_to_user_id
+            if (
+                owner_id is not None
+                and CalendarOwnership.objects.filter_by_organization(calendar.organization_id)
+                .filter(calendar_fk_id=calendar_id, user_id=owner_id)
+                .exists()
+            ):
+                owner_scoped_authorized = True
+            else:
+                raise PermissionDenied("Events cannot be created through the Public API.")
 
-        if not context.calendar_permission_service.can_perform_scheduling(
-            calendar_id=calendar_id,
-            calendar_settings=CalendarSettingsData(
-                manage_available_windows=calendar.manage_available_windows,
-                accepts_public_scheduling=calendar.accepts_public_scheduling,
-            ),
-            event=event_data,
+        # Owner-scoped tokens are limited to PERSONAL/RESOURCE calendars they own.
+        # Allowing a bundle calendar would fan-out to child calendars owned by other
+        # providers and fail confusingly; block it with an explicit, clean error.
+        if owner_scoped_authorized and calendar.calendar_type == CalendarType.BUNDLE:
+            raise PermissionDenied(
+                "Provider-scoped tokens cannot schedule events on bundle calendars."
+            )
+
+        if (
+            not owner_scoped_authorized
+            and not context.calendar_permission_service.can_perform_scheduling(
+                calendar_id=calendar_id,
+                calendar_settings=CalendarSettingsData(
+                    manage_available_windows=calendar.manage_available_windows,
+                    accepts_public_scheduling=calendar.accepts_public_scheduling,
+                ),
+                event=event_data,
+            )
         ):
             raise PermissionDenied("You do not have permission to update this event.")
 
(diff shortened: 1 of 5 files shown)


calendar_integration/services/calendar_event_service.py
```diff
@@ -237,22 +238,47 @@ def create_event(self, calendar_id: int, event_data: CalendarEventInputData) ->
...
+            owner_id = context.user_or_token.scoped_to_user_id
+            if (
+                owner_id is not None
+                and CalendarOwnership.objects.filter_by_organization(calendar.organization_id)
+                .filter(calendar_fk_id=calendar_id, user_id=owner_id)
+                .exists()
+            ):
+                owner_scoped_authorized = True
+            else:
+                raise PermissionDenied("Events cannot be created through the Public API.")
```
hugobessa, line 262: The sanctioned relaxation of the "Events cannot be created through the Public API" block. A SystemUser is authorized ONLY when it is scoped to a user who owns this calendar (independent CalendarOwnership check in the calendar's org — defense in depth, not trusting the caller). Org-wide tokens (scoped_to_user_id is None) fall to the else and stay blocked: event creation for them still requires single-use codes / public scheduling. Cross-org is impossible because the ownership row is filtered by the calendar's organization.

calendar_integration/services/calendar_event_service.py
```diff
@@ -237,22 +238,47 @@ def create_event(self, calendar_id: int, event_data: CalendarEventInputData) ->
...
+
+        # Owner-scoped tokens are limited to PERSONAL/RESOURCE calendars they own.
+        # Allowing a bundle calendar would fan-out to child calendars owned by other
+        # providers and fail confusingly; block it with an explicit, clean error.
+        if owner_scoped_authorized and calendar.calendar_type == CalendarType.BUNDLE:
+            raise PermissionDenied(
+                "Provider-scoped tokens cannot schedule events on bundle calendars."
+            )
```
hugobessa, line 270: Explicit bundle rejection (review BLOCKER fix). Without it, an owner-scoped token that owns a BUNDLE calendar would pass the gate, then recurse into per-child create_event calls that fail on other-provider child calendars — confusing, though atomic-safe. Provider tokens schedule on their own personal/resource calendars; bundles are out of scope for them.

public_api/mutations.py
```diff
@@ -643,3 +659,93 @@ def create_blocked_time(
...
+                    id__in=requested_ids,
+                    organization_memberships__organization=org,
+                    organization_memberships__is_active=True,
+                ).values_list("id", flat=True)
+            )
+            if set(requested_ids) != found_ids:
+                raise GraphQLError(
+                    "One or more attendee user_ids are not active members of this organization."
+                )
+
```
hugobessa, line 726: Internal attendees are pre-validated as active members of the caller's org before the event is built. This closes two gaps at once: a nonexistent user_id would otherwise reach the DB as an opaque IntegrityError (now a clean validation error), and a provider could otherwise attach an arbitrary out-of-org user as an attendee. With this in place the mutation no longer needs to catch IntegrityError broadly.

hugobessa:

PR #111, opened 2026-06-18
feat(public_api): adversarial owner-scope sweep + security review

hugobessa:
## Summary
Phase 5 (final) of the **Per-Owner-Scoped Public API Tokens** plan: prove the per-owner scope cannot
be bypassed across the ENTIRE reachable surface, close the leaks found, and ship the security sign-off
checklist. It found and fixed a **real cross-owner read leak** — the surface deferred from Phase 1.

## The leak (now closed)
Top-level reads were owner-filtered in Phase 1, but a provider-scoped token fetching its OWN event
could traverse **nested GraphQL fields** to other providers' data. Strawberry permission classes run
only on the decorated top-level field, so nested traversal bypassed the resource gate. Closed via
owner-scoped `@strawberry.field` resolvers on the shared types in `calendar_integration/graphql.py`,
driven by a helper `_owner_scoped_calendar_ids(info)` that returns `None` for org-wide / internal
(non-public-API) requests — so those paths stay byte-for-byte — or the owner's calendar-id set for a
scoped token.

Guarded nested fields (each filtered to the owner's calendars / suppressed for scoped tokens, unchanged
for org-wide):
- `CalendarEventGraphQLType`: `calendar`, `bundleCalendar`, `bundlePrimaryEvent`,
  `bulkModificationParent`, `parentRecurringObject`, `bundleRepresentations`, `bulkModifications`,
  `recurringInstances`, `resources`, `resourceAllocations`, `groupSelections`, `calendarGroup` (suppressed).
- **Second hop** (found in review): `groupSelections.slot` (suppressed) + `CalendarGroupSlot.calendars`
  (pool filtered) — this defeated the `calendarGroup` suppression via a sibling path; the three
  recurrence-exception types' `parent_*`/`modified_*` pointers; `EventExternalAttendance.event`.
- Proven safe (documented): scalars, `attendees`/attendances (people), `resourceAllocations.calendar` &
  `groupSelections.calendar` (parent list already owner-filtered), `CalendarGroup.slots` (both entry
  points suppressed → unreachable by a scoped token).

## What else
- `public_api/queries.py`: cleaned 3 mypy `no-redef` warnings (duplicate `request:` annotations from Phase 1).
- `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_SECURITY_REVIEW.md`: the bypass-surface
  checklist — every reachable field, its guard mechanism, and the proving test. Sign-off artifact.

## Plan reference
Plan `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_IMPLEMENTATION_PLAN.md` — Phase 5 + Goal 5.
Spec Decisions → Use-cases: "Provider token attempts to reach another provider's data" (the consolidated
negative-path guarantee).

## Review history
The first sweep missed a second-hop leak: `groupSelections.slot.calendars` exposed the entire
cross-provider slot pool, bypassing the `calendarGroup` suppression. Adversarial review (Opus) caught it
plus unguarded recurrence-exception back-pointers; both were closed with dual guards (suppress `slot` +
filter the pool) and `_scoped_event/blocked/available_or_none` resolvers, with adversarial tests that
fail if any guard is reverted. An independent completeness pass confirmed no other raw nested field
reachable by a scoped token remains unguarded.

## Test plan
- `public_api/tests/test_scoping_security.py` (23 tests) — for every reachable read field AND each
  provider write mutation: no cross-owner top-level leak, **no cross-owner leak via nested-field
  selection** (the queries explicitly select the nested fields), cross-owner writes denied with no row,
  and org-wide regression assertions per guarded field. Behavioral — each fails if its guard is removed.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2108 passed**; `calendar_integration/`
  regression **1143 passed**; mypy clean on touched files (`no-redef` gone).


calendar_integration/graphql.py
```diff
@@ -25,6 +26,108 @@
...
+    # Imported lazily to avoid a public_api -> calendar_integration -> public_api
+    # import cycle at module load.
+    from public_api.scoping import scoped_calendar_ids
+
+    request = info.context.request
+    system_user = getattr(request, "public_api_system_user", None)
+    organization = getattr(request, "public_api_organization", None)
+    if system_user is None or organization is None:
+        return None
+    return scoped_calendar_ids(system_user, organization)
```
hugobessa, line 52: The per-field-resolver counterpart of the top-level read enforcement. Returns None for org-wide tokens AND for any internal (non-public-API) GraphQL request (the request has no public_api_* attributes), so every nested resolver is a strict no-op outside provider-scoped requests — that's what keeps org-wide + internal consumers of these shared types byte-for-byte. Lazy import avoids a public_api <-> calendar_integration cycle.

calendar_integration/graphql.py
```diff
@@ -351,9 +687,22 @@ class CalendarEventGroupSelectionGraphQLType:
...
 
+    @strawberry.field
+    def slot(self, info: strawberry.Info) -> "CalendarGroupSlotGraphQLType | None":
+        """The CalendarGroupSlot this selection corresponds to.
+
+        Suppressed entirely for scoped tokens: a slot's calendar pool spans ALL
+        providers in the group, so exposing it would enumerate other-owner calendars
+        even after the per-slot ``calendars`` filter. Org-wide tokens see the slot
+        unchanged (mirror of the ``calendarGroup`` suppression at ~L340).
+        """
```
hugobessa, line 700: The review BLOCKER fix. groupSelections is owner-filtered, but each selection's `slot` exposed CalendarGroupSlot.calendars — the slot's ENTIRE candidate pool across providers — defeating the calendarGroup suppression via a sibling path. Nested traversal doesn't re-run the top-level CALENDAR_GROUP permission gate, so a token holding only CALENDAR_EVENT reached it. `slot` is now suppressed for scoped tokens; CalendarGroupSlot.calendars is also pool-filtered as defense in depth.

calendar_integration/graphql.py
```diff
@@ -131,37 +269,157 @@ class CalendarEventGraphQLType:
...
+    @strawberry.field
+    def calendar_group(self, info: strawberry.Info) -> "CalendarGroupGraphQLType | None":
+        """The CalendarGroup this event was booked through.
+
+        A CalendarGroup aggregates calendars across providers via its slots, so for a
+        scoped token we suppress it entirely (group membership is not owner data and the
+        group's slots would expose other-owner calendars). Org-wide tokens see it.
+        """
+        if _owner_scoped_calendar_ids(info) is not None:
+            return None
```
hugobessa, line 370: calendarGroup is suppressed entirely for scoped tokens — a group aggregates calendars across providers via its slots, so it is not owner data. Org-wide tokens still see it. Combined with the slot suppression above, CalendarGroupGraphQLType is unreachable by a scoped token (its only nested entry points are both closed), so its own raw `slots` field needs no separate guard.

calendar_integration/graphql.py
```diff
@@ -113,8 +228,31 @@ class EventRecurrenceExceptionGraphQLType:
...
+        """The parent recurring event this exception applies to.
+
+        Guarded so a scoped token cannot reach an event whose calendar is outside the
+        owner's set via this second-hop pointer.
+        """
+        exc = cast(EventRecurrenceException, self)
+        return cast(
+            "CalendarEventGraphQLType | None",
+            _scoped_event_or_none(exc.parent_event, info),
+        )
```
hugobessa, line 242: Second-hop guard: a recurrence exception's parent/modified event pointer returns CalendarEventGraphQLType, whose SCALARS would otherwise bypass the field resolvers. Routed through _scoped_event_or_none so a scoped token can't read an event on a calendar outside its owner set even via this pointer. Same pattern applied to the blocked-time and available-time recurrence exceptions and the external-attendance event back-pointer.

hugobessa:

PR #113, opened 2026-06-18
feat(public-api): code-gated availability reads (Phase 4)

hugobessa:
## Summary

First **unauthenticated** surface of the feature: a patient (no org token) reads availability for
a code's bound scope. Adds five query fields with NO `permission_classes` (and deliberately NOT in
`FIELD_TO_RESOURCE_MAPPING`): `availableTimesWithCode`, `availabilityWindowsWithCode`,
`unavailableWindowsWithCode`, `calendarGroupBookableSlotsWithCode`, `calendarGroupAvailabilityWithCode`.

New service primitive `CalendarPermissionService.resolve_code(code)` derives the org **from the
code** — it looks the token up by id alone (unscoped `original_manager`) and the constant-time
hash verify is the gate; org is then read off the token. `validate_code(code, org_id)` now delegates
to `resolve_code` and asserts the org, sharing one decode/verify path.

Security properties (reviewed):
- Hash verify runs before any token data is used; unknown-id and bad-hash both → identical
  `GraphQLError("Invalid or expired code.")` (no enumeration oracle).
- Scope is taken STRICTLY from the token (`token.calendar`/`token.event.calendar` for calendar
  fields; `token.calendar_group`/`token.event.calendar_group` for group fields) — no client-supplied
  calendar/group id can override it. Calendar codes can't read group data and vice versa.
- Cross-org isolation: org comes from the token; availability services run every query through
  `filter_by_organization`. (Now covered by a real non-mocked cross-org test.)
- Reads NEVER consume the code (`used_at` stays null — asserted).
- Datetime range is clamped to 366 days (DoS guard on the unauthenticated surface); rate limiter
  keys unauthenticated requests as `anon:<client_ip>`.

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 4.
- Spec: Decisions → Use-cases, Use-cases 2 & 4 (the availability-read steps). Open Question 1 (read set).

## Test plan

In-container (docker postgres): `public_api/tests/test_code_gated_reads.py` — 30 tests: per-field happy
paths (no auth header), event-bound code via `event.calendar`, non-consumption (`used_at` null after
read), wrong-scope (calendar code on group field & vice versa) → uniform error, invalid/unknown/
revoked/expired → uniform error, **tampered-secret (valid id, wrong secret)** → uniform error,
**real non-mocked cross-org isolation**, and **over-max / backwards range** rejection. `ruff` clean ·
`makemigrations --check` no changes · `check --deploy` clean · `pytest -n auto` **2102 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-3`.

### DIFF
--- a/calendar_integration/services/calendar_permission_service.py
+++ b/calendar_integration/services/calendar_permission_service.py
@@ -576,27 +576,29 @@ class CalendarPermissionService:
 
         return token, plaintext_code
 
-    def validate_code(self, code: str, organization_id: int) -> CalendarManagementToken:
-        """Decode and validate a booking code, returning the active token.
+    def resolve_code(self, code: str) -> CalendarManagementToken:
+        """Decode and validate a booking code WITHOUT a known organization.
 
-        Performs the same decode/hash/lookup as ``initialize_with_token`` but
-        additionally enforces that the token is still active (not used, not
-        revoked, not expired).
+        This method is for unauthenticated reads where the org is derived FROM
+        the code itself.  It performs the same decode/hash/verify logic as
+        ``validate_code`` but looks up the token by id alone (no org filter).
+        The org is safe to derive from the returned token's ``organization_id``
+        because the secret-hash verification gates access to the token data.
 
         Args:
             code: The base64-encoded plaintext code as returned by
                 ``create_booking_token``.
-            organization_id: Tenant scope.  The token must belong to this org.
 
         Returns:
             The active ``CalendarManagementToken`` instance, with
-            ``permissions`` and ``calendar``/``event`` pre-fetched.
+            ``permissions``, ``calendar``, ``calendar_group``, and ``event``
+            pre-fetched.
 
         Raises:
             InvalidTokenError: If the code is malformed or does not match any token.
-            TokenExpiredError: If the token's ``expires_at`` has passed.
-            TokenAlreadyUsedError: If the token was already consumed.
             TokenRevokedError: If the token was revoked.
+            TokenAlreadyUsedError: If the token was already consumed.
+            TokenExpiredError: If the token's ``expires_at`` has passed.
         """
         try:
             token_full_str = base64.b64decode(code).decode("utf-8")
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_permission_service.py
```diff
@@ -633,6 +647,36 @@ def validate_code(self, code: str, organization_id: int) -> CalendarManagementTo
 
         return token
```
hugobessa, line 648: `resolve_code` is the org-from-code primitive every unauthenticated phase (4–6) reuses. It looks the token up by id via `original_manager` (bypassing the tenant-required guard) — safe ONLY because the integer id is useless without the secret, and the constant-time `verify_long_lived_token` immediately below is the real gate. Token data is never returned/acted on before that verify. Lifecycle precedence (revoked > used > expired) matches Phase 0. `validate_code` now delegates here and asserts the org, so authenticated and unauthenticated paths share one decode/verify.

public_api/queries.py
```diff
@@ -615,6 +701,205 @@ def child_organizations(
...
+        if calendar is None and token.event is not None:
+            calendar = token.event.calendar
+        if calendar is None:
+            raise GraphQLError(_CODE_GATED_ERROR_MESSAGE)
+
+        org = _get_org_from_token(token)
+        calendar_service = _prepare_service_and_calendar_for_org(deps, org, calendar)
+
+        available_times = calendar_service.get_available_times_expanded(
+            calendar,
```
hugobessa, line 735: Representative code-gated read. No `permission_classes` (unauthenticated, à la `check_token`). Scope is taken STRICTLY from the token (`token.calendar` or `token.event.calendar`) — there is no client-supplied calendar_id arg that could redirect the read to another calendar. A group-only code yields `calendar is None` → the uniform error. The code is never consumed. Org comes from the token and the availability service filters by it, so cross-org data can't surface.

public_api/queries.py
```diff
@@ -118,6 +139,71 @@ def _prepare_service_and_calendar(
...
+    Raises ``GraphQLError`` if the range is backwards or exceeds
+    ``MAX_CODE_GATED_RANGE``.  Called BEFORE any expensive service call.
+    """
+    if end <= start:
+        raise GraphQLError("Invalid time range.")
+    if (end - start) > MAX_CODE_GATED_RANGE:
+        raise GraphQLError("Requested time range is too large.")
+
+
 @strawberry.input
```
hugobessa, line 207: DoS guard for the unauthenticated surface: `_validate_code_gated_range` rejects backwards ranges and ranges over `MAX_CODE_GATED_RANGE` (366 days) BEFORE any recurrence-expansion service call, so a multi-year window can't be used to amplify load. These messages describe the range (not code existence), so they need not match the uniform code-error message.

PR #112, opened 2026-06-18
Phase 3a — createAvailabilityWindow Public GraphQL mutation

hugobessa:
## Summary
Adds the `createAvailabilityWindow` mutation so integrators can create a single (optionally recurring) available time on a calendar. Thin wrapper over `CalendarService.create_available_time(calendar, start_time, end_time, timezone, rrule_string)`.

- `public_api/mutations.py`: `createAvailabilityWindow` mutation, `{success, errorMessage, availableTime}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`. Resolves an org-scoped Calendar, then delegates.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.CREATE_AVAILABILITY_WINDOW` + mapping.
- `public_api/migrations/0010_alter_resourceaccess_resource_name.py`: choices-only AlterField.

## Plan reference
Plan → Phase 3a. No feature flag. Stacked on Phase 2c (PR #103). First phase of the availability/blocked-time area.

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2029 passed**. Tests (`TestCreateAvailabilityWindowMutation`): happy path (DB row + payload startTime/endTime), recurring (`rruleString` → non-null recurrence_rule), non-managing calendar → success=False, cross-org → success=False, permission-denied, unauthenticated.

PR #114, opened 2026-06-18
Phase 3b — updateAvailabilityWindow Public GraphQL mutation

hugobessa:
## Summary
Adds `updateAvailabilityWindow`, implemented via a single-op `batch_modify_available_times(calendar, [{"action": "update", "id": ..., <partial fields>}])` — the project's atomic available-time mutation path. Partial-update semantics: only `start_time`/`end_time`/`timezone`/`rrule_string` keys present when supplied.

- `public_api/mutations.py`: `updateAvailabilityWindow` mutation, `{success, errorMessage, availableTime}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.UPDATE_AVAILABILITY_WINDOW` + mapping.
- Migration `0011_alter_resourceaccess_resource_name.py` (choices-only).

## Plan reference
Plan → Phase 3b. No feature flag. Stacked on Phase 3a (PR #112).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2036 passed**. Tests: happy path (payload startTime/endTime reflect update), rrule update (recurrence_rule set), non-managing calendar → success=False, missing/cross-org id → success=False, permission-denied, unauthenticated.


public_api/mutations.py
```diff
@@ -714,3 +736,66 @@ def create_availability_window(
...
+        ) as e:
+            return UpdateAvailabilityWindowResult(success=False, error_message=str(e))
+
+        # Find the updated row in the returned list.
+        updated_time = next((at for at in updated_times if at.id == input.available_time_id), None)
+        if updated_time is None:
+            return UpdateAvailabilityWindowResult(
+                success=False,
+                error_message="Updated available time not found in result set.",
+            )
```
hugobessa, line 797: Update routes through the service's single atomic mutation API (`batch_modify_available_times`) with one `action: "update"` op — note the key is `action`, not `op` (the v2 doc was wrong). The service returns the calendar's full available-time list, so the updated row is found by id; if it's somehow absent the resolver returns success=False rather than a silent `success=True, availableTime=null`. Missing/cross-org ids raise `ValueError` inside the service (scoped by org + calendar) → success=False.

PR #117, opened 2026-06-18
Phase 3d — batchUpdateAvailabilityWindows Public GraphQL mutation

hugobessa:
## Summary
Exposes the full atomic `batch_modify_available_times(calendar, operations)` as `batchUpdateAvailabilityWindows`. The resolver validates each op's `action ∈ {create, update, delete}` and (for `create`) requires `startTime`/`endTime`/`timezone` up front, maps `availableTimeId → id`, includes only provided fields, and returns the post-batch `availableTimes` list.

- `public_api/mutations.py`: `batchUpdateAvailabilityWindows` mutation + `BatchAvailabilityOperationInput` / `BatchAvailabilityInput` / result types, guarded by `IsAuthenticated` + `OrganizationResourceAccess`.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.BATCH_UPDATE_AVAILABILITY_WINDOWS` + mapping.
- Migration `0013_alter_resourceaccess_resource_name.py` (choices-only).

## Plan reference
Plan → Phase 3d. No feature flag. Stacked on Phase 3c (PR #115).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2050 passed**. Tests: mixed create+update+delete batch applied (DB asserted); atomic rollback (bad id → no partial writes, rows unchanged); invalid action → success=False; create op missing fields → success=False (no 500, no write); cross-org calendar rejected; permission-denied; unauthenticated.


public_api/mutations.py
```diff
@@ -858,3 +899,90 @@ def delete_availability_window(
...
+                )
+            if op_input.action == "create" and (
+                op_input.start_time is None
+                or op_input.end_time is None
+                or op_input.timezone is None
+            ):
+                return BatchUpdateAvailabilityWindowsResult(
+                    success=False,
+                    error_message="create operation requires startTime, endTime, and timezone",
+                    available_times=[],
```
hugobessa, line 952: Two fail-fast guards run BEFORE the service call: (1) every op's `action` must be in {create, update, delete}; (2) a `create` op must carry startTime/endTime/timezone — the service indexes `operation["start_time"]` for creates, so a missing key would otherwise be an uncaught KeyError → 500. Both return success=False without touching the DB.

public_api/mutations.py
```diff
@@ -858,3 +899,90 @@ def delete_availability_window(
...
+
+        calendar_service, org = _get_org_and_init_calendar_service(info)
+
+        try:
+            calendar = Calendar.objects.filter_by_organization(org.id).get(id=input.calendar_id)
+        except Calendar.DoesNotExist:
+            return BatchUpdateAvailabilityWindowsResult(
+                success=False, error_message="Calendar not found.", available_times=[]
+            )
+
```
hugobessa, line 935: The whole batch is atomic: `batch_modify_available_times` is decorated with `@transaction.atomic()` (availability_service.py), so any failing op (e.g. a bad id raising ValueError) rolls back every op in the batch — including creates that ran earlier — regardless of the resolver catching the exception. The rollback test asserts row counts are unchanged after a batch whose later op fails.

PR #121, opened 2026-06-18
feat(calendar): reschedule a single-calendar event with a code (Phase 6a)

hugobessa:
## Summary

Unauthenticated `rescheduleCalendarEventWithCode` (no `permission_classes`). An event-bound RESCHEDULE
code moves exactly its bound event to a new time. Flow: `resolve_code` → require RESCHEDULE + event
scope + single-calendar (`token.calendar_group is None`; group reschedule codes route to Phase 6b) →
availability pre-check → `transaction.atomic { initialize_without_provider(user_or_token=code) →
update_event → consume_code }`. `calendar_id`/`event_id` come STRICTLY from the token's bound event
(a code for event E can only reschedule E). First-write-wins; out-of-window → SLOT_UNAVAILABLE, code
not consumed, retryable; replay → ALREADY_USED.

## Two notable correctness points

1. **Details-preservation (so only RESCHEDULE is required).** `can_perform_update` derives the required
   permissions from the diff between old and new event — a title/description/attendee change would also
   require UPDATE_DETAILS/UPDATE_ATTENDEES, which a reschedule code lacks. So the mutation rebuilds
   `CalendarEventInputData` COPYING the existing event's title/description/attendances/external_attendances
   (by email)/resource allocations and overrides ONLY start/end/timezone → the diff is exactly `{RESCHEDULE}`.

2. **Pre-existing `update_event` bug fixed (shared, kept).** `start_time`/`end_time` on `CalendarEvent`
   are read-only `GeneratedField`s (db-derived from `start_time_tz_unaware` + `timezone`), so the old
   `event.start_time = …` was silently dropped — `update_event` never actually changed times for ANY
   caller. Fixed to write `*_tz_unaware` + `timezone` (mirrors `create_event`). A reviewer-flagged
   availability check that was initially added to shared `update_event` (would have regressed authenticated
   REST + bundle updates) was MOVED into this mutation, scoped to the code path.

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 6a.
- Spec: Decisions → Use-cases, Use-case 4 (patient reschedules — single-calendar).

## Test plan

In-container (docker postgres): `public_api/tests/test_reschedule_with_code.py` on a RESTRICTED calendar —
happy path (times updated, details preserved, code consumed), replay → ALREADY_USED, out-of-window →
SLOT_UNAVAILABLE + not consumed + retryable, missing RESCHEDULE → NOT_PERMITTED, group reschedule code →
NOT_PERMITTED, expired/revoked/invalid, and the rescheduled event is exactly `token.event`. `ruff` clean ·
`makemigrations --check` no changes · `check --deploy` clean · `pytest -n auto` **2146 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-5b`.

### DIFF
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -43,6 +43,7 @@ from calendar_integration.services.dataclasses import (
     EventAttendanceInputData,
     EventExternalAttendanceInputData,
     ExternalAttendeeInputData,
+    ResourceAllocationInputData,
 )
 from calendar_integration.services.webhook_analytics_service import WebhookAnalyticsService
 from organizations.models import Organization
@@ -576,6 +577,16 @@ class CreateGroupEventWithCodeInput:
     slot_selections: list[CodeSlotSelectionInput]
     external_attendee: ExternalAttendeeCodeInput
     description: str = ""
+
+
+@strawberry.input
+class RescheduleWithCodeInput:
+    """Input for the unauthenticated rescheduleCalendarEventWithCode mutation."""
+
+    code: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    timezone: str
 
 
 @strawberry.type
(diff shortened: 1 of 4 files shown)


calendar_integration/mutations.py
```diff
@@ -1362,3 +1373,226 @@ def create_calendar_group_event_with_code(
...
+            title=existing_event.title,
+            description=existing_event.description or "",
+            start_time=input.start_time,
+            end_time=input.end_time,
+            timezone=input.timezone,
+            attendances=preserved_attendances,
+            external_attendances=preserved_external_attendances,
+            resource_allocations=preserved_resource_allocations,
+        )
+
```
hugobessa, line 1535: Details-preservation is load-bearing: `can_perform_update` computes required permissions from the old↔new diff, so we COPY title/description/attendances/external_attendances(by email)/resource allocations from the existing event and override ONLY start/end/timezone. That makes the diff exactly {RESCHEDULE} — which is all the code holds. Drop or reorder any preserved field and the update would additionally require UPDATE_DETAILS/UPDATE_ATTENDEES → PermissionDenied. The restricted-calendar happy-path test would fail if preservation broke.

calendar_integration/services/calendar_event_service.py
```diff
@@ -534,8 +534,18 @@ def update_event(
...
+        # with ``db_persist=True``) that derive from ``start_time_tz_unaware`` and
+        # the IANA ``timezone``.  Assigning to the generated fields is silently
+        # ignored by Django's UPDATE statement, so we must update the underlying
+        # writable fields instead.
+        event.start_time_tz_unaware = self.convert_naive_utc_datetime_to_timezone(
+            event_data.start_time, event_data.timezone
+        )
+        event.end_time_tz_unaware = self.convert_naive_utc_datetime_to_timezone(
+            event_data.end_time, event_data.timezone
+        )
```
hugobessa, line 547: Genuine pre-existing bug fix (shared, kept). `start_time`/`end_time` are read-only `GeneratedField`s derived from `start_time_tz_unaware` + `timezone`; the old `event.start_time = …` was silently ignored by the UPDATE, so `update_event` never changed times for any caller. Now writes the underlying `*_tz_unaware` + `timezone` via the same `convert_naive_utc_datetime_to_timezone` helper `create_event` uses. The availability check that was briefly added here was moved into the reschedule mutation to avoid regressing REST/bundle updates.

calendar_integration/mutations.py
```diff
@@ -1362,3 +1373,226 @@ def create_calendar_group_event_with_code(
...
+        # the row lock and the whole transaction (including the just-updated event) rolls
+        # back, leaving exactly one update and the code consumed once.
+        try:
+            with transaction.atomic():
+                deps.calendar_service.initialize_without_provider(
+                    user_or_token=input.code, organization=org
+                )
+                event = deps.calendar_service.update_event(calendar_id, event_id, event_data)
+                deps.calendar_permission_service.consume_code(token, source_ip)
+        except (TokenAlreadyUsedError, TokenExpiredError, TokenRevokedError) as e:
```
hugobessa, line 1566: Availability enforcement scoped to the code path (not shared update_event): an out-of-window new slot returns SLOT_UNAVAILABLE BEFORE the transaction, so the code is never consumed (retryable). The happy path then runs the atomic update→consume; update before consume so a concurrent loser's consume raises under the row lock and the whole txn rolls back (single-use, first-write-wins). calendar_id/event_id are from the token's bound event only — no client override.

PR #131, opened 2026-06-18
feat(calendar): reschedule a calendar-group event with a code (Phase 6b)

hugobessa:
## Summary

Unauthenticated `rescheduleCalendarGroupEventWithCode` (no `permission_classes`). A group event-bound
RESCHEDULE code moves its grouped event to a new time, **time-only** (slot re-selection deferred —
Open Question 3), preserving the event id (the Building Blocks integration stores it).

New `CalendarGroupService.reschedule_grouped_event(event_id, start, end, tz)`:
- Updates the PRIMARY event's time via `update_event`, rebuilding `CalendarEventInputData` with all
  existing details preserved (title/description/attendances/external_attendances-by-email/resources),
  overriding only times → `can_perform_update` requires exactly `{RESCHEDULE}` (same technique as 6a).
- Updates the linked non-primary `BlockedTime`s (found by `external_id` LIKE `group-event-{id}-cal-`,
  org-scoped) to the new times.
- In-place: event id + `calendar_group_fk` preserved.

Mutation: `resolve_code` → require RESCHEDULE + event scope + GROUP scope (calendar-only codes →
NOT_PERMITTED, route to 6a) → availability pre-check on the primary calendar (scoped to the code path)
→ `transaction.atomic { calendar_service.initialize_without_provider(user_or_token=code) →
group_service.calendar_service = that instance → group_service.initialize → reschedule_grouped_event →
consume_code }`. event_id strictly from the token. First-write-wins; out-of-window → SLOT_UNAVAILABLE
(not consumed, retryable); replay → ALREADY_USED.

## Timezone-storage BLOCKER fixed (create + reschedule now agree)

The model stores `*_tz_unaware` as the naive LOCAL wall-clock paired with `timezone`; the primary event
is stored converted. But `_create_non_primary_blocked_times` was writing blocked-time `*_tz_unaware`
RAW (UTC wall-clock), so for non-UTC timezones the non-primary busy markers drifted off the primary
event (a pre-existing create bug, invisible to UTC-only tests). Fixed create to convert via
`convert_naive_utc_datetime_to_timezone`, matching the primary event and the reschedule path. Added a
non-UTC (`America/Recife`) regression test asserting the primary event and its blocked times resolve to
the same instants after reschedule (fails before the fix by the UTC offset).

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 6b; Open Question 3.
- Spec: Decisions → Use-cases, Use-case 4 (patient reschedules — group variant).

## Test plan

In-container (docker postgres): `public_api/tests/test_reschedule_group_with_code.py` — happy path on a
RESTRICTED primary calendar (times updated, linked blocked times moved, event id unchanged, details
preserved, code consumed), the non-UTC alignment test, replay → ALREADY_USED, out-of-window →
SLOT_UNAVAILABLE + not consumed, calendar-only code → NOT_PERMITTED, missing RESCHEDULE → NOT_PERMITTED,
expired/revoked/invalid. `ruff` clean · `makemigrations --check` no changes · `check --deploy` clean ·
`pytest -n auto` **2161 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-6a`.

### DIFF
--- a/calendar_integration/mutations.py
+++ b/calendar_integration/mutations.py
@@ -589,6 +589,21 @@ class RescheduleWithCodeInput:
     timezone: str
 
 
+@strawberry.input
+class RescheduleGroupWithCodeInput:
+    """Input for the unauthenticated rescheduleCalendarGroupEventWithCode mutation.
+
+    Slot selections are NOT included: v1 keeps existing group/calendar selections
+    and changes ONLY the event times.  Full slot re-selection is deferred to a
+    future version (see Open Question 3 in the implementation plan).
+    """
+
+    code: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    timezone: str
+
+
 @strawberry.type
 class CalendarGroupMutations:
     """GraphQL mutations for CalendarGroup CRUD and grouped event booking."""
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_group_service.py
```diff
@@ -480,6 +486,134 @@ def create_grouped_event(self, data: CalendarGroupEventInputData) -> CalendarEve
...
+        the Building Blocks integration) store it and rely on it remaining stable
+        across reschedules.
+
+        v1 limitation: non-primary calendar availability is NOT re-checked on
+        reschedule. Only the primary calendar is gated (in the mutation's
+        availability check). Non-primary double-booking is therefore possible and
+        intentionally unenforced for this time-only reschedule path.
+        """
+        self._assert_initialized()
+        if self.calendar_service is None:
```
hugobessa, line 520: Time-only group reschedule (slot re-selection deferred — Open Question 3). Updates the PRIMARY event in place via update_event with all details preserved (so only {RESCHEDULE} is required) and moves the linked non-primary BlockedTimes — the event id is preserved on purpose (the Building Blocks integration stores it). v1 limitation noted in the docstring: non-primary calendar availability is NOT re-checked here (only the primary is gated in the mutation), so non-primary double-booking is possible and intentionally unenforced for this path.

calendar_integration/services/calendar_group_service.py
```diff
@@ -480,6 +486,134 @@ def create_grouped_event(self, data: CalendarGroupEventInputData) -> CalendarEve
...
+            external_id__startswith=f"group-event-{event_id}-cal-"
+        )
+
+        blocked_times = list(blocked_times_qs)
+        for bt in blocked_times:
+            bt.start_time_tz_unaware = new_start_tz_unaware
+            bt.end_time_tz_unaware = new_end_tz_unaware
+            bt.timezone = tz
+
+        if blocked_times:
```
hugobessa, line 610: Linked blocked times are found by `external_id__startswith="group-event-{id}-cal-"` (org-scoped) — the trailing `-cal-` prevents prefix collisions (event 1 vs 12). They're written with the CONVERTED local wall-clock, matching the primary event. The create path (`_create_non_primary_blocked_times`, L782) was fixed in this phase to use the same conversion — previously it wrote raw UTC, drifting the busy markers off the primary event for non-UTC zones. A non-UTC regression test guards this.

calendar_integration/mutations.py
```diff
@@ -1596,3 +1611,185 @@ def reschedule_calendar_event_with_code(
...
+            return CodeEventResult(
+                success=False,
+                error_code=BookingCodeErrorCode.INVALID_CODE,
+                error_message="Invalid or unknown booking code.",
+            )
+        except TokenExpiredError:
+            return CodeEventResult(
+                success=False,
+                error_code=BookingCodeErrorCode.EXPIRED,
+                error_message="This booking code has expired.",
```
hugobessa, line 1650: Requires RESCHEDULE + event scope + GROUP scope (`token.calendar_group` set); calendar-only reschedule codes route to Phase 6a. event_id is strictly `token.event_fk_id` (no client override). The atomic block wires `group_service.calendar_service` to the code-bearing CalendarService (the 5b technique) so the primary update authorizes via the token on a restricted calendar; reschedule_grouped_event + consume_code run in one transaction, so a consume-race loser rolls back the event time AND the blocked-time moves.

PR #122, opened 2026-06-18
Phase 3g — deleteBlockedTime mutation + delete_blocked_time service method

hugobessa:
## Summary
Adds the `deleteBlockedTime` mutation and the backing `AvailabilityService.delete_blocked_time` service method (true gap — no single blocked-time delete existed). Completes the availability/blocked-time area.

- `calendar_integration/services/availability_service.py`: `delete_blocked_time(calendar, blocked_time_id)` — org+calendar-scoped delete, missing/cross-calendar → `ValueError`, `@transaction.atomic()`. Plus `CalendarService` facade delegate.
- `public_api/mutations.py`: `deleteBlockedTime` mutation, `{success, errorMessage}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.DELETE_BLOCKED_TIME` + mapping.
- Migration `0016_alter_resourceaccess_resource_name.py` (choices-only).

`deleteSeries` is omitted (consistent with deleteAvailabilityWindow): a recurring blocked time is one rrule row, so deleting it removes the whole recurrence — documented in the resolver docstring.

## Plan reference
Plan → Phase 3g. No feature flag. Stacked on Phase 3f (PR #120).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2078 passed**. Tests: service unit (`TestDeleteBlockedTimeService` — delete, missing/cross-calendar → ValueError, org-scoping with other-org row preserved); mutation (`TestDeleteBlockedTimeMutation` — happy path row-gone, recurring row-gone, missing/cross-org → success=False, permission-denied, unauthenticated).

PR #132, opened 2026-06-18
Phase 4a — calendarBundles query + CalendarBundleGraphQLType

hugobessa:
## Summary
Makes calendar bundles (Calendars with `calendar_type=BUNDLE`) a first-class Public API read surface — the first piece of the bundles area.

- `calendar_integration/graphql.py`: new `CalendarBundleGraphQLType` over `Calendar` exposing `id, name, description, children` (children via the `bundle_children` M2M → `ChildrenCalendarRelationship`, reusing `CalendarGraphQLType`). No `isPrivate`/`owners` (separate plans — non-goals).
- `public_api/queries.py`: `calendarBundles(offset, limit)` — org-scoped, `.only_listed()` (so disabled bundles drop out), `calendar_type=BUNDLE`, `prefetch_related("bundle_children")` for N+1, `_slice_qs` pagination.
- `public_api/constants.py` + `permissions.py`: new `PublicAPIResources.CALENDAR_BUNDLE` category + mapping.
- Migration `0017_alter_resourceaccess_resource_name.py` (choices-only).

## Plan reference
Plan → Phase 4a. No feature flag. Stacked on Phase 3g (PR #122).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2087 passed**. Tests (`TestCalendarBundlesQuery`): lists bundles + children resolve; org isolation; only BUNDLE type returned; inactive bundle excluded (`.only_listed()`); pagination; permission-denied.


public_api/queries.py
```diff
@@ -460,6 +462,28 @@ def calendar_groups(
...
+        Children are prefetched to avoid N+1 queries.
+        """
+        org = _get_org(info)
+        qs = (
+            Calendar.objects.filter_by_organization(org.id)
+            .only_listed()
+            .filter(calendar_type=CalendarType.BUNDLE)
+            .prefetch_related("bundle_children")
+            .order_by("pk")
+        )
```
hugobessa, line 484: `.only_listed()` is load-bearing here: Phase 4d's `disableCalendarBundle` disables a bundle by setting `visibility=INACTIVE`, and this query must exclude inactive bundles for that disable to be observable (matches the `calendars` query). `prefetch_related("bundle_children")` avoids an N+1 on the children resolver — strawberry-django's optimizer can't auto-detect the self-M2M through a custom through table, so the prefetch is explicit.

calendar_integration/graphql.py
```diff
@@ -345,6 +345,27 @@ class CalendarGroupGraphQLType:
...
+    """
+
+    id: strawberry.auto  # noqa: A003
+    name: strawberry.auto
+    description: strawberry.auto
+
+    @strawberry_django.field
+    def children(self) -> list[CalendarGraphQLType]:
+        """Return the child calendars of this bundle calendar."""
+        return list(self.bundle_children.all())  # type: ignore[union-attr]
```
hugobessa, line 366: Bundles are modeled as `Calendar` rows with `calendar_type=BUNDLE`; children come from the `bundle_children` self-M2M (through `ChildrenCalendarRelationship`, bundle_calendar→child_calendar) and reuse `CalendarGraphQLType`. The `children` resolver uses `self` (the model instance bound by strawberry-django) and reads from the prefetch cache. Deliberately no `isPrivate`/`owners` — both are separate, independently-planned changes.

PR #133, opened 2026-06-18
Phase 4b — createCalendarBundle Public GraphQL mutation

hugobessa:
## Summary
Adds `createCalendarBundle` — creates a bundle calendar from child calendars. Thin wrapper over `CalendarService.create_bundle_calendar(name, description, child_calendars, primary_calendar)`.

- `public_api/mutations.py`: `createCalendarBundle` mutation, `{success, errorMessage, bundle: CalendarBundleGraphQLType}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`. Children resolved org-scoped (count-mismatch → success=False), `children_ids` deduped, primary validated ∈ children, `description` None→"".
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.CREATE_CALENDAR_BUNDLE` + mapping.
- Migration `0018_alter_resourceaccess_resource_name.py` (choices-only).
- No `isPrivate` (non-goal).

## Plan reference
Plan → Phase 4b. No feature flag. Stacked on Phase 4a (PR #132).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2096 passed**. Tests (`TestCreateCalendarBundleMutation`): happy path (bundle calendar_type=BUNDLE + children linked) + round-trip via calendarBundles; duplicate children_ids deduped; null description → ""; cross-org child → success=False; primary-not-in-children → success=False; permission-denied; unauthenticated.


public_api/mutations.py
```diff
@@ -1182,3 +1210,60 @@ def delete_blocked_time(
...
+            if input.primary_calendar_id not in unique_children_ids:
+                return CreateCalendarBundleResult(
+                    success=False,
+                    error_message="primary_calendar_id must be one of the children_ids.",
+                )
+            primary = next(c for c in children if c.id == input.primary_calendar_id)
+
+        try:
+            bundle = calendar_service.create_bundle_calendar(
+                name=input.name,
```
hugobessa, line 1260: Children are resolved org-scoped (`filter_by_organization(org.id).filter(id__in=...)`) and the fetched count is compared to the DEDUPED id set, so a cross-org/missing child → success=False and a duplicated id doesn't falsely fail. `primary_calendar_id` must be among the children ids (validated up front) and is then selected from the already-fetched org-scoped set — so a primary can never be a cross-org calendar. `description` is normalized None→"" (Calendar.description is NOT NULL). No isPrivate (separate plan).

PR #134, opened 2026-06-18
Phase 4c — updateCalendarBundle Public GraphQL mutation

hugobessa:
## Summary
Adds `updateCalendarBundle` — renames a bundle and/or reconciles its children + primary.

- `public_api/mutations.py`: `updateCalendarBundle` mutation, `{success, errorMessage, bundle}`, guarded by `IsAuthenticated` + `OrganizationResourceAccess`. The service `update_bundle_calendar` only reconciles children/primary, so `name`/`description` are updated in the resolver (None = leave unchanged). Bundle fetch + name/description save + the service call run inside one `transaction.atomic()` so a children-validation failure rolls back the rename too.
- `public_api/constants.py` + `permissions.py`: `PublicAPIResources.UPDATE_CALENDAR_BUNDLE` + mapping.
- Migration `0019_alter_resourceaccess_resource_name.py` (choices-only).
- No `isPrivate` (non-goal).

## Plan reference
Plan → Phase 4c. No feature flag. Stacked on Phase 4b (PR #133).

## Test plan
Docker compose gate. Outer gate green: ruff ✅, check --deploy ✅, makemigrations --check ✅, pytest **2108 passed**. Tests (`TestUpdateCalendarBundleMutation`): happy path (children reconciled + name/desc updated); None name/desc preserved; missing/cross-org bundle → success=False; non-bundle id → "Bundle not found"; cross-org child → success=False; child-is-bundle rejected; primary-not-in-children → success=False; rollback (name NOT persisted on service failure); permission-denied; unauthenticated.


public_api/mutations.py
```diff
@@ -1267,3 +1296,88 @@ def create_calendar_bundle(
...
+                if update_fields:
+                    bundle.save(update_fields=update_fields)
+
+                updated_bundle = calendar_service.update_bundle_calendar(
+                    bundle_calendar=bundle,
+                    child_calendars=children,
+                    primary_calendar=primary,
+                )
+        except Calendar.DoesNotExist:
+            return UpdateCalendarBundleResult(success=False, error_message="Bundle not found.")
```
hugobessa, line 1376: The service `update_bundle_calendar` reconciles children/primary only — it does NOT touch name/description, so those are set on the bundle Calendar here (None = leave unchanged, saved via `update_fields`). The bundle fetch, the name/description save, AND the service call are wrapped in ONE `transaction.atomic()` inside the try: if the service raises (cross-org child, child-is-bundle, primary-not-in-children), the exception exits the `with` and rolls back the rename before the `except` returns success=False — preventing a partial write (renamed bundle but unchanged children). A dedicated test asserts name/description are NOT persisted on service failure.

PR #135, opened 2026-06-18
feat(calendar): cancel an event with a code (Phase 6c)

hugobessa:
## Summary

Unauthenticated `cancelEventWithCode` (input `{ code }`, no `permission_classes`). One mutation handles
both a calendar-bound (non-grouped) and a group-bound (grouped) CANCEL code. Flow: `resolve_code` →
require CANCEL + event scope → `transaction.atomic { consume_code → delete }`.

- Calendar path: `delete_event` (gated by `can_perform_update(new=None)` → requires CANCEL + event
  binding; authorized via the code threaded as the token).
- Group path: new `CalendarGroupService.cancel_grouped_event` deletes the linked non-primary
  `BlockedTime`s (by `external_id` LIKE `group-event-{id}-cal-`) then deletes the primary event
  (`CalendarEventGroupSelection` rows cascade). Without this the blocked times would orphan (string
  link, not FK).
- **consume BEFORE delete** (both inside one transaction): the token's `event` FK is `on_delete=CASCADE`,
  so deleting the event removes the token; running consume first and letting the txn roll back on any
  delete failure keeps the single-use guarantee (a FAILED cancel does not burn the code — there's a test).

## Documented deviation (accepted)

Replay of a used cancel code returns **INVALID_CODE**, not ALREADY_USED (the plan's acceptance said
ALREADY_USED): a successful cancel cascade-deletes the token with the event, so replay hits DoesNotExist.
Single-use is still genuinely enforced (the code can never perform a second action). Trade-off: the
`used_at`/IP audit row isn't retained for cancellations (the event is gone anyway). Retaining it would
need `token.event` → SET_NULL on the composite OrganizationForeignKey — deferred. (See the tracking file.)

## Plan reference

- Plan: `ai-plans/2026-06-17-SINGLE_USE_SCHEDULING_CODES_IMPLEMENTATION_PLAN.md` — Phase 6c.
- Spec: Decisions → Use-cases, Use-case 4 (patient cancels).

## Test plan

In-container (docker postgres): `public_api/tests/test_cancel_with_code.py` — calendar cancel (event gone +
consumed), group cancel (event gone + selections gone + non-primary blocked times deleted, none orphaned +
consumed), **delete-failure → code stays live + event intact** (atomicity), replay → INVALID_CODE,
wrong-permission (event survives), expired/revoked/invalid (event survives), cancels-exactly-the-bound-event.
`ruff` clean · `makemigrations --check` no changes · `check --deploy` clean · `pytest -n auto` **2174 passed**.

Stacked on `plan/single-use-scheduling-codes/phase-6b`. **Final phase of the plan.**


calendar_integration/mutations.py
```diff
@@ -1793,3 +1801,179 @@ def reschedule_calendar_group_event_with_code(
...
+                    deps.calendar_group_service.calendar_service = deps.calendar_service
+                    deps.calendar_group_service.initialize(organization=org)
+                    deps.calendar_group_service.cancel_grouped_event(
+                        event_id=event_id,
+                        delete_series=False,
+                    )
+                else:
+                    # Single-calendar cancel path.
+                    deps.calendar_service.delete_event(
+                        calendar_id=single_calendar_id,  # type: ignore[arg-type]
```
hugobessa, line 1921: consume BEFORE delete, both inside one transaction.atomic(). The token's `event` FK is on_delete=CASCADE, so deleting the event removes the token — running consume first (it sets used_at under SELECT FOR UPDATE) then delete means a FAILED delete rolls the consume back (the code stays live — there's an atomicity test), and a concurrent loser's consume raises → its delete rolls back too (first-write-wins). Routing: group-scoped code → cancel_grouped_event (blocked-time cleanup); else delete_event on the bound event's calendar. event_id is strictly from the token.

calendar_integration/services/calendar_group_service.py
```diff
@@ -486,6 +486,65 @@ def create_grouped_event(self, data: CalendarGroupEventInputData) -> CalendarEve
...
+                f"Event {event_id} is not a grouped event (calendar_group_fk is not set)."
+            )
+
+        primary_calendar_id: int = event.calendar_fk_id  # type: ignore[assignment]
+
+        # Delete non-primary BlockedTimes first (string-linked, NOT cascaded).
+        BlockedTime.objects.filter_by_organization(self.organization.id).filter(
+            external_id__startswith=f"group-event-{event_id}-cal-"
+        ).delete()
+
```
hugobessa, line 540: Group cancel must explicitly delete the non-primary BlockedTimes — they're linked to the grouped event only by the string `external_id` (`group-event-{id}-cal-{cid}`), NOT a FK, so they would orphan when the primary event is deleted (CalendarEventGroupSelection rows DO cascade via FK). Org-scoped delete by the external_id prefix, then delete the primary event. Runs in the mutation's atomic block, so a consume-race rolls back both. A test asserts zero orphaned blocked times after a group cancel.

PR #138, opened 2026-06-19
feat(public_api): owner-scope the blocked-time write mutations

hugobessa:
## Summary
Phase 1 of the **Per-Owner-Scoped Public API Token Writes** plan — the continuation of the
Per-Owner-Scoped Public API Tokens feature that delivers the provider *write* capabilities against a
`main` that already ships org-wide write mutations. This phase makes the existing
`createBlockedTime` / `updateBlockedTime` / `deleteBlockedTime` mutations **owner-aware**: a
provider-scoped token may only write blocked times on its owner's calendars; org-wide tokens are
byte-for-byte unchanged.

Rather than add parallel mutations (which would collide on the GraphQL field name and double the
enforcement surface), the existing mutations get a single shared guard call.

## What changed
- **`public_api/scoping.py`**: new shared write guard `assert_calendar_in_owner_scope(system_user,
  organization, calendar_id)`. No-op when `system_user is None` and when the token is org-wide
  (`scoped_calendar_ids` returns `None`); otherwise raises `Calendar.DoesNotExist` with the same
  message a genuinely-missing calendar produces. Reused by phases 2 and 3.
- **`public_api/mutations.py`**: the guard is called in all three blocked-time mutations, *inside* the
  existing `try/except Calendar.DoesNotExist`, before the calendar fetch and the service call — so a
  cross-owner attempt is indistinguishable from a genuinely-missing calendar (`success=False`,
  `"Calendar not found."`, no row touched). Org-wide behavior is unchanged (guard short-circuits).
- **`public_api/constants.py`**: `CREATE_BLOCKED_TIME`, `UPDATE_BLOCKED_TIME`, `DELETE_BLOCKED_TIME`
  added to `PROVIDER_SCOPED_RESOURCES` so a scoped token can be granted them — one resource vocabulary
  for org-wide and scoped tokens; owner-scope stays an orthogonal layer.

## Plan reference
Plan `ai-plans/2026-06-18-PER_OWNER_SCOPED_PUBLIC_API_TOKEN_WRITES_IMPLEMENTATION_PLAN.md` — Phase 1.
Spec `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_SPEC.md`: use-case 2 (provider manages its
owner's blocked times) + use-case 3 (cross-owner refused as not-found). Stacks on phase-3 of the
original plan (PR #107), which carries the read-scoping + mint + `scoped_calendar_ids` infra.

## Test plan
- `public_api/tests/test_scoping.py` — `assert_calendar_in_owner_scope`: no-raise for `None` user, for
  org-wide token, and for in-scope calendar; raises for out-of-scope and for a nonexistent id when scoped.
- `public_api/tests/test_mutations.py` (`TestScopedTokenBlockedTimeWrites`) — per verb (create/update/
  delete): scoped token succeeds on its owned calendar; cross-owner `calendarId` returns a not-found
  byte-identical to a genuinely-missing calendar AND writes no row; org-wide token unaffected; missing
  resource grant denied. Plus a pinned regression: a scoped token passing its OWN `calendarId` with a
  `blockedTimeId` from another calendar is rejected (`success=False`, the foreign row unchanged) by the
  service's `calendar_fk=calendar` filter.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2392 passed**.

### DIFF
--- a/public_api/constants.py
+++ b/public_api/constants.py
@@ -63,5 +63,8 @@ PROVIDER_SCOPED_RESOURCES: frozenset[str] = frozenset(
         PublicAPIResources.AVAILABILITY_WINDOWS,
         PublicAPIResources.UNAVAILABLE_WINDOWS,
         PublicAPIResources.CALENDAR,
+        PublicAPIResources.CREATE_BLOCKED_TIME,
+        PublicAPIResources.UPDATE_BLOCKED_TIME,
+        PublicAPIResources.DELETE_BLOCKED_TIME,
     ]
 )
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -31,6 +31,7 @@ from public_api.capabilities import assert_org_can_invite, assert_target_in_subt
 from public_api.constants import PROVIDER_SCOPED_RESOURCES, PublicAPIResources
 from public_api.models import ResourceAccess, SystemUser
 from public_api.permissions import IsAuthenticated, OrganizationResourceAccess
+from public_api.scoping import assert_calendar_in_owner_scope
 from public_api.services import PublicAPIAuthService
 from public_api.types import (
     BrandingResult,
@@ -1452,15 +1453,21 @@ class Mutation(CalendarGroupMutations):
 
         The mutation:
         1. Resolves the organization and initializes the calendar service via the system-user token.
-        2. Fetches the calendar org-scoped to prevent cross-org access.
-        3. Delegates to CalendarService.create_blocked_time with the supplied parameters.
-        4. Returns the created BlockedTime on success, or success=False + errorMessage on failure.
+        2. Asserts the calendar is within the token owner's scope (no-op for org-wide tokens).
+        3. Fetches the calendar org-scoped to prevent cross-org access.
+        4. Delegates to CalendarService.create_blocked_time with the supplied parameters.
+        5. Returns the created BlockedTime on success, or success=False + errorMessage on failure.
 
         The token's OrganizationResourceAccess must include the CREATE_BLOCKED_TIME resource.
         """
         calendar_service, org = _get_org_and_init_calendar_service(info)
+        request: PublicApiHttpRequest = info.context.request
 
         try:
+            # Owner-scope guard: a scoped token may only write to its owner's calendars.
+            # Raises Calendar.DoesNotExist (same as a genuinely missing calendar) so a
+            # cross-owner attempt reveals nothing about the target's existence.
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, input.calendar_id)
             calendar = Calendar.objects.filter_by_organization(org.id).get(id=input.calendar_id)
         except Calendar.DoesNotExist:
             return CreateBlockedTimeResult(success=False, error_message="Calendar not found.")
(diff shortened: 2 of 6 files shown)


public_api/scoping.py
```diff
@@ -25,3 +25,34 @@ def scoped_calendar_ids(system_user: SystemUser, organization: Organization) ->
...
+    Raises:
+        Calendar.DoesNotExist: When the token is scoped and calendar_id is outside
+            the owner's allowed calendar set. The message is intentionally identical
+            to a real not-found error to prevent existence leaks.
+    """
+    if system_user is None:
+        return
+    allowed_ids = scoped_calendar_ids(system_user, organization)
+    if allowed_ids is not None and calendar_id not in allowed_ids:
+        raise Calendar.DoesNotExist("Calendar matching query does not exist.")
```
hugobessa, line 58: The single write-side enforcement primitive, deliberately mirroring the read guard (_prepare_service_and_calendar) and reusing scoped_calendar_ids so reads and writes enforce scope identically — one place to audit. Two invariants encode the plan's Guiding Decisions: (1) it is a strict no-op for org-wide tokens (scoped_calendar_ids returns None) and for system_user=None, which is what keeps existing org-wide write behavior byte-for-byte; (2) it raises Calendar.DoesNotExist with the SAME message a genuinely-missing calendar produces, so a cross-owner write never confirms the target exists ("not-found, never forbidden"). An empty allowed set (scoped owner with zero calendars) correctly denies everything — empty set is not None, so any id is "not in" it.

public_api/mutations.py
```diff
@@ -1250,15 +1251,21 @@ def create_blocked_time(
...
         The token's OrganizationResourceAccess must include the CREATE_BLOCKED_TIME resource.
         """
         calendar_service, org = _get_org_and_init_calendar_service(info)
+        request: PublicApiHttpRequest = info.context.request
 
         try:
+            # Owner-scope guard: a scoped token may only write to its owner's calendars.
+            # Raises Calendar.DoesNotExist (same as a genuinely missing calendar) so a
+            # cross-owner attempt reveals nothing about the target's existence.
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, input.calendar_id)
```
hugobessa, line 1470: Placement is load-bearing: the guard sits INSIDE the existing `try/except Calendar.DoesNotExist`, before the org-scoped fetch. So the guard's DoesNotExist and a genuinely-missing calendar's DoesNotExist both fall into the same handler returning success=False + "Calendar not found." — the two cases are byte-identical in the response (no separate error path, no existence leak). Same pattern is applied identically to updateBlockedTime (line 1316) and deleteBlockedTime (line 1370). The guard runs before the service call, so a cross-owner attempt never reaches the DB write.

public_api/constants.py
```diff
@@ -63,5 +63,8 @@ class PublicAPIResources(TextChoices):
         PublicAPIResources.AVAILABILITY_WINDOWS,
         PublicAPIResources.UNAVAILABLE_WINDOWS,
         PublicAPIResources.CALENDAR,
+        PublicAPIResources.CREATE_BLOCKED_TIME,
+        PublicAPIResources.UPDATE_BLOCKED_TIME,
```
hugobessa, line 67: These map to the SAME resource names the org-wide mutations already gate on (createBlockedTime → CREATE_BLOCKED_TIME, etc.). Adding them to PROVIDER_SCOPED_RESOURCES lets a scoped token be granted them via the existing mint validation, without re-mapping the mutations (which would change the resource gate for existing org-wide callers). Resource grant and owner scope remain independent layers: holding the resource lets you call the mutation; the owner guard still confines every write to your owner's calendars.

public_api/tests/test_mutations.py
```diff
@@ -6578,3 +6578,818 @@ def test_inactive_member_owner_rejected(self):
...
+        assert cross_msg == missing_msg
+
+        # Row must still exist — the cross-owner delete must NOT have succeeded
+        assert BlockedTime.objects.filter(id=bt_b_id).exists()
+
+    # ------------------------------------------------------------------
+    # blockedTimeId / calendarId mismatch — service-layer filter is load-bearing
+    # ------------------------------------------------------------------
+
+    def test_update_delete_blocked_time_in_scope_calendar_with_foreign_blocked_time_id_rejected(
```
hugobessa, line 7096: This pins the second-vector defense surfaced in review: a scoped token can pass its OWN (in-scope) calendarId so the new owner guard does NOT trip, but a blockedTimeId belonging to a different calendar. The protection is pre-existing service code (availability_service scopes the BlockedTime lookup to calendar_fk=calendar → foreign id yields success=False), but the other cross-owner tests pass the cross-owner calendarId and would still pass if that filter regressed. This test fails if that load-bearing constraint is ever removed.

PR #139, opened 2026-06-19
feat(public_api): owner-scope the availability-window write mutations

hugobessa:
## Summary
Phase 2 of the **Per-Owner-Scoped Public API Token Writes** plan. Makes the four existing
availability-window write mutations — `createAvailabilityWindow`, `updateAvailabilityWindow`,
`deleteAvailabilityWindow`, `batchUpdateAvailabilityWindows` (all of which write `AvailableTime` rows)
— **owner-aware**: a provider-scoped token may only write availability on its owner's calendars;
org-wide tokens are byte-for-byte unchanged. Structurally a mirror of Phase 1 (blocked-time writes),
reusing the same shared guard.

## What changed
- **`public_api/mutations.py`**: `assert_calendar_in_owner_scope(...)` (the Phase-1 guard) is called in
  all four mutations, inside each existing `try/except Calendar.DoesNotExist`, before the calendar
  fetch and the service call — cross-owner → byte-identical not-found, no row touched. For
  `batchUpdateAvailabilityWindows` the single top-level `input.calendar_id` governs the whole atomic
  batch, so one guard up front rejects a cross-owner batch wholesale with no partial write.
- **`public_api/constants.py`**: `CREATE/UPDATE/DELETE_AVAILABILITY_WINDOW` and
  `BATCH_UPDATE_AVAILABILITY_WINDOWS` added to `PROVIDER_SCOPED_RESOURCES`.

## Plan reference
Plan `ai-plans/2026-06-18-PER_OWNER_SCOPED_PUBLIC_API_TOKEN_WRITES_IMPLEMENTATION_PLAN.md` — Phase 2.
Spec `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_SPEC.md`: use-case 2 (provider manages its
owner's availability — recurring availability + specific dates) + use-case 3 (cross-owner refused).
Stacked on phase-1 (PR #138).

## Test plan
- `public_api/tests/test_mutations.py` (`TestScopedTokenAvailabilityWrites`, 17 tests) — per verb
  (create no-rrule/rrule, update, delete, batch): scoped token succeeds on its owned calendar;
  cross-owner `calendarId` not-found byte-identical to a genuinely-missing calendar AND no
  `AvailableTime` row created/changed/deleted; org-wide token unaffected; missing grant denied. The
  batch cross-owner test mixes create+update+delete ops against a seeded B row and asserts the whole
  batch is rejected with the B row unchanged (no partial write).
- The `available_time_id` second-vector (own calendar_id + foreign available_time_id) is constrained by
  the service (`filter(calendar_fk=calendar)` → `ValueError`), verified in review.
- Outer gate (docker): `check --deploy` + `pytest -n auto` → **2455 passed**.

### DIFF
--- a/public_api/constants.py
+++ b/public_api/constants.py
@@ -66,5 +66,9 @@ PROVIDER_SCOPED_RESOURCES: frozenset[str] = frozenset(
         PublicAPIResources.CREATE_BLOCKED_TIME,
         PublicAPIResources.UPDATE_BLOCKED_TIME,
         PublicAPIResources.DELETE_BLOCKED_TIME,
+        PublicAPIResources.CREATE_AVAILABILITY_WINDOW,
+        PublicAPIResources.UPDATE_AVAILABILITY_WINDOW,
+        PublicAPIResources.DELETE_AVAILABILITY_WINDOW,
+        PublicAPIResources.BATCH_UPDATE_AVAILABILITY_WINDOWS,
     ]
 )
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -1223,16 +1223,22 @@ class Mutation(CalendarGroupMutations):
 
         The mutation:
         1. Resolves the organization and initializes the calendar service via the system-user token.
-        2. Fetches the calendar org-scoped to prevent cross-org access.
-        3. Delegates to CalendarService.create_available_time with the supplied parameters.
-        4. Returns the created AvailableTime on success, or success=False + errorMessage on failure.
+        2. Asserts the calendar is within the token owner's scope (no-op for org-wide tokens).
+        3. Fetches the calendar org-scoped to prevent cross-org access.
+        4. Delegates to CalendarService.create_available_time with the supplied parameters.
+        5. Returns the created AvailableTime on success, or success=False + errorMessage on failure.
            Note: the service raises ValueError if calendar.manage_available_windows is False.
 
         The token's OrganizationResourceAccess must include the CREATE_AVAILABILITY_WINDOW resource.
         """
         calendar_service, org = _get_org_and_init_calendar_service(info)
+        request: PublicApiHttpRequest = info.context.request
 
         try:
+            # Owner-scope guard: a scoped token may only write to its owner's calendars.
+            # Raises Calendar.DoesNotExist (same as a genuinely missing calendar) so a
+            # cross-owner attempt reveals nothing about the target's existence.
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, input.calendar_id)
             calendar = Calendar.objects.filter_by_organization(org.id).get(id=input.calendar_id)
         except Calendar.DoesNotExist:
             return CreateAvailabilityWindowResult(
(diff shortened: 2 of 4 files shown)


public_api/mutations.py
```diff
@@ -1021,16 +1021,22 @@ def create_availability_window(
...
         The token's OrganizationResourceAccess must include the CREATE_AVAILABILITY_WINDOW resource.
         """
         calendar_service, org = _get_org_and_init_calendar_service(info)
+        request: PublicApiHttpRequest = info.context.request
 
         try:
+            # Owner-scope guard: a scoped token may only write to its owner's calendars.
+            # Raises Calendar.DoesNotExist (same as a genuinely missing calendar) so a
+            # cross-owner attempt reveals nothing about the target's existence.
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, input.calendar_id)
```
hugobessa, line 1241: Same owner-guard pattern as Phase 1's blocked-time mutations, applied to all four availability write verbs (here createAvailabilityWindow; also updateAvailabilityWindow at 1089, deleteAvailabilityWindow at 1159, and the batch at ~1212). Placed inside the existing `try/except Calendar.DoesNotExist` before the org-scoped fetch, so a scoped token's cross-owner write is byte-identical to a genuinely-missing calendar (success=False, "Calendar not found.") and never reaches the service / DB. Org-wide tokens short-circuit (no-op).

public_api/mutations.py
```diff
@@ -1164,12 +1182,16 @@ def batch_update_availability_windows(
...
         1. Resolves the organization and initializes the calendar service via the system-user token.
-        2. Fetches the calendar org-scoped to prevent cross-org access.
-        3. Validates every operation's action is one of {create, update, delete}.
-        4. Translates each BatchAvailabilityOperationInput into the service dict shape
+        2. Asserts the calendar is within the token owner's scope (no-op for org-wide tokens).
+           The single input.calendar_id governs the whole atomic batch; one guard call up front
+           rejects a cross-owner batch wholesale with no partial write (individual operations
+           share the same calendar_id — they carry no per-op calendar_id of their own).
+        3. Fetches the calendar org-scoped to prevent cross-org access.
+        4. Validates every operation's action is one of {create, update, delete}.
```
hugobessa, line 1392: Batch atomicity rationale (the highest-risk spot): BatchAvailabilityInput carries ONE top-level calendar_id and the per-operation BatchAvailabilityOperationInput entries carry NO calendar_id of their own — so a single guard call governs every op. A cross-owner batch is therefore rejected wholesale before the service runs, with no partial write. (Confirmed against the input type in review.) A foreign available_time_id paired with an owned calendar_id can't slip a write through either: the availability service constrains every update/delete to filter(calendar_fk=calendar).

public_api/constants.py
```diff
@@ -66,5 +66,9 @@ class PublicAPIResources(TextChoices):
         PublicAPIResources.CREATE_BLOCKED_TIME,
         PublicAPIResources.UPDATE_BLOCKED_TIME,
         PublicAPIResources.DELETE_BLOCKED_TIME,
+        PublicAPIResources.CREATE_AVAILABILITY_WINDOW,
+        PublicAPIResources.UPDATE_AVAILABILITY_WINDOW,
+        PublicAPIResources.DELETE_AVAILABILITY_WINDOW,
+        PublicAPIResources.BATCH_UPDATE_AVAILABILITY_WINDOWS,
```
hugobessa, line 72: These four map to the SAME resource names the org-wide availability mutations already gate on. Adding them to PROVIDER_SCOPED_RESOURCES lets a scoped token be granted them via the existing mint validation, without re-mapping the mutations (which would change the gate for existing org-wide callers). Resource grant and owner scope stay independent layers.

public_api/tests/test_mutations.py
```diff
@@ -7392,4 +7392,1039 @@ def test_delete_blocked_time_scoped_token_missing_grant_denied(self):
...
+        assert "errors" not in data or len(data.get("errors", [])) == 0
+        result = data["data"]["batchUpdateAvailabilityWindows"]
+        assert result["success"] is True
+        assert result["availableTimes"] is not None
+
+    # ------------------------------------------------------------------
+    # batchUpdateAvailabilityWindows — cross-owner not-found + no partial write
+    # ------------------------------------------------------------------
+
+    def test_batch_update_availability_windows_cross_owner_rejected_wholesale(self):
```
hugobessa, line 8004: The batch wholesale-rejection test was hardened after review: it seeds a pre-existing AvailableTime row on provider B's calendar and the cross-owner batch mixes create + update + delete ops targeting that B row, then asserts the entire batch is rejected AND B's row still exists with unchanged start/end values (captured before the attempt). This proves no partial update/delete leaks through — a count-only check would have caught a stray create but not a partial mutation.

PR #140, opened 2026-06-19
feat(public_api): add owner-scoped scheduleEvent mutation

hugobessa:
## Summary
Phase 3 of the **Per-Owner-Scoped Public API Token Writes** plan — the only net-new mutation. A
provider-scoped token can now **schedule events on its owner's calendars** via `scheduleEvent`. This
required a narrow, sanctioned relaxation of `CalendarEventService.create_event`, which previously
hard-blocked ALL public-API (`SystemUser`) event creation.

## Why the service changed
`create_event` raised `PermissionDenied("Events cannot be created through the Public API.")` for every
`SystemUser`. This phase relaxes ONLY the `create_event` site (the `scheduleEvent` path —
`update_event`/`delete_event` blocks are untouched) to allow event creation when, and only when:
- the token is owner-scoped (`scoped_to_membership_fk_id is not None`), AND
- an **independent** `CalendarOwnership` lookup (filtered by the calendar's own organization) confirms
  the token's owner owns the target calendar — nothing from the caller is trusted.

Org-wide tokens stay blocked (event creation still routes through single-use codes / public scheduling).
Bundle calendars are rejected for scoped tokens (avoids confusing cross-provider fan-out).

## What changed
- **`calendar_integration/services/calendar_event_service.py`**: `_scoped_system_user_owns_calendar`
  helper (independent ownership verification); owner-scoped allowance + bundle rejection in
  `create_event`; skip `can_perform_scheduling` ONLY on the verified-ownership path (that method is a
  pure token-permission gate, not availability enforcement — availability still runs); fixed the
  post-commit audit actor to `getattr`-guard the permission token and fall back to the `SystemUser`.
- **`public_api/mutations.py`**: new `scheduleEvent` mutation (`ScheduleEventInput`), owner-guarded via
  `assert_calendar_in_owner_scope` (defense in depth); de-duplicated active-org-member attendee
  validation; external-attendee translation; title-length guard; clean GraphQL error mapping.
- **`public_api/permissions.py`**: `scheduleEvent → CALENDAR_EVENT` (already in `PROVIDER_SCOPED_RESOURCES`).

## Plan reference
Plan `ai-plans/2026-06-18-PER_OWNER_SCOPED_PUBLIC_API_TOKEN_WRITES_IMPLEMENTATION_PLAN.md` — Phase 3.
Spec `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_SPEC.md`: use-case 2 (schedule-event write)
+ use-case 3 (cross-owner refused). Stacked on phase-2 (PR #139).

## Test plan
- `public_api/tests/test_mutations.py` (`TestScopedTokenScheduleEvent`, 10 tests) — scoped owner
  schedules on a default (non-public) owned calendar with a covering availability window → success;
  recurring; internal + external attendees persisted; out-of-org attendee → clean error, no row;
  cross-owner calendar → not-found identical to missing + no row; **org-wide token → denied, no row**;
  bundle → clean error, no row; no availability window → clean error; missing `CALENDAR_EVENT` → denied.
- `calendar_integration/tests/services/test_calendar_event_service.py` (4 tests) — `create_event` allows
  an owner-scoped `SystemUser` on an owned calendar; still blocks an org-wide `SystemUser`; still blocks
  a scoped token on a non-owned calendar; the on_commit audit actor falls back to the `SystemUser`.
- Outer gate (docker): `check --deploy` + `makemigrations --check` (no changes) + `pytest -n auto` →
  **2469 passed**.

### DIFF
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -44,6 +44,7 @@ from calendar_integration.models import (
     Calendar,
     CalendarEvent,
     CalendarManagementToken,
+    CalendarOwnership,
     EventAttendance,
     EventBulkModification,
     EventExternalAttendance,
@@ -188,6 +189,39 @@ class CalendarEventService:
             context.organization,
         )
 
+    @staticmethod
+    def _scoped_system_user_owns_calendar(system_user: SystemUser, calendar: Calendar) -> bool:
+        """Independently verify the scoped token's owner owns ``calendar``.
+
+        This is the narrow, sanctioned event-creation allowance for owner-scoped public-API
+        tokens. The check does NOT trust anything supplied by the caller: it re-derives the
+        ownership relation from the database, scoped to the calendar's own organization, so a
+        cross-organization or cross-owner token can never pass.
+
+        Ownership is confirmed when a :class:`CalendarOwnership` row exists for ``calendar``
+        whose ``user`` is the member behind the token's ``scoped_to_membership`` (an
+        ``OrganizationMembership``). Org-wide tokens (``scoped_to_membership_fk_id is None``)
+        always return ``False`` — event creation stays blocked for them.
+
+        Args:
+            system_user: The token (``SystemUser``) attempting the write.
+            calendar: The target calendar the event would be created on.
+
+        Returns:
+            ``True`` only when the token is scoped AND its owner independently owns the
+            target calendar in that calendar's organization; ``False`` otherwise.
+        """
+        if system_user.scoped_to_membership_fk_id is None:
+            return False
+        return (
+            CalendarOwnership.objects.filter_by_organization(calendar.organization_id)
+            .filter(
+                calendar_fk_id=calendar.id,
+                user__organization_memberships=system_user.scoped_to_membership_fk_id,
+            )
+            .exists()
+        )
+
     def _serialize_event(self, event: CalendarEvent) -> CalendarEventData:
         """Build webhook payload for calendar event."""
         return _serialize_event_util(event)
@@ -237,6 +271,10 @@ class CalendarEventService:
 
         calendar = self._get_calendar_by_id(calendar_id)
 
+        # When True, authorization is granted by independently-verified calendar ownership
+        # (the owner-scoped public-API allowance) rather than by the permission-token flow.
+        is_owner_scoped_system_user = False
+
         if isinstance(context.user_or_token, User):
             context.calendar_permission_service.initialize_with_user(
                 context.user_or_token,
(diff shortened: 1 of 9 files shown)


calendar_integration/services/calendar_event_service.py
```diff
@@ -188,6 +189,39 @@ def _get_calendar_by_id(self, calendar_id: int) -> Calendar:
...
+
+        Returns:
+            ``True`` only when the token is scoped AND its owner independently owns the
+            target calendar in that calendar's organization; ``False`` otherwise.
+        """
+        if system_user.scoped_to_membership_fk_id is None:
+            return False
+        return (
+            CalendarOwnership.objects.filter_by_organization(calendar.organization_id)
+            .filter(
```
hugobessa, line 218: The sanctioned event-creation allowance, isolated in one auditable helper. It does NOT trust the caller: it re-derives ownership from the DB scoped to the calendar's OWN organization (filter_by_organization(calendar.organization_id)) and matches the calendar's CalendarOwnership to the membership behind the token (user__organization_memberships=<membership pk>, a single concrete membership → exactly one user, so no M2M ambiguity). Org-wide tokens (scoped_to_membership_fk_id is None) return False → event creation stays blocked. Cross-org is structurally impossible.

calendar_integration/services/calendar_event_service.py
```diff
@@ -237,22 +271,45 @@ def create_event(self, calendar_id: int, event_data: CalendarEventInputData) ->
...
 
-        if not context.calendar_permission_service.can_perform_scheduling(
-            calendar_id=calendar_id,
-            calendar_settings=CalendarSettingsData(
-                manage_available_windows=calendar.manage_available_windows,
-                accepts_public_scheduling=calendar.accepts_public_scheduling,
-            ),
-            event=event_data,
+        # The permission-token scheduling check applies to the User / token flows. The
+        # owner-scoped path's authorization is the independently-verified ownership above,
```
hugobessa, line 302: The relaxation point. Only the create_event SystemUser block is relaxed (update_event/delete_event keep their blocks). A scoped token reaches here only if it owns the calendar (checked just above); everything else still raises the original PermissionDenied. Bundle calendars are rejected BEFORE any fan-out — without this, an owner-scoped token owning a bundle would recurse into per-child creates spanning other providers' calendars (the BLOCKER from the original attempt at this change).

calendar_integration/services/calendar_event_service.py
```diff
@@ -237,22 +271,45 @@ def create_event(self, calendar_id: int, event_data: CalendarEventInputData) ->
...
+        # so it bypasses this check (the permission service has no token initialized for it).
+        if not is_owner_scoped_system_user and (
```
hugobessa, line 304: can_perform_scheduling is SKIPPED only on the verified-ownership path. That method is a pure token-permission gate (accepts_public_scheduling / token-calendar-scoped+CREATE / group-slot) and has no token initialized for a SystemUser, so it would always deny — but it does NOT enforce availability windows (that runs separately, still active for this path). Ownership IS the authorization here, so skipping the token gate is correct and loses no availability protection.

calendar_integration/services/calendar_event_service.py
```diff
@@ -402,17 +459,23 @@ def create_event(self, calendar_id: int, event_data: CalendarEventInputData) ->
...
 
+        # Resolve the audit actor *before* queueing the post-commit side-effect.
+        # The owner-scoped public-API path never initializes a permission token, so
+        # ``calendar_permission_service.token`` may be unset entirely — read it through
+        # ``getattr`` to avoid an AttributeError after commit, and fall back to the
+        # SystemUser caller so owner-scoped events are never actor-less.
+        permission_token = getattr(context.calendar_permission_service, "token", None)
+        if permission_token is not None and permission_token.user:
+            audit_actor: Any = permission_token.user
+        elif permission_token is not None:
```
hugobessa, line 470: on_commit audit-actor fix (the second BLOCKER from the original attempt). The owner-scoped path never initializes a permission token, so reading permission_service.token directly would AttributeError AFTER commit (invisible under pytest's on_commit rollback). Now getattr-guarded, with the actor falling back to the SystemUser so owner-scoped events are never actor-less. Equivalent to the old expression for the existing User/token callers (which always set .token).

public_api/mutations.py
```diff
@@ -1578,3 +1625,99 @@ def disable_calendar_bundle(
             return DisableCalendarBundleResult(success=False, error_message=str(e))
 
         return DisableCalendarBundleResult(success=True)
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def schedule_event(
```
hugobessa, line 1832: scheduleEvent returns a bare CalendarEventGraphQLType (per plan API design), so errors surface as GraphQLError. The owner guard (assert_calendar_in_owner_scope) runs FIRST and raises "Calendar not found." for a cross-owner OR scoped-but-non-owned calendar — BEFORE the service's PermissionDenied could surface a different message that reveals the calendar exists. Cross-owner and genuinely-missing are byte-identical (no existence leak). Internal attendees are de-duplicated and pre-validated as active org members before any write, so a nonexistent/out-of-org id is a clean error, not an IntegrityError 500.

PR #141, opened 2026-06-19
feat(public_api): owner-scope nested GraphQL field traversal (security sweep)

hugobessa:
## Summary
Phase 4 (final) of the **Per-Owner-Scoped Public API Token Writes** plan — the bypass-surface security
capstone. Top-level reads and writes are already owner-scoped (phases 1–3 + the merged read scoping),
but Strawberry runs field logic only on the decorated top-level field, so a provider-scoped token
fetching its OWN object could traverse **nested GraphQL fields** to another owner's data. This phase
closes that surface across the shared types and ships the security sign-off.

## How
`_owner_scoped_calendar_ids(info)` returns `None` (no filtering) for org-wide tokens AND internal
(non-public-API) requests — so those consumers stay byte-for-byte unchanged — or the owner's calendar-id
`set[int]` for a provider-scoped token. Every cross-owner-reachable nested field becomes a
`@strawberry.field` resolver that filters/suppresses for scoped tokens and is a strict no-op otherwise.

Guarded on `CalendarEventGraphQLType`: `calendar`, `bundleCalendar`, `bundlePrimaryEvent`,
`bulkModificationParent`, `parentRecurringObject`, `resources`, `resourceAllocations`, `groupSelections`,
`bundleRepresentations`, `bulkModifications`, `recurringInstances`; `calendarGroup` suppressed.
**Second hop** (the leak the original sweep first missed) closed two ways: `CalendarEventGroupSelection.slot`
suppressed for scoped tokens AND `CalendarGroupSlot.calendars` pool filtered. Also guarded
`BlockedTime.calendar` / `AvailableTime.calendar` + their recurrence-exception `parent_*`/`modified_*`
back-pointers, and `EventExternalAttendance.event`.

## What else
- Documented proven-safe (not over-guarded): scalars; people (`UserGraphQLType` is a leaf — no calendar
  relation re-exposed); `CalendarGroup.slots` (unreachable for scoped tokens — both nested entry points
  suppressed and `CALENDAR_GROUP` is not in `PROVIDER_SCOPED_RESOURCES`).
- Fixed a latent `recurringInstances` bug (the field mapped to a non-existent reverse accessor; the
  scoped resolver uses the correct one).
- `ai-plans/2026-06-18-PER_OWNER_SCOPED_PUBLIC_API_TOKEN_WRITES_SECURITY_REVIEW.md` — the bypass-surface
  checklist (every reachable field, its guard, its proving test). Sign-off artifact.

## Residual assumption (called out)
The top-level `calendarGroup*` resolvers are not owner-scoped; this is safe ONLY because `CALENDAR_GROUP`
is absent from `PROVIDER_SCOPED_RESOURCES`. If any group-returning resource is ever added to the
provider allow-list, those resolvers must gain owner filtering. Documented in the review doc.

## Plan reference
Plan `ai-plans/2026-06-18-PER_OWNER_SCOPED_PUBLIC_API_TOKEN_WRITES_IMPLEMENTATION_PLAN.md` — Phase 4.
Spec `ai-plans/2026-06-17-PER_OWNER_SCOPED_PUBLIC_API_TOKENS_SPEC.md`: objective 5 (no bypass) + use-case
3 (consolidated negative-path guarantee). Stacked on phase-3 (PR #140).

## Test plan
- `public_api/tests/test_scoping_security.py` (23 tests) — for every guarded nested field: a scoped
  token selecting it (including the second-hop `groupSelections { slot { calendars } }`) gets no
  cross-owner data, AND an org-wide/internal counterpart sees everything unchanged. Each fails if its
  guard is reverted (mutation-verified). Includes a direct-resolver test pinning the
  `CalendarGroupSlot.calendars` pool filter.
- Outer gate (docker): `check --deploy` + `makemigrations --check` (no changes) + `pytest -n auto` →
  **2492 passed**.


calendar_integration/graphql.py
```diff
@@ -26,6 +26,112 @@
...
+    if system_user is None:
+        return None
+    organization = getattr(request, "public_api_organization", None)
+    if organization is None:
+        return None
+
+    # Lazy import to break the public_api <-> calendar_integration import cycle.
+    from public_api.scoping import scoped_calendar_ids
+
+    return scoped_calendar_ids(system_user, organization)
```
hugobessa, line 76: The single source of truth for nested-field scoping. Returns None — meaning "do not filter" — for BOTH org-wide public-API tokens (scoped_calendar_ids returns None) AND internal/non-public-API requests (no public_api_system_user on the request). That None-means-unchanged contract is what keeps every existing org-wide and internal consumer of these shared types byte-for-byte identical: every resolver below early-returns the raw value when this is None. Lazy import of public_api.scoping avoids a public_api <-> calendar_integration cycle. An empty set (scoped owner with zero calendars) is not None, so it correctly filters everything.

calendar_integration/graphql.py
```diff
@@ -132,37 +252,113 @@ class CalendarEventGraphQLType:
...
+        return _scoped_calendar_list(list(self.resources.all()), _owner_scoped_calendar_ids(info))  # type: ignore
+
+    @strawberry.field
+    def calendar_group(self, info: strawberry.Info) -> "CalendarGroupGraphQLType | None":
+        """The booking CalendarGroup. Suppressed entirely for scoped tokens: a group
+        aggregates calendars across providers, so exposing it (and its slots' candidate
+        pool) would leak other owners' calendars."""
+        if _owner_scoped_calendar_ids(info) is not None:
+            return None
+        return self.calendar_group  # type: ignore[attr-defined,return-value]
```
hugobessa, line 322: calendarGroup is suppressed entirely for scoped tokens — a group aggregates calendars across providers via its slots, so it is not owner data. Org-wide tokens still see it. This is one of the two suppression points that together make CalendarGroup / CalendarGroupSlot unreachable by a scoped token (the other is the slot suppression below), which is why CalendarGroup.slots needs no separate guard.

calendar_integration/graphql.py
```diff
@@ -373,8 +605,21 @@ class CalendarEventGroupSelectionGraphQLType:
...
-    slot: CalendarGroupSlotGraphQLType = strawberry_django.field()
-    calendar: CalendarGraphQLType = strawberry_django.field()
+    @strawberry.field
+    def slot(self, info: strawberry.Info) -> "CalendarGroupSlotGraphQLType | None":
+        """The slot this selection belongs to. Suppressed for scoped tokens: a slot's
+        candidate pool aggregates calendars across providers, and reaching it via
+        ``groupSelections.slot`` is the second-hop bypass of the ``calendar_group``
+        suppression. The pool itself is also filtered in
+        ``CalendarGroupSlotGraphQLType.calendars`` as defence in depth."""
+        if _owner_scoped_calendar_ids(info) is not None:
```
hugobessa, line 615: The second-hop fix. groupSelections is owner-filtered, but each selection's `slot` exposed CalendarGroupSlot.calendars — the slot's ENTIRE candidate pool across providers — which would defeat the calendarGroup suppression via a sibling path (groupSelections.slot.calendars). Nested traversal doesn't re-run the top-level CALENDAR_GROUP permission gate, so a token holding only CALENDAR_EVENT could reach it. `slot` is now suppressed for scoped tokens; the pool at line 560 is ALSO filtered as defense-in-depth (with its own direct regression test).

calendar_integration/graphql.py
```diff
@@ -332,7 +556,15 @@ class CalendarGroupSlotGraphQLType:
...
 
-    calendars: list[CalendarGraphQLType] = strawberry_django.field()
+    @strawberry.field
+    def calendars(self, info: strawberry.Info) -> list["CalendarGraphQLType"]:
+        """The slot's candidate-calendar pool, filtered to the owner's set for scoped tokens.
+
+        This is the SECOND-HOP leak: a scoped token cannot reach a slot via the
+        suppressed ``calendar_group``, but it can still reach one through the sibling
+        path ``calendarEvent.groupSelections.slot.calendars``. Filtering the pool here
+        closes that path; the entire cross-provider candidate pool is otherwise exposed."""
```
hugobessa, line 566: Defense-in-depth pool filter on CalendarGroupSlot.calendars. The slot is already suppressed for scoped tokens (line 609), so this is currently unreachable via the scoped path — but it is pinned by a direct-resolver regression test (test_slot_calendars_pool_scoped_filters_cross_owner) so that if a future change ever re-exposes the slot, the pool stays owner-filtered rather than silently leaking the cross-provider candidate set. This is exactly the gap the original sweep shipped.

PR #143, opened 2026-06-19
feat(calendar): list a provider's events via calendarEvents userId filter

hugobessa:
## Summary
Adds the server-side foundation for letting the Scheduler/Calendar screen list a single
provider's events in one `calendarEvents` call, instead of resolving that user's calendars
client-side and querying per calendar.

This PR collects the whole plan on one branch (modular commits, one PR):

- **Phase 1 (this commit set):** new `CalendarEventService.get_calendar_events_expanded_for_calendars(calendar_ids, start, end, optimize_queryset=None)` (+ a thin `CalendarService` facade). It expands recurring events across a set of calendars in one pass, org-scoped, and dedups occurrences so an event reachable through multiple owned calendars appears once.
- **Phase 2 (follows on this branch):** wire an optional `userId` argument into the public_api `calendarEvents` query that resolves the user's owned calendars (via `CalendarOwnership`), applies org + scoped-token boundaries, intersects with `calendarId` when both are given, and calls the Phase 1 method.

## Plan reference
`ai-plans/2026-06-18-CALENDAR_EVENTS_USER_FILTER_IMPLEMENTATION_PLAN.md` (see Goals, Guiding Decisions, Phased Rollout).

## Key decisions
- Multi-calendar expansion lives in the service layer (not a resolver loop) so recurrence + dedup logic stays in one place.
- Dedup preserves distinct generated occurrences: a recurring master is single-calendar, so its expanded occurrences (which carry no pk) are never cross-calendar duplicates and are kept unconditionally; only persisted rows and bundle events are collapsed by id. This avoids wrongly dropping two different 9am series owned by the same provider.
- Org guard (`organization_id` filter) is applied on every path so a caller cannot cross tenants by passing foreign calendar ids.
- No feature flag — `userId` will be an additive optional argument; omitting it leaves existing callers byte-for-byte unchanged.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_get_calendar_events_expanded_for_calendars.py -vs`
- `docker compose run --rm api uv run pytest -n auto`
- `docker compose run --rm api uv run python manage.py check --deploy`


calendar_integration/services/calendar_event_service.py
```diff
@@ -1150,6 +1150,131 @@ def get_calendar_events_expanded(
...
+        """
+        if not is_initialized_or_authenticated_calendar_service(
+            cast("BaseCalendarService", self._context)
+        ):
+            raise PermissionDenied("Calendar service is not initialized.")
+
+        # Materialise to a set so we can cheaply check emptiness and avoid repeated
+        # iteration over a generator.
+        id_set: set[int] = set(calendar_ids) if calendar_ids is not None else set()
+        if not id_set:
```
hugobessa, line 1196: Empty/None `calendar_ids` short-circuits to `[]` before any query — the resolver (Phase 2) returns this directly when a user owns no calendars in the org (or a scoped token resolves to an empty set), so we never issue a pointless `calendar_fk__in=()` query. Covered by test_empty_calendar_ids_returns_empty_without_query.

calendar_integration/services/calendar_event_service.py
```diff
@@ -1150,6 +1150,131 @@ def get_calendar_events_expanded(
...
+
+        org_id = self._context.organization.id
+
+        base_qs = (
+            CalendarEvent.objects.annotate_recurring_occurrences_on_date_range(start_date, end_date)
+            .select_related("recurrence_rule")
+            .filter(
+                parent_recurring_object__isnull=True,  # Master events only
+                organization_id=org_id,
+                calendar_fk__in=id_set,
```
hugobessa, line 1207: Org guard is mandatory on every path: filtering `organization_id` here (not just `calendar_fk__in`) means even if a caller passes calendar ids from another tenant, nothing leaks. Phase 2's resolver also pre-filters owned ids by org + scoped token, but this is the defensive backstop.

calendar_integration/services/calendar_event_service.py
```diff
@@ -1150,6 +1150,131 @@ def get_calendar_events_expanded(
...
+        :param start_date: Start of the date range (inclusive).
+        :param end_date: End of the date range (inclusive).
+        :param optimize_queryset: Optional callable applied to the recurring-master
+            queryset so prefetched relations are inherited by generated occurrences.
+        :return: Deduplicated, sorted list of event instances in the range.
+        """
+        if not is_initialized_or_authenticated_calendar_service(
+            cast("BaseCalendarService", self._context)
+        ):
+            raise PermissionDenied("Calendar service is not initialized.")
```
hugobessa, line 1191: Raises a meaningful `PermissionDenied` instead of a bare `raise` (which would throw `RuntimeError: No active exception to re-raise`). The sibling single-calendar method still has the old bare-raise style; left untouched here to avoid scope creep, but the new method gets the correct behavior.

calendar_integration/services/calendar_event_service.py
```diff
@@ -1150,6 +1150,131 @@ def get_calendar_events_expanded(
...
+            elif event.id is not None:
+                # Persisted row — unique by id; collapse repeats reached via the union.
+                if event.id not in seen_event_ids:
+                    seen_event_ids.add(event.id)
+                    unique_events.append(event)
+            else:
+                # Generated recurring occurrence (pk=None) — distinct by construction;
+                # never collapse, so two series producing occurrences at the same
+                # start_time both survive.
+                unique_events.append(event)
```
hugobessa, line 1273: Dedup is identity-aware, and this is the subtle part. Generated recurring occurrences are in-memory copies with `pk=None`, so keying them by `(id, start_time)` would collapse two DISTINCT series on two owned calendars that share a start time. A recurring master belongs to a single calendar, so its occurrences are produced exactly once and can never be a cross-calendar duplicate — hence the `else` branch keeps every generated occurrence, while only persisted rows (`event.id is not None`) and bundle events are deduped by id. Regression: test_distinct_recurring_series_same_start_time_not_collapsed.

calendar_integration/tests/services/test_get_calendar_events_expanded_for_calendars.py
```diff
@@ -0,0 +1,554 @@
...
+
+
+# ---------------------------------------------------------------------------
+# (g) two DISTINCT recurring series on two calendars producing occurrences at the
+#     SAME start_time → BOTH series survive (regression: dedup collision)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_distinct_recurring_series_same_start_time_not_collapsed(
```
hugobessa, line 423: This is the regression guard for the dedup-collision bug found in review: two distinct daily masters (one per owned calendar) firing at the same start_time must both survive. Asserts on a Counter of (title, start_time), not a set of start_times, so a dropped or duplicated occurrence can't hide.

calendar_integration/services/calendar_service.py
```diff
@@ -1006,6 +1006,24 @@ def get_calendar_events_expanded(
             calendar, start_date, end_date, optimize_queryset
         )
 
+    def get_calendar_events_expanded_for_calendars(
```
hugobessa, line 1009: Thin facade mirroring the existing `get_calendar_events_expanded` delegation — no logic here on purpose; all expansion/dedup behavior lives in CalendarEventService so the facade stays a stable public entrypoint for the resolver.

PR #145, opened 2026-06-19
feat(audit): injectable audit-trail module with repository-backed read-only admin

hugobessa:
## Summary
Adds a self-contained, injectable `audit` Django app that records actions taken by
**memberships** (`OrganizationMembership`), **system users** (`public_api.SystemUser`),
**single-use codes** (`calendar_integration.CalendarManagementToken`), and the **system**.

The module is built around a backend-agnostic `AuditRepository` interface (append + read
only — no update, no delete) with a first `DjangoORMAuditRepository` implementation, a
DI-injected `AuditService.record(...)` that other modules call explicitly (async
persistence via Celery), and a **read-only, repository-backed Django admin** (filters,
search, detail view, CSV export) that works against any repository implementation.

This PR collects the whole plan on one branch (modular commits, one PR). Phases land as
atomic commits in plan order:

- **Phase 0:** scaffold the `audit` app (this commit set).
- **Phase 1:** enums (`AuditAction`, `AuditActorType`), portable DTOs (`ActorSnapshot`, `SubjectRef`, `AuditRecordData`, `AuditRecord`, `AuditQuery`, `AuditPage`), and the `AuditRepository` ABC.
- **Phase 2:** `Audit` model + `AuditAffectedMembership` through table + migration.
- **Phase 3:** `DjangoORMAuditRepository` (`add` / `get` / `query` with filters, search, pagination, count).
- **Phase 4:** DI wiring + `compute_diff(before, after)` helper.
- **Phase 5:** `AuditService.record(...)` with synchronous actor-snapshotting + Celery persistence task.
- **Phases 6–9:** repository-backed read-only admin — list+filters, search, detail view, CSV export.

## Plan reference
`ai-plans/2026-06-19-AUDIT_TRAIL_IMPLEMENTATION_PLAN.md` (see Goals, Guiding Decisions, Phased Rollout).

## Key decisions
- **Backend-swappable**: `subject` and `actor` use soft refs (type label + id), `affected_memberships` is exchanged as a plain id list across the interface (the ORM through table is an implementation detail). Lets a future non-ORM backend slot in without touching callers.
- **Async + snapshot-at-emit**: `record(...)` captures mutable actor context (role, scopes, scoped-to-membership) synchronously, then a Celery task persists it — so a later membership/system-user change can't corrupt the historical record.
- **Org-scoped**: `Audit` is an `OrganizationModel`; the admin read path uses the unscoped manager (staff context) with `organization` as a filter.
- **No feature flag**: purely additive new surface (new app, tables, admin, DI providers); the repo has no flag system and no existing path is altered.

## Test plan
- `docker compose run --rm api uv run pytest audit/tests/ -n auto`
- `docker compose run --rm api uv run pytest -n auto`
- `docker compose run --rm api uv run python manage.py check --deploy`
- `docker compose run --rm api uv run python manage.py makemigrations --check --dry-run`


audit/repositories.py
```diff
@@ -0,0 +1,59 @@
...
+
+import abc
+
+from audit.types import AuditPage, AuditQuery, AuditRecord, AuditRecordData
+
+
+class AuditRepository(abc.ABC):
+    """Backend-agnostic interface for audit record storage.
+
+    Read + append only. No update, no delete.
```
hugobessa, line 40: Append + read only is a deliberate Guiding Decision: no update()/delete() on the interface. An audit trail must be immutable from the app's perspective. test_repository_interface.py asserts their absence so a future backend can't quietly reintroduce them.

audit/types.py
```diff
@@ -0,0 +1,93 @@
...
+
+from dataclasses import dataclass, field
+from datetime import datetime
+
+
+@dataclass(frozen=True)
+class ActorSnapshot:
+    """Snapshot of actor context at the time the audit record was emitted.
+
+    Captures mutable actor state synchronously (in the main request) so the
```
hugobessa, line 11: Snapshot-at-emit is the core async-correctness guarantee. Persistence is deferred to Celery (Phase 5), so mutable actor context (role, scopes, scoped-to-membership) is captured synchronously here. The worker never re-reads the membership/system-user, which could have changed or been deleted by the time the task runs.

audit/types.py
```diff
@@ -0,0 +1,93 @@
...
+    actor_role: str | None = None
+    system_user_scopes: list[str] | None = None
+    system_user_scoped_to_membership: int | None = None
+
+
+@dataclass(frozen=True)
+class SubjectRef:
+    """Soft reference to the subject of an audited action.
+
+    Portable across any backend; survives row deletion; no ORM coupling.
```
hugobessa, line 29: Soft reference (type label + string id) instead of a Django GenericForeignKey: keeps the DTO free of ORM/contenttypes coupling so a non-ORM backend stores it as-is, and lets the record survive deletion of the subject row. String subject_id keeps it PK-shape-agnostic.

audit/types.py
```diff
@@ -0,0 +1,93 @@
...
+    subject_type: str
+    subject_id: str
+    subject_label: str | None = None
+
+
+@dataclass(frozen=True)
+class AuditRecordData:
+    """Portable audit record data, passed in Celery payload for persistence.
+
+    This is what AuditService.record() constructs and enqueues to be persisted.
```
hugobessa, line 41: AuditRecordData is the exact payload record() enqueues to Celery, so it must be JSON-serializable end-to-end. It deliberately carries no datetime (created_at lives only on AuditRecord) and exchanges affected_membership_ids as plain list[int]. The json.dumps/loads round-trip in test_types.py guards this.

audit/constants.py
```diff
@@ -0,0 +1,23 @@
...
+    SYSTEM = "system", "System"
+    MEMBERSHIP = "membership", "Membership"
+    SYSTEM_USER = "system_user", "System user"
+    SINGLE_USE_CODE = "single_use_code", "Single-use code"
+
+
+class AuditAction(models.TextChoices):
+    """Action recorded in audit trail.
+
+    Central, extensible enum. Owning modules add members as they instrument
```
hugobessa, line 16: Central but extensible: owning modules append namespaced members (e.g. "calendar.event.reschedule") as they instrument call sites later. Seeded with generic CREATE/UPDATE/DELETE so the enum + admin are usable on day one.

hugobessa (commented):
Phase 1 — Enums, DTOs, and the AuditRepository interface. Inline notes on the load-bearing contract decisions.

audit/models.py
```diff
@@ -0,0 +1,119 @@
...
+class Audit(OrganizationModel):
+    """Immutable audit record capturing an action taken by an actor on a subject.
+
+    Every field is populated at emit time and never mutated afterwards.
+    """
+
+    # Duplicates BaseModel.created by design: created_at is append-only and
+    # semantically clearer for audit ordering. Prefer created_at over created
+    # when ordering or filtering Audit records.
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
```
hugobessa, line 32: created_at duplicates BaseModel's `created` by design: it is the explicit, append-only contract timestamp the DTO/repository order by. auto_now_add (not auto_now) keeps it immutable; test_created_at_is_not_updated_on_save guards that.

audit/models.py
```diff
@@ -0,0 +1,119 @@
...
+    diff = models.JSONField(null=True, blank=True)
+
+    affected_memberships = models.ManyToManyField(
+        "organizations.OrganizationMembership",
+        through="audit.AuditAffectedMembership",
+        # through_fields must name the concrete FK columns (_fk suffix) because
+        # TenantSafeForeignKey contributes a ForeignObject (virtual) and a real
+        # ForeignKey under <name>_fk; Django's M2M plumbing resolves ambiguity
+        # by looking at ForeignKey fields on the through model.
+        through_fields=("audit_fk", "membership_fk"),
```
hugobessa: through_fields is required because OrganizationForeignKey (TenantSafeForeignKey) contributes TWO fields per relation — a concrete ForeignKey `<name>_fk` and a virtual ForeignObject `<name>` that joins on (`<name>_fk`, organization_id). With two such relations on the through model, Django can't auto-pick the M2M ends, so we name the concrete `_fk` columns explicitly.

audit/models.py
```diff
@@ -0,0 +1,119 @@
...
+        on_delete=models.CASCADE,
+        related_name="audit_affected_links",
+    )
+
+    class Meta:
+        constraints: ClassVar = [
+            # Use the _fk concrete columns (OrganizationForeignKey generates <name>_fk).
+            # organization is included per project convention (lead composite keys with org).
+            # audit_fk already pins the org, so the uniqueness guarantee is equivalent.
+            models.UniqueConstraint(
```
hugobessa, line 122: Indexes and the unique constraint lead with `organization` and reference the concrete `_fk` columns (the ForeignObject is virtual, no backing column). uniq_audit_membership is on (organization, audit_fk, membership_fk) — audit_fk already pins the org, so it is the same guarantee, following the org-leading-keys convention. Cross-org links can be written at the raw column level (project-wide ForeignObject limitation), but scoped reads via the affected_memberships ForeignObject join filter them out, and record() always builds ids from one org context.

audit/models.py
```diff
@@ -0,0 +1,119 @@
+"""Audit models.
+
+`Audit` and `AuditAffectedMembership` are both OrganizationModel subclasses.
+They are append-only by intent — the repository never calls update or delete.
+
+Unscoped read access (Phase 3):
+  OrganizationModel already provides `original_manager = models.Manager()` on
+  every subclass.  Phase 3's DjangoORMAuditRepository and Phase 6's admin
+  should use `Audit.original_manager.all()` (filtered by `organization_id`
```
hugobessa, line 9: Unscoped reads for Phase 3's repository + Phase 6's admin use the established original_manager (OrganizationModel ships it on every subclass). Staff/admin context has no active-membership tenant scope, so the default objects manager (which requires an org) can't be used there; organization becomes an explicit query filter instead.

hugobessa (commented):
Phase 2 — Audit model + AuditAffectedMembership through table + migration. Inline notes on the multi-tenant model decisions.

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+        Phase 4's compute_diff returns None for no-change, so callers should
+        rarely pass {} — but we normalize defensively.
+
+        Returns:
+            The persisted AuditRecord.
+        """
+        # Deferred import to avoid loading Django models before the app registry
+        # is ready (audit/__init__.py is imported at app-load time and triggers
+        # this module; model imports must not execute at that point).
+        from audit.models import Audit, AuditAffectedMembership
```
hugobessa, line 131: Model imports are deferred inside each method, not at module top. audit/__init__.py re-exports AuditRepository from this module and is imported at Django app-load time; a top-level `from audit.models import ...` would run before the app registry is ready and raise AppRegistryNotReady. TYPE_CHECKING covers the annotations.

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+                action=data.action,
+                actor_type=data.actor.actor_type,
+                actor_id=data.actor.actor_id,
+                actor_role=data.actor.actor_role,
+                system_user_scopes=data.actor.system_user_scopes,
+                system_user_scoped_to_membership=data.actor.system_user_scoped_to_membership,
+                subject_type=data.subject.subject_type,
+                subject_id=data.subject.subject_id,
+                subject_label=data.subject.subject_label,
+                diff=(data.diff or None),
```
hugobessa, line 150: Cross-phase contract locked here: diff is always None or a non-empty dict. `data.diff or None` normalizes an empty {} to None at write time so the has_diff filter (diff__isnull) is meaningful. Phase 4's compute_diff returns None for no-change, upholding the same invariant.

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Ordering is whitelisted to _ALLOWED_ORDERING_FIELDS; invalid values fall
+        back to _DEFAULT_ORDERING.
+
+        total is counted on the fully-filtered queryset before pagination so
+        callers always get the complete match count, not the page size.
+        """
+        from audit.models import Audit
+
+        qs = Audit.original_manager.all()
```
hugobessa, line 234: Reads start from original_manager (unscoped) because staff/admin context has no active-membership tenant scope — the default objects manager would raise. organization_id is applied as an explicit filter only when the caller supplies it (admin is intentionally cross-org; Phase 6 exposes organization as a filter facet).

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if q.subject_id is not None:
+            qs = qs.filter(subject_id=q.subject_id)
+
+        if q.affected_membership_id is not None:
+            # Join via the through table's reverse relation. Use distinct() to
+            # avoid duplicate Audit rows when a membership appears multiple times
+            # (though the unique constraint prevents it in practice; safe guard).
+            qs = qs.filter(
+                affected_membership_links__membership_fk_id=q.affected_membership_id
```
hugobessa: affected_membership_id filters via the through-table reverse relation and adds .distinct() so the JOIN can't multiply Audit rows. total = qs.count() is taken after this (distinct) filter and before pagination, so the page total reflects the true distinct match count (covered by test_affected_membership_id_filter_counts_multiple_audits_distinct).

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+        if q.search is not None:
+            term = q.search
+            search_q = (
+                Q(subject_type__icontains=term)
+                | Q(subject_id__icontains=term)
+                | Q(subject_label__icontains=term)
+            )
+            # actor_id is a BigIntegerField — only add the equality filter when
+            # the search term looks like an integer, to avoid a DB type error.
+            if term.isdigit():
```
hugobessa, line 283: actor_id is a BigIntegerField, so the equality clause is only OR-ed into the search when the term is all-digits — a non-numeric search term would otherwise raise a DB type error. The subject_* icontains clauses always apply; the whole Q(OR) group is ANDed with the other filters.

hugobessa (commented):
Phase 3 — DjangoORMAuditRepository (add / get / query). Inline notes on the read/write contract and query translation.

audit/repositories.py
```diff
@@ -0,0 +1,342 @@
...
+                search_q |= Q(actor_id=int(term))
+            qs = qs.filter(search_q)
+
+        # --- total (before pagination) ---
+        total = qs.count()
+
+        # --- ordering ---
+        safe_ordering = ordering if ordering in _ALLOWED_ORDERING_FIELDS else _DEFAULT_ORDERING
```
hugobessa, line 291: Ordering is whitelisted (created_at/-created_at) and anything else falls back to -created_at, so callers can't inject arbitrary order_by field names. Phase 6 admin must extend the whitelist if it wants to order by action/actor_type/etc.

audit/diff.py
```diff
@@ -0,0 +1,46 @@
...
+        after_value = after.get(key)
+        if before_value != after_value:
+            result[key] = {"old": before_value, "new": after_value}
+
+    # Check all keys in after: added (keys not in before).
+    for key, after_value in after.items():
+        if key not in before:
+            result[key] = {"old": None, "new": after_value}
+
+    return result if result else None
```
hugobessa, line 46: Returns None (not {}) when nothing changed, upholding the diff invariant locked in Phase 3: diff is always None or a non-empty dict. This keeps has_diff (diff__isnull) meaningful. Added keys use old=None, removed keys use new=None; a None value absent from `after` is treated as no-change (None == absent).

di_core/containers.py
```diff
@@ -42,6 +43,8 @@
 class AppContainer(containers.DeclarativeContainer):
     config = providers.Configuration()
 
+    audit_repository = providers.Singleton(DjangoORMAuditRepository)
```
hugobessa, line 47: Singleton (not Factory) because DjangoORMAuditRepository is stateless — no per-request state, so one shared instance is safe and avoids re-instantiation. audit_service is intentionally NOT wired here; it's added in Phase 5 where AuditService is defined (referencing it now would import a nonexistent symbol and break container load, which check --deploy exercises).

hugobessa (commented):
Phase 4 — compute_diff helper + DI wiring of audit_repository.

audit/services.py
```diff
@@ -0,0 +1,197 @@
...
+                persist_audit_record.delay(payload)
+            except Exception:
+                logger.exception(
+                    "Failed to enqueue audit record for action %r on organization %s. "
+                    "The record will not be persisted.",
+                    action,
+                    organization_id,
+                )
+
+        transaction.on_commit(_enqueue)
```
hugobessa, line 192: Dispatch is wrapped in transaction.on_commit (not a bare .delay()): record() runs from request/resolver paths under ATOMIC_REQUESTS=True, so enqueuing only after the surrounding transaction commits prevents persisting audits for actions that later roll back, and guarantees the worker sees committed state. The .delay() lives inside _enqueue() with its own try/except so a broker failure at commit time is logged and swallowed — the business action is never broken.

audit/services.py
```diff
@@ -0,0 +1,197 @@
...
+
+        Args:
+            system_user: A public_api.SystemUser instance.
+
+        Returns:
+            An ActorSnapshot with actor_type=SYSTEM_USER, scopes list, and
+            scoped_to_membership from the FK.
+        """
+        scopes = [
+            ra.resource_name
```
hugobessa, line 95: Snapshot-at-emit: available_resources.all() is forced to a list HERE (synchronously), so the system-user scopes captured are the values at action time. If we stored a lazy queryset it would re-evaluate in the worker and capture changed/deleted state. test_persisted_role_is_snapshot_not_current proves the analogous guarantee for membership role (build snapshot → mutate role → persist → asserts the OLD role).

audit/tasks.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Args:
+        payload: A JSON-safe dict produced by dataclasses.asdict(AuditRecordData).
+    """
+    # Resolve repository from the DI container at runtime.
+    # di_core.containers.container is set in di_core/apps.py's ready(), which
+    # always runs before any task executes (eager or deferred).
+    from di_core import containers
+
+    di_container = containers.container
```
hugobessa: The task resolves the repository from di_core.containers.container at runtime (with a None-guard) rather than via @inject. @inject decorates at import time, but audit.tasks is imported through the container's own import chain (containers → audit.services → audit.tasks) before container.wire() runs in apps.ready(), so @inject would silently no-op. The module-level `container` is set by apps.ready() before any task executes.

audit/tasks.py
```diff
@@ -0,0 +1,105 @@
...
+    except Exception:
+        logger.exception(
+            "persist_audit_record: malformed payload, cannot reconstruct AuditRecordData. "
+            "Payload: %r",
+            payload,
+        )
+        return
+
+    try:
+        repository = di_container.audit_repository()
```
hugobessa: Fire-and-forget: repository.add() failures (and payload reconstruction failures above) are logged and swallowed so a bad payload or transient DB error never crashes the worker. Per the plan's async write-path decision, a dropped audit on failure is the accepted tradeoff; acks_late means a retry may create a duplicate row, which is acceptable noise for an append-only log.

hugobessa (commented):
Phase 5 — AuditService.record() + persist_audit_record Celery task. Notes on async correctness.

audit/admin.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def has_delete_permission(self, request: HttpRequest, obj: Any = None) -> bool:
+        """Deny all delete permissions — audit records are immutable."""
+        return False
+
+    # ------------------------------------------------------------------ #
+    # Repository-backed changelist                                        #
+    # ------------------------------------------------------------------ #
+
+    def changelist_view(
```
hugobessa, line 226: changelist_view is fully overridden — it builds an AuditQuery from GET, calls container.audit_repository().query(...) and renders a custom template, bypassing Django's ORM ChangeList entirely. This is what makes the admin backend-agnostic (the plan's 'Admin backend-agnosticism' decision): swap the repository and the list still renders. test_get_params_translated_to_audit_query + the stub-repository override prove it.

audit/admin.py
```diff
@@ -0,0 +1,279 @@
...
+        This queryset is only used by Django's internal view plumbing, NOT by the
+        ``changelist_view`` override (which reads exclusively from the repository).
+        """
+        return Audit.original_manager.all()
+
+    # ------------------------------------------------------------------ #
+    # Read-only enforcement                                               #
+    # ------------------------------------------------------------------ #
+
+    def has_add_permission(self, request: HttpRequest) -> bool:
```
hugobessa, line 209: Read-only: add/change/delete permissions all return False, so POSTs to the add/change/delete admin URLs raise PermissionDenied (tested). The audit log is append-only — the only write path is AuditService.record(). Django 6 still shows a read-only per-object view (has_view_permission default True) with no Save button; Phase 8 replaces it with a repository-driven detail.

audit/admin.py
```diff
@@ -0,0 +1,279 @@
...
+        change/delete view infrastructure calls ``get_queryset`` before checking
+        permissions, which would crash with a 500.  Using ``original_manager``
+        (the unscoped manager added by ``OrganizationModel``) avoids that crash
+        while still honouring the fact that all three permission methods return
+        ``False`` (so no actual change or delete will ever occur).
+
+        This queryset is only used by Django's internal view plumbing, NOT by the
+        ``changelist_view`` override (which reads exclusively from the repository).
+        """
+        return Audit.original_manager.all()
```
hugobessa, line 203: get_queryset uses the unscoped original_manager ONLY for Django's internal change/delete view plumbing (which calls get_object before permission checks; the tenant-scoped objects manager would raise without an org). The changelist itself never uses this — it reads exclusively from the repository.

audit/admin.py
```diff
@@ -0,0 +1,279 @@
...
+
+        The list branch applies ``or None`` so that an empty submitted value
+        (e.g. ``actor_type=''`` → ``['']``) normalises to ``None`` rather
+        than ``''``, which would cause the repository to filter on an empty
+        string instead of skipping the filter entirely.
+        """
+        if v is None:
+            return None
+        if isinstance(v, list):
+            return (v[0] or None) if v else None
```
hugobessa, line 144: Empty-filter normalization: a form submitting `actor_type=` arrives as {'actor_type': ['']}; without `(v[0] or None)` the empty string would become a real AuditQuery.actor_type='' filter and return zero rows. This normalizes blank submitted values to None (no filter) for every dropdown/date/id field.

hugobessa (commented):
Phase 6 — Repository-backed, read-only admin changelist (list + filters).

audit/admin.py
```diff
@@ -0,0 +1,285 @@
...
+        return v or None
+
+    action = _first(params.get("action"))
+    actor_type = _first(params.get("actor_type"))
+    created_after = _parse_datetime(_first(params.get("created_after")))
+    created_before = _parse_datetime(_first(params.get("created_before")))
+    has_diff = _parse_has_diff(_first(params.get("has_diff")))
+    organization_id = _parse_int(_first(params.get("organization_id")))
+    search = _first(params.get("search"))
```
hugobessa, line 153: Pure wiring, no new query logic: `search` maps to AuditQuery.search (the repository ORs subject_type/subject_id/subject_label icontains + numeric actor_id) and `affected_membership_id` maps to the repository's through-table join filter. Both go into active_filters, so base_querystring carries them across pagination; empty/invalid values normalize to None (no filter) via the shared _first/_parse_int helpers.

hugobessa (commented):
Phase 7 — Admin search wiring (search + affected_membership_id).

di_core/containers.py
```diff
@@ -42,6 +44,9 @@
...
+    audit_service = providers.Factory(AuditService)
```
hugobessa, line 48: providers.Factory(AuditService) intentionally passes NO explicit repository — @inject + Provide["audit_repository"] on AuditService.__init__ resolves it. This matches webhook_service = providers.Factory(WebhookService) and eliminates the earlier DIWiringWarning (which fired precisely because an explicitly-passed dep made @inject redundant). container.audit_repository.override(stub) still affects the injected repository because @inject resolves through this same wired container (stub-override admin tests prove it).

audit/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+
+if TYPE_CHECKING:
+    from audit.repositories import AuditRepository
+
+
+logger = logging.getLogger(__name__)
+
+
+@app.task
+@inject
```
hugobessa, line 39: Now the canonical Celery-task DI pattern from webhooks/tasks.py: @app.task on top, @inject below, repository as an Annotated[..., Provide[...]] = None keyword arg with a None guard. Runtime container resolution removed. Root cause of why @inject silently no-op'd before: this module had `from __future__ import annotations` (PEP 563), which stringifies all annotations so dependency-injector never sees the Provide marker — removed here (the canonical modules deliberately omit that import).

audit/services.py
```diff
@@ -0,0 +1,192 @@
...
+
+class AuditService:
+    """Service that records audit trail entries asynchronously via Celery.
+
+    Actor context is captured synchronously at call time and serialized into
+    the Celery task payload so the worker never re-reads mutable state that
+    may have changed or been deleted by the time it runs.
+    """
+
+    @inject
```
hugobessa, line 49: AuditService.__init__ injects repository via @inject + Annotated[AuditRepository, Provide["audit_repository"]] — the organizations/services.py style. `from __future__ import annotations` removed and AuditRepository imported at module level so the Provide marker is a live object at decoration time.

audit/admin.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def has_delete_permission(self, request: HttpRequest, obj: Any = None) -> bool:
+        """Deny all delete permissions — audit records are immutable."""
+        return False
+
+    # ------------------------------------------------------------------ #
+    # Repository-backed changelist                                        #
+    # ------------------------------------------------------------------ #
+
+    @inject
```
hugobessa, line 225: changelist_view injects the repository via @inject too (no more runtime container lookup). @inject works on this bound method because the audit package is wired in di_core/apps.py ready(); Django's admin dispatch calls changelist_view normally and @inject fills the repository kw-arg. The None guard preserves the empty-page fallback.

hugobessa (commented):
Amendment (2026-06-19) — rework DI to method-argument injection (@inject + Provide), matching the project's webhooks/tasks.py + organizations/services.py convention. Applied as a forward corrective commit (no history rewrite). See the plan's Amendments log + the 'DI injection style' Guiding Decision.

hugobessa (commented):
Phase 8 — read-only, repository-backed admin detail view.

audit/admin.py
```diff
@@ -0,0 +1,390 @@
...
+            request,
+            self.change_list_template,
+            context,
+        )
+
+    # ------------------------------------------------------------------ #
+    # URL routing                                                         #
+    # ------------------------------------------------------------------ #
+
+    def get_urls(self):  # type: ignore[no-untyped-def]
```
hugobessa, line 335: Detail view registered via get_urls (custom `<int:audit_id>/view/`, name audit_audit_detail, wrapped in admin_site.admin_view for staff auth) rather than Django's ORM change_view — so the per-object page is repository-driven and read-only, not the ORM change form. Changelist rows link here.

audit/admin.py
```diff
@@ -0,0 +1,390 @@
...
+                name="audit_audit_detail",
+            ),
+        ]
+        return custom_urls + urls
+
+    # ------------------------------------------------------------------ #
+    # Repository-backed detail view                                       #
+    # ------------------------------------------------------------------ #
+
+    @inject
```
hugobessa, line 362: detail_view injects the repository via @inject (matching the 2026-06-19 DI amendment and the changelist_view), calls repository.get(audit_id), and raises Http404 on miss (or when repository is None). No runtime container resolution. Read-only: GET-only rendering, no form/save/delete; the stub-repo override test proves the page renders from repository.get, not the ORM.

audit/admin.py
```diff
@@ -0,0 +1,575 @@
...
+                    diff_json,
+                ]
+                yield writer.writerow(row)
+
+            offset += chunk_size
+            if offset >= page.total:
+                break
+
+    @inject
+    def export_view(
```
hugobessa, line 532: export_view: GET-only (405 on other methods), repository injected via @inject (matches changelist/detail), builds the SAME AuditQuery as the changelist via _build_audit_query so the export honors the active filters + search, and streams via StreamingHttpResponse(text/csv; charset=utf-8). Read-only — no mutation path.

audit/admin.py
```diff
@@ -0,0 +1,575 @@
...
+            "subject_label",
+            "affected_membership_ids",
+            "diff",
+        ]
+        yield writer.writerow(header)
+
+        # --- Paginate through all records ---
+        offset = 0
+        while True:
+            page = repository.query(q, offset=offset, limit=chunk_size, ordering="-created_at")
```
hugobessa, line 491: _csv_row_generator pages through repository.query in chunk_size batches and yields rows via a single csv.writer over an _Echo pseudo-buffer (the canonical Django streaming pattern), so memory stays bounded regardless of result size. test_large_result_set_no_truncation seeds 1500 rows (>1 chunk) and asserts both the full count AND id-uniqueness — proving no truncation and no chunk-boundary duplication.

audit/admin.py
```diff
@@ -0,0 +1,575 @@
...
+    - ``YYYY-MM-DDTHH:MM:SSZ`` or ``YYYY-MM-DDTHH:MM:SS+00:00`` (offset-aware)
+
+    Falls back to explicit ``strptime`` patterns for extra safety, then returns
+    ``None`` on any unparseable input so the filter is silently skipped.
+    """
+    if not value:
+        return None
+    # Python 3.11+ fromisoformat handles Z suffix and offsets directly.
+    try:
+        result = datetime.fromisoformat(value)
```
hugobessa, line 93: Review-found bug fix (shared by the Phase 6 changelist + this export): _parse_datetime previously only accepted YYYY-MM-DDTHH:MM / YYYY-MM-DD, so any ISO value with seconds/microseconds/Z/offset silently returned None and the date filter was skipped. Now uses datetime.fromisoformat() (Python 3.13 handles Z + offsets) with a strptime fallback and forces UTC-awareness when naive.

hugobessa (commented):
Phase 9 (final) — filtered, streamed CSV export. Plus a shared-parser bug fix found in review.

hugobessa:
✅ **All 10 phases complete** (audit-trail plan), plus the DI method-argument-injection amendment.

Final full suite: **2870 passed**, `manage.py check --deploy` clean (no DIWiringWarning), `makemigrations --check` clean.

Phases: 0 scaffold · 1 contract (enums/DTOs/AuditRepository ABC) · 2 model+migration · 3 DjangoORMAuditRepository · 4 DI wiring + compute_diff · 5 AuditService + Celery task (on_commit, snapshot-at-emit) · 6 admin list+filters · 7 admin search · 8 admin detail · 9 admin CSV export. Amendment: DI reworked to `@inject` + `Provide` (webhooks/tasks.py + organizations/services.py convention).

Each phase was adversarially reviewed (reviewer + fixer loop). Inline review threads above annotate the load-bearing decisions.

audit/models.py
```diff
@@ -0,0 +1,136 @@
...
+        related_name="affected_membership_links",
+    )
+    # OrganizationMembershipForeignKey contributes:
+    #   - membership_user_id (BigIntegerField) — concrete column, the org-scoped user_id
+    #   - membership (ForeignObject) — joins (organization_id, membership_user_id) →
+    #     OrganizationMembership(organization_id, user_id)
+    # on_delete=DO_NOTHING: no DB constraint is added (audit is append-only and
+    # must survive membership deletion). PROTECT integrity from the raw-SQL composite
+    # FK used by CalendarOwnership is intentionally OMITTED here.
+    membership = OrganizationMembershipForeignKey(
```
hugobessa, line 110: AuditAffectedMembership.membership now uses OrganizationMembershipForeignKey → concrete `membership_user_id` + a `membership` ForeignObject joining (organization_id, membership_user_id) → OrganizationMembership(organization_id, user_id), matching CalendarOwnership. on_delete=DO_NOTHING and — unlike CalendarOwnership — we deliberately do NOT add the per-table raw-SQL composite PROTECT FK: audit is append-only and must survive membership deletion (a since-deleted membership simply reads back as None via the ForeignObject; the stored user_id is preserved). null=False (every link names a membership), so the unique is a plain (audit_fk, membership_user_id).

audit/models.py
```diff
@@ -0,0 +1,136 @@
...
+    diff = models.JSONField(null=True, blank=True)
+
+    affected_memberships = models.ManyToManyField(
+        "organizations.OrganizationMembership",
+        through="audit.AuditAffectedMembership",
+        # through_fields references the ForeignObject descriptor name "membership"
+        # (the name given to OrganizationMembershipForeignKey) per the convention
+        # established by CalendarOwnership — the M2M target end is the ForeignObject
+        # descriptor, NOT a _fk column.
+        through_fields=("audit_fk", "membership"),
```
hugobessa, line 72: M2M through_fields second end is the ForeignObject descriptor name `"membership"` (not a `_fk` column) — the required form when the membership end is an OrganizationMembershipForeignKey (per the CalendarOwnership precedent).

audit/services.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Args:
+            membership: An OrganizationMembership instance.
+
+        Returns:
+            An ActorSnapshot with actor_type=MEMBERSHIP and actor_role set.
+        """
+        return ActorSnapshot(
+            actor_type=AuditActorType.MEMBERSHIP,
+            actor_id=membership.user_id,  # type: ignore[attr-defined]
```
hugobessa, line 75: Consequence of main's composite-PK migration: OrganizationMembership no longer has a scalar `id`, so the MEMBERSHIP actor is identified by its org-scoped `user_id` (the audit row's organization_id disambiguates). This is now consistent across the module — affected_membership_ids, AuditQuery.affected_membership_id, and system_user_scoped_to_membership (line 102, reading scoped_to_membership_user_id) are all org-scoped user_ids. The DTO docstrings in types.py document this.

hugobessa (commented):
Follow-up (main drifted): adopt OrganizationMembershipForeignKey for membership references in the audit module. Branch had origin/main merged in; main gave OrganizationMembership a composite PK (no `id`) and switched membership references to the (organization_id, user_id) composite convention. Mirrors the CalendarOwnership precedent. Existing 0001_initial fixed in place (no data yet).

PR #146, opened 2026-06-20
feat(common): add OrganizationMembershipForeignKey field type

hugobessa:
## Summary
Phase 0 (scaffolding) of the **Membership-Scoped Calendar References** plan. Adds a reusable
`OrganizationMembershipForeignKey` field type to `common/fields.py`. No model adopts it yet and no
migration is generated — this phase ships the foundation that Phases 1–6 consume to move calendar
`User` references onto `OrganizationMembership`.

## How
Modelled on the existing `TenantSafeForeignKey` in the same file. `contribute_to_class` injects **two**
Django fields onto the host model:
1. a concrete `<name>_user_id` `BigIntegerField` (the denormalized user id — `User`'s PK is `BigAutoField`), and
2. a non-editable `ForeignObject` named `<name>` joining `(host.organization_id, host.<name>_user_id)` →
   `OrganizationMembership(organization_id, user_id)`.

Django 6 forbids a real `ForeignKey` to a composite-PK model (`OrganizationMembership` gains a composite
PK in Phase 7), so the relation **must** be a `ForeignObject` regardless — and denormalizing `user_id`
onto the host row is what makes resolving the owning user joinless. The empirically-compiled JOIN is:
`INNER JOIN organizations_organizationmembership ON (host.membership_user_id = …user_id AND host.organization_id = …organization_id)`.

## What else
- The `ForeignObject` creates **no** DB constraint; PROTECT delete semantics are enforced by a per-table
  raw-SQL composite FK added in the cutover phases (2/4/6). The `on_delete` kwarg is ORM-bookkeeping only.
- The field deliberately does **not** bake in a single-column index on `membership_user_id`: under
  multi-tenant queries the join always filters `organization_id AND membership_user_id`, so the useful
  index is the tenant-leading composite `(organization_id, <name>_user_id)`, which adopting models declare
  in their own migration (documented in the field docstring).
- No `deconstruct` needed: `makemigrations --check` is clean because only the contributed sub-fields land
  in migration state, not the wrapper (same as `TenantSafeForeignKey`).

## Plan reference
Plan `ai-plans/2026-06-19-MEMBERSHIP_SCOPED_CALENDAR_REFERENCES_IMPLEMENTATION_PLAN.md` — Phase 0
(Data Model Changes 3.1). First phase; branches off `main`.

## Test plan
- `common/tests/test_fields.py` — 5 structural tests (concrete column, `from_fields`/`to_fields`, null
  propagation) + a DB-backed `TestOrganizationMembershipForeignKeyBehavior` (table created at test time
  via `schema_editor`) asserting: the descriptor resolves the membership, `select_related("membership")`
  issues exactly one query, and `filter(membership__role=…)` traverses to `OrganizationMembership`.
- Gate: `pytest -n auto` → 2619 passed; `manage.py check --deploy` clean; `makemigrations --check` → no
  changes (proves the phase adds no migration).

### DIFF
--- a/common/fields.py
+++ b/common/fields.py
@@ -187,6 +187,16 @@ class OrganizationMembershipForeignKey(models.Field):
        - The ``ForeignObject`` below already provides all ORM relationship
          features (``select_related``, reverse-accessor, filter traversal).
 
+       **Required index**: This field deliberately does NOT add a single-column
+       index on ``<name>_user_id``.  Multi-tenant queries always filter
+       ``organization_id = X AND <name>_user_id = Y``, so the useful index is
+       the tenant-leading composite ``(organization_id, <name>_user_id)``.
+       A bare single-column index on ``<name>_user_id`` alone would be mostly
+       wasted.  Adopting models MUST declare a composite index on
+       ``(organization_id, <name>_user_id)`` in the migration that adds this
+       field — alongside the raw-SQL composite FK constraint added in the
+       cutover phase.
+
     2. **ForeignObject descriptor** ``<name>`` — a non-editable ``ForeignObject``
        joining::
 
--- a/common/fields.py
+++ b/common/fields.py
@@ -1,3 +1,33 @@
+"""Custom Django field types for the Vinta Schedule API.
+
+Module rationale — OrganizationMembershipForeignKey
+----------------------------------------------------
+Django 6 forbids a real ``ForeignKey`` (with a DB-level FK constraint) to a
+model whose primary key is composite (``CompositePrimaryKey``).  Even before
+``OrganizationMembership`` gains a composite PK (Phase 7), using a
+``ForeignObject`` for the relationship is the only design that will survive
+that migration without a second round of model rewrites.
+
+Why denormalize ``user_id``?
+    A bare ``ForeignObject`` on ``(organization_id, …)`` requires the *host*
+    row to already carry the second join column.  The host row always has
+    ``organization_id`` (every ``OrganizationModel`` does), but it does NOT
+    natively carry ``user_id``.  Storing a denormalized ``<name>_user_id``
+    column on the host row makes the join instant — no extra look-up needed
+    to resolve the owning user — and keeps the ``ForeignObject`` column
+    mapping simple.
+
+Why no real DB foreign-key constraint here?
+    ``ForeignObject`` creates no DB-level constraint; the ORM join is purely
+    at the Python/SQL level.  PROTECT delete semantics are enforced by a
+    raw-SQL composite ``FOREIGN KEY (organization_id, <name>_user_id)
+    REFERENCES organization_membership(organization_id, user_id) ON DELETE
+    RESTRICT`` constraint added per referencing table in the cutover phases
+    (Phases 2, 4, 6).  This separation is intentional: the ORM field provides
+    ``select_related`` / descriptor / queryset conveniences; the DB constraint
+    provides integrity.
+"""
+
 import uuid
 
 from django.db import models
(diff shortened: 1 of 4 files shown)


common/fields.py
```diff
@@ -131,3 +161,120 @@ def contribute_to_class(self, cls, name):
...
+
+    def contribute_to_class(self, cls, name: str) -> None:  # type: ignore[override]
+        """Inject the concrete ``<name>_user_id`` column and the ForeignObject descriptor."""
+        user_id_field_name = f"{name}_user_id"
+
+        # 1. Concrete column: plain BigIntegerField matching the User PK type
+        #    (DEFAULT_AUTO_FIELD = BigAutoField → 64-bit integer).  No DB FK
+        #    constraint — the composite FK to OrganizationMembership is added
+        #    per table as raw SQL in the cutover phases.
+        user_id_field = models.BigIntegerField(
```
hugobessa, line 254: Concrete column is a plain `BigIntegerField`, not a `ForeignKey(User)`, on purpose: contributing a `ForeignKey` as `"<name>_user_id"` would produce an attname `<name>_user_id_id` (Django appends `_id`), and the integrity constraint we actually want is the composite `(organization_id, <name>_user_id)` → `OrganizationMembership`, added as raw SQL per table in the cutover phases — not a bare `User` FK. `User`'s PK is `BigAutoField`, so `BigIntegerField` is the matching width.

common/fields.py
```diff
@@ -131,3 +161,120 @@ def contribute_to_class(self, cls, name):
...
+        #      - "organization_id": the inherited FK attname on OrganizationModel
+        #    to_fields references the *attnames* on OrganizationMembership:
+        #      - "user_id": attname of the `user` ForeignKey
+        #      - "organization_id": attname of the `organization` ForeignKey
+        #    This mirrors TenantSafeForeignKey's convention of using attnames
+        #    (e.g. "organization_id") rather than field names in to_fields.
+        fo_field = ForeignObject(
+            "organizations.OrganizationMembership",
+            from_fields=[user_id_field_name, "organization_id"],
+            to_fields=["user_id", "organization_id"],
```
hugobessa, line 274: `to_fields=["user_id", "organization_id"]` uses attnames (not field names `user`/`organization`). This mirrors `TenantSafeForeignKey` above, which uses the attname `"organization_id"` in its `to_fields`. Verified to compile against the real model: the ON clause maps `membership_user_id → user_id` and `organization_id → organization_id`.

common/fields.py
```diff
@@ -131,3 +161,120 @@ def contribute_to_class(self, cls, name):
...
+         raw-SQL constraint per table in the cutover phases (Phases 2, 4, 6).
+       - The ``ForeignObject`` below already provides all ORM relationship
+         features (``select_related``, reverse-accessor, filter traversal).
+
+       **Required index**: This field deliberately does NOT add a single-column
+       index on ``<name>_user_id``.  Multi-tenant queries always filter
+       ``organization_id = X AND <name>_user_id = Y``, so the useful index is
+       the tenant-leading composite ``(organization_id, <name>_user_id)``.
+       A bare single-column index on ``<name>_user_id`` alone would be mostly
+       wasted.  Adopting models MUST declare a composite index on
```
hugobessa, line 195: Index responsibility lives here by design. No single-column index is baked onto `membership_user_id` because the tenant-leading composite `(organization_id, <name>_user_id)` is the one the planner can use under multi-tenant filtering. Each adopting model (Phases 1/3/5 expand migrations) must declare that composite index alongside the raw-SQL composite FK — flagged so the hot-table cutovers do not silently ship an unindexed join key.

common/tests/test_fields.py
```diff
@@ -0,0 +1,327 @@
+"""Unit tests for ``common.fields.OrganizationMembershipForeignKey``.
```
hugobessa, line 1: The behavioral test creates the throwaway host table at runtime via `schema_editor` under `transaction=True` (an `OrganizationModel` carries a DEFERRABLE INITIALLY DEFERRED FK to Organization, so the table can only be dropped outside a savepoint). The model class is defined inside the fixture, not at module scope, so Django's `sync_apps` does not try to create it during test-DB init before the `organizations_*` tables exist.

PR #147, opened 2026-06-20
feat(calendar): expand CalendarOwnership with membership reference + backfill

hugobessa:
## Summary
Phase 1 (expand) of the **Membership-Scoped Calendar References** plan. `CalendarOwnership` gains a
`membership` reference (the Phase-0 `OrganizationMembershipForeignKey`) **alongside** its existing `user`
FK, plus a backfill that copies `user_id → membership_user_id` wherever a matching
`OrganizationMembership(user_id, organization_id)` exists. Behaviour is byte-for-byte unchanged: no
read/write site moved, `user` FK kept, `Calendar.users` M2M untouched, no API change, no PROTECT FK yet
(that lands in the Phase 2 cutover).

## How
- `calendar_integration/models.py`: `membership = OrganizationMembershipForeignKey(on_delete=PROTECT,
  related_name="calendar_ownerships", null=True, blank=True)`. Contributes the concrete
  `membership_user_id BigIntegerField(null=True)` + a non-editable `ForeignObject` joining
  `(organization_id, membership_user_id) → OrganizationMembership(organization_id, user_id)`.
  New `Meta.indexes` adds the tenant-leading composite `(organization, membership_user_id)`
  (`calownership_org_member_idx`) — the index responsibility flagged on the field in Phase 0.
- Migration `0022` (schema): nullable column add (no rewrite) + index. Migration `0023` (data,
  `atomic=False`): batched PK-range `UPDATE … WHERE membership_user_id IS NULL AND EXISTS(membership)`,
  idempotent + resumable. Orphans (no membership) stay NULL and are written to a CSV under
  `.vinta-ai-workflows/one-off-runs/` (with a WARNING; per-row log fallback on `OSError`). Clean reverse
  nulls the column.

## What else
- `related_name="calendar_ownerships"` is deliberately shared by `user` (→`User`) and `membership`
  (→`OrganizationMembership`) — different target models, so no Django clash; this lets Phase 2 drop `user`
  and hand the name to `membership` with no churn.
- Backfill helpers extracted to `_0023_backfill_helpers.py` so tests call them directly.

## Plan reference
Plan `ai-plans/2026-06-19-MEMBERSHIP_SCOPED_CALENDAR_REFERENCES_IMPLEMENTATION_PLAN.md` — Phase 1
(Ownership expand). Stacked on phase-0 (PR #146).

## Test plan
- `calendar_integration/tests/test_ownership_expand.py` (12 tests): matched rows get
  `membership_user_id == user_id` + `.membership` descriptor resolves; orphans stay NULL + appear in the
  report; matched rows absent from report; backfill idempotent; CSV report written (monkeypatched
  `LOG_DIR`); `OSError` fallback logs per-row; reverse nulls the column; behaviour-unchanged for
  `filter(user=…)` + `user.calendar_ownerships`; multi-org/multi-user.
- Gate: `pytest -n auto` → 2631 passed; `migrate` forward/reverse/forward clean on the dedicated DB;
  `makemigrations --check` → no changes; `check --deploy` clean.


calendar_integration/models.py
```diff
@@ -235,6 +240,12 @@ class CalendarOwnership(OrganizationModel):
         on_delete=models.CASCADE,
         related_name="calendar_ownerships",
     )
+    membership = OrganizationMembershipForeignKey(
```
hugobessa, line 243: `on_delete=PROTECT` here is ORM-bookkeeping only — a `ForeignObject` creates no DB constraint. Real PROTECT against membership deletion is the raw-SQL composite FK added in the Phase 2 cutover. `null=True` because the backfill leaves orphan `(user, org)` pairs (users removed from an org but still owning a calendar) NULL rather than auto-creating memberships.

calendar_integration/models.py
```diff
@@ -243,6 +254,14 @@ class CalendarOwnership(OrganizationModel):
         ),
     )
 
+    class Meta:
+        indexes: ClassVar = [
```
hugobessa, line 258: Tenant-leading composite index `(organization, membership_user_id)` — the index responsibility flagged on `OrganizationMembershipForeignKey` in Phase 0. A bare single-column index on `membership_user_id` would be mostly wasted because every join also filters `organization_id`.

calendar_integration/migrations/0023_backfill_calendarownership_membership_user_id.py
```diff
@@ -0,0 +1,135 @@
...
+    with connection.cursor() as cursor:
+        cursor.execute(
+            "UPDATE calendar_integration_calendarownership SET membership_user_id = NULL"
+        )
+
+
+class Migration(migrations.Migration):
+    """Backfill CalendarOwnership.membership_user_id (Phase 1 data migration)."""
+
+    atomic = False
```
hugobessa, line 123: `atomic = False` is load-bearing: it makes the batched PK-range backfill genuinely resumable (each batch commits independently, so a mid-run failure does not roll back completed batches) and avoids holding row locks across the whole table in one transaction. This is the pattern the hot-table expand phases (EventAttendance / CalendarManagementToken, Phases 3/5) will reuse.

calendar_integration/migrations/_0023_backfill_helpers.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def backfill_membership_user_id_sql() -> None:
+    """Batch-UPDATE membership_user_id = user_id where a matching membership exists.
+
+    Iterates CalendarOwnership PKs in BATCH_SIZE chunks.  Rows that already have
+    membership_user_id set (non-NULL) are skipped by the WHERE guard, making the
+    function idempotent.
+
+    Intentional cross-organization raw SQL: this is a one-off data-migration backfill
```
hugobessa, line 23: Intentional tenant-scope exception: this raw `cursor.execute` deliberately runs across ALL organizations (no org filter) because it is a one-off cross-org data-migration backfill — the only correct way to populate every tenants rows. AGENTS.md flags `cursor.execute` on tenant tables as a bypass to avoid; the comment marks this as a sanctioned exception.

PR #148, opened 2026-06-20
feat(calendar): route CalendarOwnership through membership (app-layer cutover)

hugobessa:
## Summary
Phase 2a (first half of the Phase 2 cutover) of **Membership-Scoped Calendar References**. Routes every
`CalendarOwnership.user` read/write through `membership` / `membership_user_id`, replaces the
`Calendar.users` M2M with `Calendar.memberships`, and exposes the calendar owner as a membership
`{ user_id, organization_id, role }` in GraphQL + REST. The `user` **column is intentionally kept** this
phase — **Phase 2b** drops it and adds the raw-SQL PROTECT FK.

## How
- **Reads/writes**: `request.user`-keyed ownership lookups resolve the active membership via
  `get_active_organization_membership(request.user)` (or filter by `membership_user_id`); off-request
  services filter `membership_user_id` directly. Creates set `membership_user_id` alongside the still-NOT-NULL
  `user`.
- **M2M**: `Calendar.users` → `Calendar.memberships` (through `CalendarOwnership.membership`, the ForeignObject).
  Migration `0024` is **state-only** (M2M-with-explicit-through manages no table, so it never touches the
  `user` column). Verified by spike: the two-hop `Calendar → CalendarOwnership → OrganizationMembership`
  query compiles correctly.
- **API**: GraphQL `CalendarOwnershipGraphQLType.user` (UserGraphQLType) → `membership`
  (`OwnershipMembershipGraphQLType`); REST `CalendarOwnershipSerializer` exposes `membership`. Membership has
  no scalar id (composite PK arrives in Phase 7), so the shape is `{ user_id, organization_id, role }` per
  the plan's Open Question #1 default. `schema.yml` regenerated (unchanged — the ownership serializer isn't
  bound to a REST endpoint; the GraphQL `owners` field is the consumer-visible change).

## Orphan semantics
Ownerships whose `(user, org)` has no membership (`membership_user_id IS NULL`, from Phase 1) are
intentionally excluded from membership-based reads / the M2M, and orphan-only calendars return HTTP 400 on
owner-resolution side-effect paths (`admin_sync`, event `transfer`) — covered by new tests. This is the
intended end state (an ex-member shouldn't own/drive a calendar).

## Review hardening (post-review fixes in this branch)
- Write-adapter owner resolution no longer dereferences `ownership.membership.user` (which would crash on a
  stale `membership_user_id` — possible until the Phase 2b PROTECT FK); it resolves the user from the
  scalar `membership_user_id`.
- Sync `update_or_create` keeps keying on the stable `user` FK during the dual-column window (avoids
  duplicate/`MultipleObjectsReturned` on a nullable `membership_user_id` key); Phase 2b switches the key +
  adds a unique constraint.

## Plan reference
Plan `…2026-06-19-MEMBERSHIP_SCOPED_CALENDAR_REFERENCES_IMPLEMENTATION_PLAN.md` — Phase 2 (app-layer half).
Stacked on phase-1 (PR #147).

## Test plan
- `calendar_integration/tests/test_ownership_cutover.py` + updates across calendar/organizations/public_api
  tests: ownership create sets membership; owner-permission + default-calendar resolve via membership;
  `Calendar.memberships` returns owners; API returns the membership shape; orphan exclusion; orphan-only
  `admin_sync`/`transfer` → 400.
- Gate: `pytest -n auto` → 2642 passed; `makemigrations --check` clean; `check --deploy` clean.


calendar_integration/models.py
```diff
@@ -141,12 +141,14 @@ class Calendar(OrganizationModel):
         ),
     )
 
-    users: "models.ManyToManyField[User, CalendarOwnership]" = models.ManyToManyField(
-        User,
-        related_name="calendars",
-        through="CalendarOwnership",
-        through_fields=("calendar_fk", "user"),
-        blank=True,
+    memberships: "models.ManyToManyField[OrganizationMembership, CalendarOwnership]" = (
```
hugobessa, line 144: Calendar.users → Calendar.memberships through the membership ForeignObject. The plan flagged M2M-through-a-ForeignObject as unproven Django; an orchestrator spike confirmed it compiles a correct two-hop JOIN (Calendar → CalendarOwnership → OrganizationMembership on (organization_id, membership_user_id)), so no fallback to explicit querysets was needed.

calendar_integration/migrations/0024_remove_calendar_users_calendar_memberships.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('calendar_integration', '0023_backfill_calendarownership_membership_user_id'),
+        ('organizations', '0011_organizationbranding'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
hugobessa, line 14: State-only migration: an M2M with an explicit `through` model manages no database table, so RemoveField(users) + AddField(memberships) change only Django model state — they do NOT touch the `user` column on CalendarOwnership (that drop is Phase 2b).

calendar_integration/graphql.py
```diff
@@ -133,17 +133,45 @@ def _scoped_calendar_list(
     return [c for c in calendars if c.id in allowed_ids]
 
 
+@strawberry.type
+class OwnershipMembershipGraphQLType:
```
hugobessa, line 137: Consumer-visible breaking change: CalendarOwnershipGraphQLType.user (UserGraphQLType) is replaced by membership. A membership has no scalar id (composite PK lands in Phase 7), so the external shape is { user_id, organization_id, role } — the plan Open Question #1 default. Resolver returns None for orphan ownerships.

public_api/scoping.py
```diff
@@ -21,7 +21,7 @@ def scoped_calendar_ids(system_user: SystemUser, organization: Organization) ->
         return None
     return set(
         Calendar.objects.filter_by_organization(organization.id)
-        .filter(ownerships__user__organization_memberships=system_user.scoped_to_membership_fk_id)
+        .filter(ownerships__membership__id=system_user.scoped_to_membership_fk_id)
```
hugobessa, line 24: Security-sensitive (per-owner-scoped public API tokens). The new `ownerships__membership__id=` filter is equal-or-stricter than the old `ownerships__user__organization_memberships=`: the membership ForeignObject additionally binds the ownership organization_id to the scoped membership org, so a scoped token cannot see more calendars than before. Covered by public_api/tests/test_scoping*.py.

calendar_integration/services/calendar_service.py
```diff
@@ -946,17 +950,22 @@ def _get_write_adapter_for_calendar(self, calendar: Calendar) -> CalendarAdapter
...
                 .first()
             )
 
-            if ownership:
-                return CalendarService.get_calendar_adapter_for_account(ownership.user)[0]
+            # Resolve the owning user from the denormalized scalar instead of
+            # dereferencing ``ownership.membership`` (the ForeignObject), which
+            # resolves to ``None`` when ``membership_user_id`` is stale (the
+            # membership was deleted) and would raise ``AttributeError``.
+            if ownership and ownership.membership_user_id:
```
hugobessa, line 965: Stale-membership guard: until the Phase 2b PROTECT FK exists, a membership_user_id can point at a deleted membership, making the membership ForeignObject resolve to None. The owner user is resolved from the scalar membership_user_id (== user id) rather than ownership.membership.user, which would AttributeError.

PR #149, opened 2026-06-20
feat(calendar)!: drop CalendarOwnership.user, enforce membership PROTECT FK

hugobessa:
## Summary
Phase 2b (DB half of the Phase 2 cutover) of **Membership-Scoped Calendar References**. Drops the
`CalendarOwnership.user` column (and `User.calendar_ownerships` reverse accessor), adds a partial unique
constraint on `(calendar_fk, membership_user_id)`, and enforces PROTECT at the database via a raw-SQL
composite FK. Ownership is now membership-only. **Breaking**: the `user` column is gone.

## How — migrations
- `0025` (schema): `RemoveField(user)` wrapped in `SeparateDatabaseAndState` so the reverse re-adds
  `user_id` **nullable** and backfills `user_id = membership_user_id` (a bare auto-reverse would re-add a
  NOT NULL column and fail rollback on a populated table). Adds partial unique
  `(calendar_fk, membership_user_id) WHERE membership_user_id IS NOT NULL`.
- `0026` (DB integrity, `atomic = False`): raw-SQL composite FK
  `(membership_user_id, organization_id) → OrganizationMembership(user_id, organization_id)`,
  **`ON DELETE NO ACTION DEFERRABLE INITIALLY DEFERRED`**, added `NOT VALID` then `VALIDATE` in separate
  transactions.
- `0027` (state-only): switches the `membership` ForeignObject to `on_delete=DO_NOTHING`.

## Why DEFERRABLE + DO_NOTHING (the load-bearing design)
The `membership` relation is a `ForeignObject`, which emits no DB constraint — so PROTECT must come from
the raw FK. Two corrections landed in review:
1. The FK must be **`DEFERRABLE INITIALLY DEFERRED`** (matching every other FK on this table), not a
   non-deferrable `RESTRICT`. Deleting an `Organization` cascades (Django's Python collector) to BOTH the
   membership and the ownership; a non-deferrable check fires mid-cascade and aborts the whole delete. With
   the deferred check, a same-transaction cascade that removes both rows commits cleanly, while a
   membership-only delete still raises at commit (PROTECT preserved).
2. The ForeignObject itself must be **`on_delete=DO_NOTHING`**, not `PROTECT`: carrying `PROTECT` made
   Django's Python collector raise `ProtectedError` *eagerly* during the org cascade, defeating the
   deferred DB check. PROTECT now lives solely at the DB FK — exactly what the field type's docstring
   already promised. This change applies to all future adopters (EventAttendance, CalendarManagementToken).

## Behaviour change (documented)
Deleting a `User` cascades to their `OrganizationMembership`, which the FK now blocks while a live
`CalendarOwnership` references it. Owner-resolution writes (`create_*`, sync) set `membership_user_id`
only when a matching membership exists (a `_resolve_owner_membership_user_id` guard) so the FK never
rejects a non-member; orphans get `NULL`.

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 2 (DB half). Stacked on phase-2a (PR #148).
Note: the plan's prose wrote the FK column order as `(organization_id, membership_user_id) → (user_id,
organization_id)` (positionally wrong); the migration correctly uses
`(membership_user_id, organization_id) → (user_id, organization_id)`.

## Test plan
- `calendar_integration/tests/test_ownership_protect_fk.py`: membership-delete-while-referenced raises
  (deferred IntegrityError at `atomic` exit); **Organization-delete cascade with a member-owned ownership
  succeeds** (the deferrable regression guard); non-member `membership_user_id` rejected; NULL orphans
  allowed; partial-unique violation. Plus a sync regression for re-sync with pre-existing NULL orphans.
- Migration forward/reverse-to-0024/forward verified on the worktree DB; `pg_constraint` shows
  `condeferrable=t, condeferred=t`. Gate: `pytest -n auto` → 2640 passed; `makemigrations --check` clean.


calendar_integration/migrations/0026_calendarownership_membership_protect_fk.py
```diff
@@ -0,0 +1,119 @@
...
+The ``CalendarOwnership.membership`` relation is a Django ``ForeignObject`` and
+therefore carries **no** DB-level foreign-key constraint. PROTECT delete
+semantics are enforced here by a raw-SQL composite FK:
+
+    ALTER TABLE calendar_integration_calendarownership
+      ADD CONSTRAINT calownership_membership_protect_fk
+      FOREIGN KEY (membership_user_id, organization_id)
+      REFERENCES organizations_organizationmembership (user_id, organization_id)
+      ON DELETE NO ACTION
+      DEFERRABLE INITIALLY DEFERRED
```
hugobessa, line 12: The PROTECT FK is NO ACTION DEFERRABLE INITIALLY DEFERRED, matching every other FK on this table. A non-deferrable RESTRICT broke Organization deletion: the org cascade removes the membership and the ownership in one transaction, and an immediate check fires mid-cascade. The deferred check runs at COMMIT — same-transaction cascade passes, membership-only delete still raises. NOT VALID then VALIDATE in separate transactions (atomic=False) avoids a long lock.

common/fields.py
```diff
@@ -268,11 +268,26 @@ def contribute_to_class(self, cls, name: str) -> None:  # type: ignore[override]
...
+        #    constraint lets such whole-object cascades succeed while a
+        #    membership-only delete (referencing row still live) still raises at
+        #    commit. ``self.on_delete`` is retained for introspection/documentation
+        #    of the intended semantics only.
         fo_field = ForeignObject(
             "organizations.OrganizationMembership",
             from_fields=[user_id_field_name, "organization_id"],
             to_fields=["user_id", "organization_id"],
-            on_delete=self.on_delete,
+            on_delete=models.DO_NOTHING,
```
hugobessa, line 290: The ForeignObject is wired on_delete=DO_NOTHING regardless of the configured self.on_delete (kept for introspection). Carrying PROTECT here made Django Python cascade collector raise ProtectedError eagerly, defeating the deferred DB FK during whole-object cascades. PROTECT lives solely at the DB level (the raw composite FK). This applies to every adopter, including the EventAttendance / CalendarManagementToken phases.

calendar_integration/migrations/0025_remove_calendarownership_user_and_more.py
```diff
@@ -0,0 +1,83 @@
...
+      NULL``, so a calendar cannot have two ownerships for the same member while
+      multiple NULL (orphan) rows per calendar remain allowed.
+
+    The raw-SQL composite PROTECT FK lives in the next migration (0026), which is
+    ``atomic = False`` so its NOT VALID / VALIDATE steps don't share one
+    long-locking transaction.
+
+    Data-preserving reverse
+    -----------------------
+    The column drop is wrapped in ``SeparateDatabaseAndState`` so the model-state
```
hugobessa, line 48: RemoveField(user) is wrapped in SeparateDatabaseAndState: the state half removes the field (so makemigrations --check stays clean) while the DB half reverse re-adds user_id NULLABLE and backfills user_id = membership_user_id. A bare auto-reverse would re-add a NOT NULL column with no default and crash rollback on a populated table.

calendar_integration/models.py
```diff
@@ -263,9 +266,16 @@ class Meta:
                 name="calownership_org_member_idx",
             ),
         ]
+        constraints: ClassVar = [
+            models.UniqueConstraint(
+                fields=["calendar_fk", "membership_user_id"],
+                condition=models.Q(membership_user_id__isnull=False),
+                name="calownership_uniq_cal_member",
```
hugobessa, line 273: Partial unique (calendar_fk, membership_user_id) WHERE membership_user_id IS NOT NULL: one ownership per (calendar, member), while unlimited NULL/orphan rows per calendar remain allowed (a composite FK with a NULL column is unenforced under MATCH SIMPLE).

calendar_integration/services/calendar_service.py
```diff
@@ -126,14 +126,32 @@
...
-from organizations.models import Organization
+from organizations.models import Organization, OrganizationMembership
 from public_api.models import SystemUser
 from users.models import User
 
 
 logger = logging.getLogger(__name__)
 
 
+def _resolve_owner_membership_user_id(user: User, organization: Organization) -> int | None:
```
hugobessa, line 137: _resolve_owner_membership_user_id guard: now that the DB FK rejects a membership_user_id with no matching membership, every owner-resolution write sets the id only when an OrganizationMembership(user, org) exists, else NULL (orphan). Module-level to avoid the narrowed-self Protocol mypy class. Same guard inlined in the sync + bundle paths.

PR #150, opened 2026-06-20
feat(calendar): expand EventAttendance with membership reference + backfill

hugobessa:
## Summary
Phase 3 (expand) of **Membership-Scoped Calendar References** — the same EXPAND pattern as Phase 1
(CalendarOwnership), applied to `EventAttendance`. Adds a nullable `membership`
(`OrganizationMembershipForeignKey`) + `membership_user_id` column + composite index alongside the kept
`user` FK, and backfills `membership_user_id = user_id` where a matching `OrganizationMembership` exists.
Behaviour-preserving: no read/write/M2M/API change, no PROTECT FK (that's Phase 4).

## How
- `EventAttendance.membership = OrganizationMembershipForeignKey(related_name="event_attendances", null=True)`
  (the ForeignObject is `DO_NOTHING` internally per the Phase-2b design; PROTECT comes via Phase 4's raw FK).
  `related_name` is shared with the `user` FK — different target models (User vs OrganizationMembership), no
  clash, sets up a clean `user` drop in Phase 4. Composite index `(organization, membership_user_id)`
  = `evattend_org_member_idx`.
- Migration `0028` (schema): nullable column + index. Migration `0029` (data, `atomic=False`): batched
  PK-range idempotent/resumable `UPDATE … WHERE membership_user_id IS NULL AND user_id IS NOT NULL AND
  EXISTS(membership)`; orphans (user set, no membership) → NULL + CSV report to `.vinta-ai-workflows/one-off-runs/`
  (+ WARNING, OSError fallback); clean reverse.

## Note on null-user
`EventAttendance.user_id` is `NOT NULL` at the DB level (the sync service references `user=None` in code —
a pre-existing latent inconsistency this phase does not touch), so the `user_id IS NOT NULL` guard in the
backfill is defensive/forward-compat. Every row has a user; orphans are users with no membership in the
event's org.

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 3 (EventAttendance expand). Stacked on phase-2b (PR #149).
Mirrors the reviewed Phase 1 templates (`0022`/`0023`).

## Test plan
- `calendar_integration/tests/test_attendance_expand.py` (15 tests): matched backfill + `.membership`
  descriptor resolves; orphan stays NULL + reported; idempotent; behaviour-unchanged (`filter(user=…)`,
  `{a.user_id: a}` maps, `event.attendances`); multi-org isolation; CSV report (monkeypatched LOG_DIR);
  OSError fallback; reverse.
- Gate: `pytest -n auto` → 2655 passed; migrate forward/reverse/forward clean; `makemigrations --check` clean.


calendar_integration/models.py
```diff
@@ -450,12 +457,26 @@ class EventAttendance(OrganizationModel):
     user = models.ForeignKey(
         settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name="event_attendances"
     )
+    membership = OrganizationMembershipForeignKey(
```
hugobessa, line 460: EventAttendance gains membership alongside the kept user FK (expand step). null=True because the backfill leaves orphan rows (user not a member of the event org) NULL. The field ForeignObject is DO_NOTHING internally (Phase-2b design) — PROTECT is enforced by the raw composite FK added in the Phase 4 cutover, not here. related_name="event_attendances" is shared with user (different target models → no clash), so Phase 4 can drop user and hand the name to membership.

calendar_integration/models.py
```diff
@@ -450,12 +457,26 @@ class EventAttendance(OrganizationModel):
...
         max_length=50,
         choices=RSVPStatus,
         default=RSVPStatus.PENDING,
     )
 
+    class Meta:
+        indexes: ClassVar = [
+            models.Index(
+                fields=["organization", "membership_user_id"],
+                name="evattend_org_member_idx",
```
hugobessa, line 476: Tenant-leading composite index (organization, membership_user_id) — the per-table index responsibility documented on OrganizationMembershipForeignKey. EventAttendance is higher-volume than CalendarOwnership, so the index matters more for the Phase-4 join + FK validation.

calendar_integration/migrations/0029_backfill_eventattendance_membership_user_id.py
```diff
@@ -0,0 +1,140 @@
...
+def reverse_backfill_membership_user_id(apps, schema_editor):
+    """Roll back: clear membership_user_id on all EventAttendance rows."""
+    with connection.cursor() as cursor:
+        cursor.execute("UPDATE calendar_integration_eventattendance SET membership_user_id = NULL")
+
+
+class Migration(migrations.Migration):
+    """Backfill EventAttendance.membership_user_id (Phase 3 data migration)."""
+
+    atomic = False
```
hugobessa, line 128: atomic = False + batched PK-range UPDATE = genuinely resumable on this higher-volume table (committed batches persist on a mid-run failure; no whole-table lock). Same vetted shape as Phase 1 0023.

calendar_integration/migrations/_0029_backfill_helpers.py
```diff
@@ -0,0 +1,105 @@
...
+with ``membership_user_id = NULL`` and are reported via the caller.
+"""
+
+from django.db import connection
+
+
+BATCH_SIZE = 500
+
+
+def backfill_membership_user_id_sql() -> None:
```
hugobessa, line 31: Backfill sets membership_user_id = user_id only where a matching OrganizationMembership exists (orphans stay NULL, reported). The raw cursor.execute deliberately runs across ALL organizations — a sanctioned cross-org data-migration exception (AGENTS.md flags tenant-table cursor.execute as a bypass to avoid).

PR #151, opened 2026-06-20
feat(calendar): route EventAttendance through membership (app-layer cutover)

hugobessa:
## Summary
Phase 4a (app-layer half of the EventAttendance cutover) of **Membership-Scoped Calendar References** —
mirrors the merged Phase 2a (CalendarOwnership). Routes `EventAttendance.user` reads/writes through
`membership`/`membership_user_id`, replaces `CalendarEvent.attendees` (M2M→User) with
`attendee_memberships` (M2M→OrganizationMembership through `EventAttendance.membership`), and exposes the
attendee as a membership `{ user_id, organization_id, role }` in GraphQL + REST. The `user` **column is
kept** — Phase 4b drops it + adds the raw PROTECT FK.

## How
- Attendee identity resolves from the denormalized `attendance.membership_user_id` (survives the 4b column
  drop; stale-safe vs `attendance.membership.user`). Orphan attendances (`membership_user_id IS NULL` — the
  attendee is not a member of the event's org) are excluded from membership reads / serialization / the M2M.
- Bulk attendance writes set `membership_user_id` only for members via `resolve_member_user_ids` (the bulk
  analogue of Phase 2b's `_resolve_owner_membership_user_id`), so the Phase-4b FK never rejects a non-member.
- `attendee_memberships = ManyToManyField(OrganizationMembership, through=EventAttendance,
  through_fields=("event","membership"), related_name="attended_events")`. Migration `0030` is **state-only**
  (M2M-through declares no table). Calendar **adapters** (`event_data.attendees` DTO) are untouched.
- API: `EventAttendanceGraphQLType.user` → `membership`; `CalendarEvent.attendees` GraphQL →
  `attendee_memberships`; REST attendance serializer exposes membership; `EventAttendanceInputData` keeps
  `user_id` on input, resolved to membership server-side. `schema.yml` regenerated.

## Review hardening (post-review fix in this branch)
- **Permission-diff symmetry**: `serialize_event` dropped orphans but `serialize_event_data_input` included
  them, so the attendee diff (old=serialize_event vs new=serialize_event_data_input) falsely flagged an
  unchanged orphan attendee as "added" → spurious `UPDATE_ATTENDEES` / `PermissionDenied`. Fixed by filtering
  the input side through `resolve_member_user_ids` too, so both sides agree on the member set.
- Batched per-event user lookups (removed an N+1 over `users.User` in the serialize / permission-grant loops).

## Orphan / data-loss note (for Phase 4b)
[...]

PR #152, opened 2026-06-20
feat(calendar)!: drop EventAttendance.user, enforce membership PROTECT FK

hugobessa:
## Summary
Phase 4b (DB half of the EventAttendance cutover) of **Membership-Scoped Calendar References** — mirrors the
merged Phase 2b (CalendarOwnership). Drops `EventAttendance.user` and enforces PROTECT at the DB via a raw-SQL
composite FK. **Breaking**: the `user` column is gone; attendance is membership-only.

## User-confirmed decision (data loss accepted)
`EventAttendance.user` was NOT NULL but `membership_user_id` is NULL for non-member ("orphan") attendances.
Dropping `user` permanently erases attendee identity for orphan rows (they were CSV-reported in Phase 3). The
user explicitly chose this uniform end-state over keeping a nullable fallback (plan Open Question #3).

## How — migrations
- `0031` (schema): `RemoveField(user)` wrapped in `SeparateDatabaseAndState` — state half removes the field;
  DB half drops `user_id` and its reverse re-adds `user_id` **nullable** + backfills `user_id =
  membership_user_id` (lossy for orphans — documented). No partial unique added (EventAttendance has no
  `(event, member)` upsert/unique, and a new one could fail on pre-existing duplicate rows).
- `0032` (DB integrity, `atomic=False`): raw-SQL composite FK `(membership_user_id, organization_id) →
  OrganizationMembership(user_id, organization_id)` `ON DELETE NO ACTION DEFERRABLE INITIALLY DEFERRED`,
  `NOT VALID` then `VALIDATE` in separate transactions. `pg_constraint`: `condeferrable=t, condeferred=t`.

## Review hardening (post-review fix in this branch)
- **Sync duplicate-orphan regression (BLOCKER)**: after the column drop, a synced attendee resolving to a
  User who is NOT a member had no identity to dedupe on, so `_process_event_attendees` appended a fresh
  orphan attendance on EVERY sync (unbounded growth). Fixed: non-member resolved Users now route to the
  **external-attendee** path (deduped by email) — internal attendance is membership-only post-cutover. A
  twice-call idempotency test guards it.

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 4 (DB half). Stacked on phase-4a (PR #151).

## Test plan
- `calendar_integration/tests/test_attendance_protect_fk.py` (9 tests, mirrors 2b): membership-delete-blocked
  (deferred IntegrityError at atomic exit), user-cascade-blocked, **Org-cascade-succeeds** (`transaction=True`
  regression guard), non-member `membership_user_id` rejected (insert + update), NULL orphans allowed.
- `test_process_event_attendees_non_member_becomes_external_and_is_idempotent` — non-member → external, no
  duplicate on re-sync.
- Gate: `pytest -n auto` → 2662 passed; migrate forward/reverse/forward clean; `makemigrations --check` clean.


calendar_integration/models.py
```diff
@@ -438,14 +438,17 @@ def __str__(self):
 
 class EventAttendance(OrganizationModel):
```
hugobessa, line 439: EventAttendance.user dropped — attendance is membership-only. USER-CONFIRMED data-loss decision: orphan attendances (non-member, membership_user_id IS NULL) permanently lose attendee identity (CSV-reported in Phase 3). Migration 0031 reverse re-adds user_id nullable + backfills from membership_user_id, lossy for orphans.

calendar_integration/migrations/0032_eventattendance_membership_protect_fk.py
```diff
@@ -0,0 +1,119 @@
...
+The ``EventAttendance.membership`` relation is a Django ``ForeignObject`` and
+therefore carries **no** DB-level foreign-key constraint. PROTECT delete
+semantics are enforced here by a raw-SQL composite FK:
+
+    ALTER TABLE calendar_integration_eventattendance
+      ADD CONSTRAINT evattendance_membership_protect_fk
+      FOREIGN KEY (membership_user_id, organization_id)
+      REFERENCES organizations_organizationmembership (user_id, organization_id)
+      ON DELETE NO ACTION
+      DEFERRABLE INITIALLY DEFERRED
```
hugobessa, line 12: Same deferrable PROTECT FK pattern as Phase 2b (NO ACTION DEFERRABLE INITIALLY DEFERRED, NOT VALID then VALIDATE, atomic=False): membership-delete-while-referenced raises at commit; an Org cascade that removes membership + attendance in one txn passes. The membership ForeignObject is DO_NOTHING, so this is the only PROTECT enforcement.

calendar_integration/migrations/0031_remove_eventattendance_user.py
```diff
@@ -0,0 +1,86 @@
...
+      new partial unique on ``(event_fk, membership_user_id)`` could fail on
+      pre-existing duplicate member attendances, so it is deliberately skipped.
+
+    The raw-SQL composite PROTECT FK lives in the next migration (0032), which is
+    ``atomic = False`` so its NOT VALID / VALIDATE steps don't share one
+    long-locking transaction.
+
+    Data-preserving (but orphan-lossy) reverse
+    ------------------------------------------
+    The column drop is wrapped in ``SeparateDatabaseAndState`` so the model-state
```
hugobessa, line 56: SeparateDatabaseAndState: state half removes the field (makemigrations stays clean); DB half reverse re-adds user_id NULLABLE and backfills user_id = membership_user_id so a rollback does not crash on the populated table — lossy for orphan rows. No partial unique added (no (event, member) upsert; a new one could fail on pre-existing dup rows).

calendar_integration/services/calendar_sync_service.py
```diff
@@ -864,28 +864,41 @@ def _process_event_attendees(
...
                         organization_id=existing_event.organization_id,
                         event=existing_event,
-                        user=user,
                         membership_user_id=membership_user_id,
                         status=attendee.status,
                     )
                 )
             elif (
-                not user
+                membership_user_id is None
```
hugobessa, line 901: Sync BLOCKER fix: a synced attendee whose email resolves to a non-member User now routes to the external-attendee branch (membership_user_id is None), deduped by email — instead of creating an identity-less internal orphan attendance on every sync (unbounded growth, since there is no dedupe key after the user column drop). Internal attendance == membership post-cutover; non-members are external.

PR #153, opened 2026-06-20
feat(calendar): expand CalendarManagementToken with membership reference + backfill

hugobessa:
## Summary
Phase 5 (expand) of **Membership-Scoped Calendar References** — the same EXPAND pattern as Phases 1/3,
applied to `CalendarManagementToken`. Adds a nullable `membership` + `membership_user_id` column + composite
index alongside the kept nullable `user` FK, and backfills `membership_user_id = user_id` where the user is a
member. Behaviour-preserving: no read/write/API change, no PROTECT FK (that's Phase 6).

## Token-specific nuance (null-user ≠ orphan)
`CalendarManagementToken.user` is NULLABLE — a token belongs to EITHER a `user` OR an `external_attendee`
(booking codes). Null-user (external) tokens are legitimately member-less and are **never backfilled and never
reported**. Orphans = `user_id IS NOT NULL` but no matching `OrganizationMembership` → NULL + CSV report. Both
the backfill UPDATE and `collect_orphans` carry the `AND user_id IS NOT NULL` guard.

## How
- `membership = OrganizationMembershipForeignKey(related_name="calendar_event_management_tokens", null=True)`
  (ForeignObject is DO_NOTHING; PROTECT via Phase 6's raw FK). `related_name` shared with `user` (different
  target models → no clash; sets up the Phase-6 drop). Index `(organization, membership_user_id)` =
  `calmgmttoken_org_member_idx`.
- Migration `0033` (schema): nullable column + index. Migration `0034` (data, `atomic=False`): batched,
  idempotent/resumable backfill; orphan CSV to `.vinta-ai-workflows/one-off-runs/` (+ WARNING, OSError
  fallback); clean reverse.

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 5 (token expand). Stacked on phase-4b (PR #152).
Mirrors the reviewed Phase 3 templates (`0028`/`0029`).

## Test plan
- `calendar_integration/tests/test_management_token_expand.py` (10 tests): member-token backfill +
  `.membership` descriptor resolves; **null-user (external) token stays NULL + not reported**; orphan stays
  NULL + reported; idempotent; behaviour-unchanged (`filter(user=…)`); CSV report (monkeypatched LOG_DIR);
  OSError fallback; reverse; multi-org.
- Gate: `pytest -n auto` → 2672 passed; migrate forward/reverse/forward clean; `makemigrations --check` clean.

PR #154, opened 2026-06-20
feat(calendar)!: route CalendarManagementToken through membership, drop user + PROTECT FK

hugobessa:
## Summary
Phase 6 — the full cutover for `CalendarManagementToken` (no M2M, so app-rewrite + drop + PROTECT FK in one
phase). Routes `token.user` reads/writes through `membership`/`membership_user_id`, drops the `user` column,
and enforces PROTECT via a deferrable raw composite FK. Token actor is now membership-scoped. **Breaking**:
the `user` column is gone.

## How
- **Audit actor** (`_resolve_token_audit_actor`): preserves the pre-cutover 3-way behavior exactly — a member
  token resolves to its `User` (via `membership_user_id`); an external/null-membership token uses the token
  itself as actor; no token → `context.user_or_token`. (The side-effects `actor=` already accepted a
  `CalendarManagementToken` for external tokens.)
- **Mint guard** (`_resolve_token_membership_user_id`): member token-mint sets `membership_user_id` only when
  an `OrganizationMembership(user, org)` exists (else NULL — a non-member can't act through a membership), so
  the new FK never aborts a mint. External-attendee + booking tokens leave `membership_user_id` NULL (the FK is
  not enforced for NULLs).
- **Actor comparison**: `user_id == token.membership_user_id` (was `== token.user_id`) — equivalent for
  members; for an external token (NULL) it's always False, same as before.
- Migration `0035`: `RemoveField(user)` via `SeparateDatabaseAndState` (reverse re-adds `user_id` NULLABLE +
  backfills from `membership_user_id`). No partial unique (a member may hold many tokens). Migration `0036`
  (`atomic=False`): deferrable raw PROTECT FK `(membership_user_id, organization_id) → OrganizationMembership
  (user_id, organization_id)` `ON DELETE NO ACTION DEFERRABLE INITIALLY DEFERRED`, `NOT VALID` then `VALIDATE`.
- No API exposure changed — tokens are auth artifacts (no serializer/GraphQL exposed the token's user;
  `schema.yml` unchanged).

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 6. Stacked on phase-5 (PR #153). Combines the vetted
2a/4a app-rewrite + 2b/4b drop+deferrable-FK patterns.

## Test plan
- `calendar_integration/tests/test_management_token_protect_fk.py` (mirrors 2b/4b): membership-delete-blocked
  (deferred IntegrityError at atomic exit), user-delete-blocked, **Org-cascade-succeeds** (`transaction=True`),
  delete-allowed-after-token-removed, non-member `membership_user_id` rejected, NULL (external) allowed.
- Service tests retargeted for the membership actor / mint / consume flows.
- Gate: `pytest -n auto` → 2671 passed; migrate forward/reverse/forward clean; `makemigrations --check` clean;
  FK `condeferrable=t, condeferred=t`.


calendar_integration/services/calendar_event_service.py
```diff
@@ -98,6 +98,29 @@
 from users.models import User
 
 
+def _resolve_token_audit_actor(
```
hugobessa, line 101: _resolve_token_audit_actor preserves the pre-cutover 3-way audit behavior exactly: member token → resolved User (via membership_user_id); external/null-membership token → the token itself; (no token → context.user_or_token at the call site). Verified against phase-5 logic. The side-effects actor= already accepted a CalendarManagementToken for external tokens.

calendar_integration/services/calendar_permission_service.py
```diff
@@ -34,13 +34,37 @@
...
 )
+from organizations.models import OrganizationMembership
 from users.models import User
 
 
 if TYPE_CHECKING:
     from public_api.models import SystemUser
 
 
+def _resolve_token_membership_user_id(user: User, organization_id: int) -> int | None:
```
hugobessa, line 45: _resolve_token_membership_user_id mint guard (the token analogue of _resolve_owner_membership_user_id): a member token-mint sets membership_user_id only when the membership exists, else NULL — so the new FK never rejects a non-member on mint. External/booking token mints leave it NULL (FK unenforced for NULLs).

calendar_integration/migrations/0036_calendarmanagementtoken_membership_protect_fk.py
```diff
@@ -0,0 +1,113 @@
...
+
+The ``CalendarManagementToken.membership`` relation is a Django ``ForeignObject``
+and therefore carries **no** DB-level foreign-key constraint. PROTECT delete
+semantics are enforced here by a raw-SQL composite FK:
+
+    ALTER TABLE calendar_integration_calendarmanagementtoken
+      ADD CONSTRAINT calmgmttoken_membership_protect_fk
+      FOREIGN KEY (membership_user_id, organization_id)
+      REFERENCES organizations_organizationmembership (user_id, organization_id)
+      ON DELETE NO ACTION
```
hugobessa, line 11: Same deferrable PROTECT FK as Phase 2b/4b (NO ACTION DEFERRABLE INITIALLY DEFERRED, NOT VALID then VALIDATE, atomic=False). The membership ForeignObject is DO_NOTHING, so this is the only PROTECT enforcement. Org-cascade passes; membership-only delete raises at commit.

calendar_integration/migrations/0035_remove_calendarmanagementtoken_user.py
```diff
@@ -0,0 +1,77 @@
...
+      no ``(scope, membership_user_id)`` upsert key that needs a partial unique, and
+      adding one would reject valid duplicate member tokens.
+
+    The raw-SQL composite PROTECT FK lives in the next migration (0036), which is
+    ``atomic = False`` so its NOT VALID / VALIDATE steps don't share one
+    long-locking transaction.
+
+    Data-preserving reverse
+    -----------------------
+    The column drop is wrapped in ``SeparateDatabaseAndState`` so the model-state
```
hugobessa, line 50: SeparateDatabaseAndState: state half removes the field (makemigrations stays clean); DB half reverse re-adds user_id NULLABLE (matching its original nullable definition) and backfills user_id = membership_user_id. No partial unique added — a member can legitimately hold multiple tokens.

PR #155, opened 2026-06-20
refactor(membership)!: convert remaining real membership FKs to (org,user) ForeignObject

hugobessa:
## Summary
Phase 7a — the enabling step before Phase 7b makes `OrganizationMembership`'s PK composite. Django 6 forbids
a real FK to a composite-PK model, so the two remaining real references are converted to the
`(organization_id, user_id)` ForeignObject pattern, and every scalar `membership.id`/`.pk`/`membership_id`
consumer is rewritten. `OrganizationMembership` STILL has its `id` PK this phase (composite PK is 7b).

## What changed
- **`OrganizationInvitation.membership`** (was `OneToOneField`) → `OrganizationMembershipForeignKey`. OneToOne
  semantics preserved by a partial `UniqueConstraint(organization, membership_user_id) WHERE membership_user_id
  IS NOT NULL`; `membership.invitation` reverse accessor kept (now a reverse manager — no code relied on
  single-object semantics).
- **`public_api.SystemUser.scoped_to_membership`** (was `OrganizationForeignKey`, a real FK to `membership.id`)
  → `OrganizationMembershipForeignKey`. NULL still = org-wide token.
- **All `membership.id`/`.pk`/`membership_id` rewritten** to `(user_id, organization_id)` / `membership_user_id`
  across `organizations/services.py`, `public_api/*`, `webhooks/services/*`, `calendar_event_service.py`.
- Two migrations (`organizations/0012`, `public_api/0008`): add `*_user_id`, backfill from the legacy FK's
  `.user_id`, drop the legacy real-FK columns via `SeparateDatabaseAndState` (reverse re-adds nullable +
  backfills) — this drops `membership_id` + `scoped_to_membership_fk_id`, unblocking Phase 7b's PK swap.

## Security — token scoping (verified equal-or-stricter, no cross-org leak)
`scoped_calendar_ids` / `_scoped_system_user_owns_calendar` changed from `ownerships__membership__id=fk_id`
to `ownerships__membership_user_id=scoped_to_membership_user_id` within `filter_by_organization(org)`. This is
equivalent because the `organization` is always the token's own org (the public-API middleware sets it from
`system_user.organization`) and `scoped_to_membership` is always a membership in that org (constrained at
token-creation). A user who is a member of multiple orgs cannot leak across orgs — the org is fixed.

## Consumer-visible
- Webhook `OrganizationMemberCreatedWebhookPayload` drops the scalar `membership_id`; membership identity is the
[...]

PR #156, opened 2026-06-20
feat(organizations)!: composite primary key (user, organization) on OrganizationMembership

hugobessa:
## Summary
Phase 7b (FINAL) — `OrganizationMembership`'s primary key becomes the composite `(user_id, organization_id)`
and the implicit `id` BigAutoField is dropped. A membership's identity *is* the (user, org) pair. Enabled by
Phase 7a (which removed the last real FKs + all `.id` references).

## How (the tricky bits)
- **`SeparateDatabaseAndState`**: Django's `CompositePrimaryKey` is virtual — the autodetector emits only
  `DROP COLUMN id` and treats `AddField(pk)` as a DB no-op, leaving NO real PRIMARY KEY. So the *state* half is
  Django's `RemoveField(id)`+`AddField(pk)` (keeps `makemigrations --check` clean) and the *database* half is
  raw SQL: `DROP CONSTRAINT _pkey; DROP COLUMN id; ADD PRIMARY KEY (user_id, organization_id)`.
- **Option A (keep the unique)**: the 3 calendar PROTECT FKs bind to `uniq_membership_user_organization`.
  Keeping that unique means the FKs are never dropped/re-added; the composite PK is added alongside (redundant
  but legal). Django `check` is clean with the redundancy. Verified post-swap: composite PK present, `id` gone,
  all 3 FKs still `condeferrable=t, condeferred=t`, valid.
- **`SafeCompositePrimaryKey`** (`common/fields.py`): Django's stock `CompositeAttribute.__get__` crashes on
  class-level `Model.pk` access (`dir(cls)` introspection in django_virtual_models). The subclass returns the
  descriptor on class-level access (standard descriptor convention); instance access is unchanged.
  `deconstruct` round-trips to `common.fields.SafeCompositePrimaryKey`, so makemigrations stays clean.
- **Cross-app migration deps**: organizations 0012, calendar_integration 0036 (the 3 FKs must exist), AND
  public_api 0008 — critical, because the legacy `SystemUser.scoped_to_membership` FK depended on the *id PK
  index*; it must be dropped before `DROP CONSTRAINT _pkey`.
- Views use `lookup_field="user_id"` for the org-scoped membership detail routes; membership identity is
  `(user_id, organization_id)` in serializers.

## Lock / rollout (production)
The PK swap takes `ACCESS EXCLUSIVE` on `organizations_organizationmembership` (no Postgres `ADD PRIMARY KEY
CONCURRENTLY`). The table is low-cardinality (one row per user-per-org) so the index build is fast, but it
briefly blocks all membership access (which gates every tenant request) — **schedule in a low-traffic
window**. `atomic=True` ensures no no-PK window. Does NOT lock the 3 referencing calendar tables. Reverse
restores the `id` identity PK (original id *values* can't be recovered — documented).

## Plan reference
Plan `…2026-06-19-…_IMPLEMENTATION_PLAN.md` — Phase 7b (final). Stacked on phase-7a (PR #155).

## Test plan
- `organizations/tests/test_models.py`: composite-PK tuple lookup `get(pk=(user_id, org_id))`; `membership.pk`
  is the tuple; duplicate `(user, org)` raises `IntegrityError` (DB-level, post-swap).
- Existing membership / invitation / scoped-token / calendar suites pass against the composite PK.
- Gate: `pytest -n auto` → 2674 passed; `makemigrations --check` clean; `check --deploy` clean; migrate
  forward→reverse(0012)→forward clean; pg_constraint confirms composite PK + 3 deferrable FKs valid, `id` gone.


organizations/migrations/0013_organizationmembership_composite_pk.py
```diff
@@ -0,0 +1,173 @@
+"""Phase 7b (FINAL): swap ``OrganizationMembership`` to a composite primary key.
```
hugobessa, line 1: The crux of the plan. SeparateDatabaseAndState is required because Django CompositePrimaryKey is virtual (autodetector emits only DROP COLUMN id, leaving no real PK) — the DB half hand-writes the PK swap. Option A keeps uniq_membership_user_organization so the 3 calendar PROTECT FKs that bind to it stay valid without drop/re-add. Cross-app deps order the legacy-FK drops (public_api 0008, organizations 0012) before DROP CONSTRAINT _pkey, which depends on the id PK index.

common/fields.py
```diff
@@ -33,12 +33,50 @@
...
 from django.db.backends.base.operations import BaseDatabaseOperations
 from django.db.models import AutoField, UUIDField
+from django.db.models.fields.composite import CompositeAttribute, CompositePrimaryKey
 from django.db.models.fields.related import ForeignObject
 
 
 BaseDatabaseOperations.integer_field_ranges["UUIDField"] = (0, 0)
 
 
+class _SafeCompositeAttribute(CompositeAttribute):
```
hugobessa, line 43: SafeCompositePrimaryKey works around Django CompositeAttribute.__get__ crashing on class-level Model.pk access (getattr(cls, "pk") during dir(cls) introspection, e.g. django_virtual_models). It returns the descriptor on class access (standard descriptor convention), unchanged for instances; deconstruct round-trips so migrations stay clean.

PR #160, opened 2026-06-23
feat(calendar): edit internal resource-calendar capacity via updateResourceCalendar

hugobessa:
## Summary
Lets an integration edit a manual (`provider=INTERNAL`) resource calendar's `capacity` (and other resource attributes) through a dedicated Public GraphQL mutation, while rejecting Google-synced calendars. Implements the "Resource-calendar capacity edit" item from the Building Blocks v3 integration surface. Single branch, atomic commits per logical unit (modular-commits strategy).

## How
A new sibling service method `CalendarService.update_resource_calendar(...)` mirrors the existing `update_calendar` / `disable_resource_calendar` pair rather than overloading `update_calendar` (which is hard-scoped to `CalendarType.PERSONAL`). The guard requires `provider == INTERNAL` **and** `calendar_type == RESOURCE`; it runs **after** the org-scoped lookup so cross-org access can never leak existence. A module-level `_UNCHANGED` sentinel gives `capacity` three states: omit (untouched), explicit `null` (clear to unlimited), int (set) — `None` can't double as "no change" because clearing is a valid op. The `updateResourceCalendar` Public GraphQL mutation (org-wide token only, new `update_resource_calendar` resource grant) wires on top; `visibility` is validated against `CalendarVisibility` choices.

## Phases (commit groups on this branch)
- **Phase 1** ✅ — `update_resource_calendar` service method + INTERNAL/RESOURCE guard + unit tests.
- **Phase 2** ✅ — `updateResourceCalendar` Public GraphQL mutation + org-wide resource grant + integration tests + service-side `visibility` choice validation.

## Plan reference
Plan `ai-plans/2026-06-22-UPDATE_RESOURCE_CALENDAR_IMPLEMENTATION_PLAN.md`. No feature flag (purely additive surface: new mutation, new resource grant, new service method; no existing caller changes).

## Test plan
- `calendar_integration/tests/services/test_calendar_service.py::TestUpdateResourceCalendarService` (16 tests) — three capacity states (set / clear / omit), each editable field, synced-provider and wrong-type guards, cross-org `DoesNotExist`, audit branch (recorded on change, skipped on no-op).
- `public_api/tests/test_calendar_mutations.py::TestUpdateResourceCalendarMutation` (17 tests) — happy paths per field, three capacity states, synced + wrong-type rejection, cross-org `"Calendar not found."`, `organizationId`-ignored (token org wins), no-grant denial, unauthenticated denial, grant-absent-from-`PROVIDER_SCOPED_RESOURCES`, invalid-visibility rejection.
- Outer gate (re-verified by orchestrator): `python manage.py check --deploy` (warnings-only) + full `pytest -n auto` (exit 0).


calendar_integration/services/calendar_service.py
```diff
@@ -145,6 +145,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Sentinel for partial updates: distinguishes "omit capacity" from "explicit null"
+_UNCHANGED = object()
```
hugobessa, line 149: Module-level sentinel for the capacity field's three-state contract. `capacity` is the one editable field where `None` is a meaningful value (clear to unlimited), so it cannot also mean "leave unchanged" the way None does for every other partial-update field here. `_UNCHANGED` is the default, so an omitted `capacity` is left untouched, an explicit `None` clears it, and an int sets it. The GraphQL layer in Phase 2 maps `strawberry.UNSET` to this sentinel.

calendar_integration/services/calendar_service.py
```diff
@@ -947,6 +950,105 @@ def disable_resource_calendar(self, calendar_id: int) -> Calendar:
...
+                f"Calendar {calendar_id} is synced from an external provider "
+                f"(provider={calendar.provider}) and cannot be edited."
+            )
+
+        if calendar.calendar_type != CalendarType.RESOURCE:
+            raise ValueError(
+                f"Calendar {calendar_id} is not a resource calendar "
+                f"(type={calendar.calendar_type})."
+            )
+
```
hugobessa, line 1006: Guard ordering is load-bearing: the org-scoped `filter_by_organization(...).get()` runs FIRST, so a calendar in another org raises `Calendar.DoesNotExist` (surfaced as a generic "not found") before the provider/type checks ever run. That means the INTERNAL-vs-synced and RESOURCE-vs-other `ValueError`s can only fire on a calendar the caller's org actually owns — no cross-org existence leak. Synced (`provider != INTERNAL`) calendars are rejected because edits here are not propagated back to the external provider; this method is INTERNAL-only by design.

calendar_integration/services/calendar_service.py
```diff
@@ -947,6 +950,105 @@ def disable_resource_calendar(self, calendar_id: int) -> Calendar:
...
+        if name is not None:
+            calendar.name = name
+            update_fields.append("name")
+        if description is not None:
+            calendar.description = description
+            update_fields.append("description")
+        if capacity is not _UNCHANGED:
+            calendar.capacity = capacity
+            update_fields.append("capacity")
```
hugobessa, line 1025: The sentinel check is `is not _UNCHANGED` (identity), not a truthiness or `is not None` test — that is what lets an explicit `capacity=None` fall through and clear the column while an omitted capacity is skipped. The other five fields keep the established `None = no change` convention from `update_calendar`.

calendar_integration/services/calendar_service.py
```diff
@@ -947,6 +950,105 @@ def disable_resource_calendar(self, calendar_id: int) -> Calendar:
...
+                "name": calendar.name,
+                "description": calendar.description,
+                "capacity": calendar.capacity,
+                "manage_available_windows": calendar.manage_available_windows,
+                "accepts_public_scheduling": calendar.accepts_public_scheduling,
+                "visibility": calendar.visibility,
+            }
+            self._audit_calendar_write(
+                AuditAction.UPDATE, calendar, diff=compute_diff(before, after)
+            )
```
hugobessa, line 1052: Audit + save are gated on `update_fields` being non-empty, matching the `update_calendar` precedent: a call that changes nothing performs no write and emits no audit event. `compute_diff` produces the `{field: {old, new}}` shape consumed by the audit trail. (The audit-on-change and no-audit-on-no-op paths both have explicit tests.)

public_api/permissions.py
```diff
@@ -62,6 +62,7 @@ class OrganizationResourceAccess(BasePermission):
         "updateCalendar": PublicAPIResources.UPDATE_CALENDAR,
         "createResourceCalendar": PublicAPIResources.CREATE_RESOURCE_CALENDAR,
         "disableResourceCalendar": PublicAPIResources.DISABLE_RESOURCE_CALENDAR,
+        "updateResourceCalendar": PublicAPIResources.UPDATE_RESOURCE_CALENDAR,
```
hugobessa, line 65: The new grant is mapped here for the GraphQL field gate, but deliberately NOT added to PROVIDER_SCOPED_RESOURCES — `updateResourceCalendar` is org-wide-token-only, mirroring `updateCalendar` / `createResourceCalendar` / `disableResourceCalendar`. A provider-scoped token cannot call it. There is an explicit regression test asserting the grant's absence from PROVIDER_SCOPED_RESOURCES so a future edit can't silently widen the surface.

public_api/mutations.py
```diff
@@ -273,6 +274,43 @@ class DisableResourceCalendarResult:
...
+    capacity: If provided as an integer, sets the capacity to that value. If provided as
+        null (None), clears the capacity to unlimited. If omitted entirely (UNSET), the
+        existing capacity is left unchanged.
+    """
+
+    organization_id: int
+    calendar_id: int
+    name: str | None = None
+    description: str | None = None
+    capacity: int | None = strawberry.UNSET  # type: ignore[assignment]
```
hugobessa, line 299: `capacity` defaults to `strawberry.UNSET` (not None) so the resolver can tell "field omitted" from "explicitly sent as null". That distinction is the whole point of the feature: omit = leave capacity unchanged, explicit null = clear to unlimited, int = set. The `# type: ignore` is because UNSET isn't an `int | None`.

public_api/mutations.py
```diff
@@ -1346,6 +1384,62 @@ def disable_resource_calendar(
...
+        if input.is_private is not None:
+            accepts_public_scheduling = not input.is_private
+
+        # Translate UNSET → the _UNCHANGED sentinel so the service knows to leave capacity alone.
+        # An explicit None clears the capacity; an integer sets it.
+        # mypy: _UNCHANGED is object() but CalendarService.update_resource_calendar accepts it as
+        # the sentinel "int | None"; suppress the mismatch, matching the service's own annotation.
+        capacity: int | None = (  # type: ignore[assignment]
+            _UNCHANGED if input.capacity is strawberry.UNSET else input.capacity  # type: ignore[assignment]
+        )
```
hugobessa, line 1424: The bridge between the GraphQL UNSET convention and the service's `_UNCHANGED` sentinel: UNSET maps to `_UNCHANGED` (service leaves capacity alone); anything else (including an explicit `None`) is passed straight through so None reaches the service as "clear". Org is taken from the token via `_get_org_and_init_calendar_service(info)` — `input.organization_id` is never trusted (proved by `test_input_organization_id_is_ignored_token_org_wins`).

calendar_integration/services/calendar_service.py
```diff
@@ -947,6 +950,109 @@ def disable_resource_calendar(self, calendar_id: int) -> Calendar:
...
+            calendar.capacity = capacity
+            update_fields.append("capacity")
+        if manage_available_windows is not None:
+            calendar.manage_available_windows = manage_available_windows
+            update_fields.append("manage_available_windows")
+        if accepts_public_scheduling is not None:
+            calendar.accepts_public_scheduling = accepts_public_scheduling
+            update_fields.append("accepts_public_scheduling")
+        if visibility is not None:
+            if visibility not in CalendarVisibility.values:
```
hugobessa, line 1033: `visibility` is the only choice-constrained editable field, and this is the first public-API surface to expose it for writes, so the service validates it against `CalendarVisibility.values` and raises `ValueError` (→ `success=false`) for anything else. Without this, an external integration could persist an out-of-choices string (Django doesn't enforce choices on `.save()`). This is the plan's recommended default ("ship as str, validated by the service").

PR #161, opened 2026-06-23
feat(calendar): allow owner-scoped & org-wide public tokens to update/delete events

hugobessa:
## Summary
Phase 0a of the **Owner-Scoped Public Reschedule / Cancel Mutations** plan. This is the service-layer foundation that the upcoming `rescheduleCalendarEvent` / `rescheduleCalendarGroupEvent` / `cancelEvent` public GraphQL mutations need — it does **not** add any GraphQL surface yet.

Today `CalendarEventService.update_event` and `delete_event` reject **every** public-API `SystemUser` principal with `PermissionDenied("Events cannot be created through the Public API.")`. The `*WithCode` mutations dodge this because they authenticate with a booking-code token (a `str`), not a `SystemUser`. `scheduleEvent`/`create_event` works because `create_event` has a sanctioned owner-scoped allowance that update/delete lack.

This phase adds the missing allowance via a single choke-point, `_public_token_may_write(system_user, calendar)`:
- **Org-wide** tokens (`scoped_to_membership_user_id is None`) → allowed for any event **in their own organization** (tenant boundary re-derived from the loaded objects, not trusted from the caller).
- **Owner-scoped** tokens → allowed only when the token's owner independently owns the target calendar (`_scoped_system_user_owns_calendar`, re-derived from `CalendarOwnership` rows).
- Otherwise → `PermissionDenied("Calendar matching query does not exist.")` — byte-identical to a genuinely missing calendar, so a cross-owner caller cannot tell a forbidden calendar from a missing one (no existence leak).

Deliberately **diverges from `create_event`**, which stays org-wide-blocked (per the plan's Guiding Decisions: "org-wide tokens act org-wide" for reschedule/cancel, but not for create). A regression test pins that `create_event` for an org-wide token is still rejected.

Also: scoped tokens are rejected on BUNDLE calendars (org-wide follow existing bundle rules), and the side-effects/webhook audit actor on the public-token path now falls back to the `SystemUser` caller (parity with `create_event`) instead of being actor-less.

## Plan reference
`ai-plans/2026-06-22-OWNER_SCOPED_RESCHEDULE_CANCEL_MUTATIONS_IMPLEMENTATION_PLAN.md` — **Phase 0a**. First phase of 5 (0a → 0b → 1 → 2 → 3). No feature flag (purely additive surface). No migrations.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_calendar_event_service.py -n auto` — 12 new service-layer tests (owner-scoped allow, cross-owner deny w/ not-found parity, org-wide allow, cross-tenant org-wide deny, `create_event` stays blocked, scoped bundle reject, Django-user path unchanged, audit-actor = SystemUser on update + delete).
- Outer gate: `docker compose run --rm api uv run python manage.py check --deploy` and `docker compose run --rm api uv run pytest -n auto` → 3150 passed.

### DIFF
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -1022,7 +1022,11 @@ class CalendarEventService:
             if not context.calendar_side_effects_service:
                 return
 
-            actor = _resolve_token_audit_actor(context.calendar_permission_service.token)
+            permission_token = getattr(context.calendar_permission_service, "token", None)
+            if permission_token is not None:
+                actor: Any = _resolve_token_audit_actor(permission_token)
+            else:
+                actor = context.user_or_token
             context.calendar_side_effects_service.on_update_event(
                 actor=actor,
                 event=self._serialize_event(event),
@@ -1598,10 +1602,16 @@ class CalendarEventService:
 
         event.delete()
 
+        permission_token = getattr(context.calendar_permission_service, "token", None)
+        if permission_token is not None:
+            audit_actor: Any = _resolve_token_audit_actor(permission_token)
+        else:
+            audit_actor = context.user_or_token
+
         transaction.on_commit(
             lambda: (
                 context.calendar_side_effects_service.on_delete_event(
-                    actor=_resolve_token_audit_actor(context.calendar_permission_service.token),
+                    actor=audit_actor,
                     event=serialized_event,
                     organization=event.organization,
                 )
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -356,8 +356,10 @@ class CalendarEventService:
             ``True`` when the token may proceed; ``False`` when it must be denied.
         """
         if system_user.scoped_to_membership_user_id is None:
-            # Org-wide token — permitted for any event in the org.
-            return True
+            # Org-wide token — permitted for any event in its OWN organization. Re-derive the
+            # tenant boundary from the loaded objects (defense in depth) rather than trusting
+            # the caller-supplied context.organization, mirroring the scoped path.
+            return system_user.organization_id == calendar.organization_id
         # Owner-scoped token — only when the token's owner independently owns the calendar.
         return self._scoped_system_user_owns_calendar(system_user, calendar)
 
@@ -687,6 +689,9 @@ class CalendarEventService:
             # fan-out spans multiple providers and can produce confusing partial
             # failures. Org-wide tokens follow the existing service rules (bundle
             # primary events reach _update_bundle_event below, which handles them).
+            # This guard targets bundle PRIMARY events on the BUNDLE calendar only;
+            # bundle representation/child events (on PERSONAL/RESOURCE calendars with
+            # bundle_primary_event set) are caught by the is_bundle_event ValueError below.
             if (
                 system_user.scoped_to_membership_user_id is not None
                 and event.calendar.calendar_type == CalendarType.BUNDLE
@@ -1530,6 +1535,9 @@ class CalendarEventService:
             # fan-out spans multiple providers and can produce confusing partial
             # failures. Org-wide tokens follow the existing service rules (bundle
             # primary events reach _delete_bundle_event below, which handles them).
+            # This guard targets bundle PRIMARY events on the BUNDLE calendar only;
+            # bundle representation/child events (on PERSONAL/RESOURCE calendars with
+            # bundle_primary_event set) are caught by the is_bundle_event ValueError below.
             if (
                 system_user.scoped_to_membership_user_id is not None
                 and event.calendar.calendar_type == CalendarType.BUNDLE
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -685,20 +685,14 @@ class CalendarEventService:
                 # Surface a not-found-parity message so a cross-owner caller cannot
                 # distinguish a forbidden calendar from a genuinely missing one.
                 raise PermissionDenied("Calendar matching query does not exist.")
-            # Bundle calendars are rejected for scoped (owner) tokens: the bundle
-            # fan-out spans multiple providers and can produce confusing partial
-            # failures. Org-wide tokens follow the existing service rules (bundle
-            # primary events reach _update_bundle_event below, which handles them).
-            # This guard targets bundle PRIMARY events on the BUNDLE calendar only;
-            # bundle representation/child events (on PERSONAL/RESOURCE calendars with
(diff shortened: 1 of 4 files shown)


calendar_integration/services/calendar_event_service.py
```diff
@@ -334,6 +334,35 @@ def _scoped_system_user_owns_calendar(system_user: SystemUser, calendar: Calenda
             .exists()
         )
 
+    def _public_token_may_write(self, system_user: SystemUser, calendar: Calendar) -> bool:
```
hugobessa, line 337: Single choke-point for the update/delete public-token allowance, intentionally
distinct from `create_event`. Per the plan's Guiding Decisions, reschedule/cancel
let **org-wide** tokens act org-wide (allowed here), whereas `create_event` keeps
org-wide blocked. A future feature flag, if ever needed, gates both update and
delete from this one method.

calendar_integration/services/calendar_event_service.py
```diff
@@ -334,6 +334,35 @@ def _scoped_system_user_owns_calendar(system_user: SystemUser, calendar: Calenda
...
+            calendar: The target calendar whose event the caller wants to update/delete.
+
+        Returns:
+            ``True`` when the token may proceed; ``False`` when it must be denied.
+        """
+        if system_user.scoped_to_membership_user_id is None:
+            # Org-wide token — permitted for any event in its OWN organization. Re-derive the
+            # tenant boundary from the loaded objects (defense in depth) rather than trusting
+            # the caller-supplied context.organization, mirroring the scoped path.
+            return system_user.organization_id == calendar.organization_id
```
hugobessa, line 362: Defense-in-depth: the org-wide branch re-derives the tenant boundary from the two
loaded objects (`system_user.organization_id == calendar.organization_id`) instead
of trusting the caller-supplied `context.organization`. This mirrors the scoped
branch (which derives ownership from `CalendarOwnership`), so an org-wide token can
never write across tenants even if a future caller mis-binds the service context.
Added in review.

calendar_integration/services/calendar_event_service.py
```diff
@@ -639,17 +668,41 @@ def update_event(
...
                 organization_id=event.organization_id,
                 event_id=event_id,
             )
         elif isinstance(context.user_or_token, SystemUser):
-            raise PermissionDenied("Events cannot be created through the Public API.")
+            system_user = context.user_or_token
+            if not self._public_token_may_write(system_user, event.calendar):
+                # Surface a not-found-parity message so a cross-owner caller cannot
+                # distinguish a forbidden calendar from a genuinely missing one.
+                raise PermissionDenied("Calendar matching query does not exist.")
```
hugobessa, line 687: Existence-leak invariant: a cross-owner / cross-tenant write raises the **same**
message a genuinely missing calendar produces. The GraphQL layer (Phases 1–3) maps
this to `"Calendar not found."`; keeping the service-layer message identical means a
race that bypasses the GraphQL guard still cannot reveal whether the target exists.
Same raise mirrored in `delete_event`.

calendar_integration/services/calendar_event_service.py
```diff
@@ -639,17 +668,41 @@ def update_event(
...
+                and event.calendar.calendar_type == CalendarType.BUNDLE
+            ):
+                raise PermissionDenied(
+                    "Events cannot be updated on a bundle calendar through the Public API."
+                )
+            is_public_token_write = True
 
         serialized_old_event = self._serialize_event(event)
-        if not context.calendar_permission_service.can_perform_update(
+        if not is_public_token_write and not context.calendar_permission_service.can_perform_update(
```
hugobessa, line 699: `is_public_token_write` bypasses the `can_perform_update` permission-token check for
the sanctioned public path. That check requires an initialized `CalendarManagementToken`
(the booking-code flow), which the `SystemUser` path never has — without this bypass it
would always fail. Authorization for this path is fully covered by `_public_token_may_write`
above. This mirrors how `create_event`'s owner-scoped path bypasses `can_perform_scheduling`.

calendar_integration/services/calendar_event_service.py
```diff
@@ -639,17 +668,41 @@ def update_event(
...
+            # primary events reach _update_bundle_event below, which handles them).
+            # This guard targets bundle PRIMARY events on the BUNDLE calendar only;
+            # bundle representation/child events (on PERSONAL/RESOURCE calendars with
+            # bundle_primary_event set) are caught by the is_bundle_event ValueError below.
+            if (
+                system_user.scoped_to_membership_user_id is not None
+                and event.calendar.calendar_type == CalendarType.BUNDLE
+            ):
+                raise PermissionDenied(
+                    "Events cannot be updated on a bundle calendar through the Public API."
```
hugobessa: Bundle rejection is **scoped-token-only** by design: an owner-scoped token's bundle
fan-out would span other providers' calendars and produce confusing partial failures.
Org-wide tokens deliberately fall through to the existing bundle rules. The predicate
targets bundle *primary* events on the BUNDLE calendar; bundle representation/child
events are handled by the existing downstream `is_bundle_event` ValueError.

calendar_integration/services/calendar_event_service.py
```diff
@@ -969,7 +1022,11 @@ def call_side_effects():
             if not context.calendar_side_effects_service:
                 return
 
-            actor = _resolve_token_audit_actor(context.calendar_permission_service.token)
+            permission_token = getattr(context.calendar_permission_service, "token", None)
+            if permission_token is not None:
+                actor: Any = _resolve_token_audit_actor(permission_token)
+            else:
+                actor = context.user_or_token
```
hugobessa, line 1023: Audit/webhook actor parity with `create_event`: on the public-token path no permission
token is initialized, so the actor falls back to the `SystemUser` caller rather than
`None`. This matters for Phases 1 & 3 — `rescheduleCalendarEvent` / `cancelEvent` fire
these side-effects via public tokens, and an actor-less webhook would lose attribution.
Same fallback applied in `delete_event`'s on-commit callback.

hugobessa:
⚠️ **Force-pushed (amend-plan, 2026-06-23)** — please re-review.

Removed the scoped-token **BUNDLE-calendar rejection** from `update_event` and `delete_event`: bundle events are now reschedulable/cancellable through the Public API. Updating/deleting an existing bundle **primary** event flows to `_update_bundle_event` / `_delete_bundle_event`, gated only by `_public_token_may_write` (owner-scoped must own the bundle calendar; org-wide acts org-wide). This diverges intentionally from `create_event`, which stays bundle-blocked. The two bundle unit tests now assert allowance. Full suite green (3150 passed on this branch). Old SHA eedbff8 → new 6606bcd.

PR #162, opened 2026-06-23
feat(calendar): add single-occurrence reschedule/cancel service methods

hugobessa:
## Summary
Phase 0b of the **Owner-Scoped Public Reschedule / Cancel Mutations** plan — the second service-layer foundation phase. Still no GraphQL surface; this adds the primitives Phases 1 & 3 need for **single-occurrence** reschedule/cancel of a recurring series.

Why this isn't just `update_event`/`delete_event`: an un-materialized occurrence of a recurring series has **no row of its own** (occurrences are expanded on the fly), so it can't be addressed by an `event_id`. And passing a recurring **master** id to `delete_event` with `delete_series=false` would delete the master outright. So a true single-occurrence op must be addressed by `recurrence_id` (the occurrence's original start datetime, `== CalendarEvent.recurrence_id`) and go through the exception primitives.

Adds to `CalendarEventService` (+ `CalendarService` facade delegators):
- `reschedule_event_occurrence(calendar_id, master_event_id, recurrence_id, start_time, end_time, timezone) -> CalendarEvent` — builds a modified-occurrence `CalendarEvent` (`parent_recurring_object=master`, `is_recurring_exception=True`, new times) and links it via `master.create_exception(exception_date=recurrence_id, is_cancelled=False, modified_object=…)`. **Idempotent update-in-place** on a repeated `recurrence_id` (reuses the existing `external_id` via `write_adapter.update_event`, so no external event is orphaned).
- `cancel_event_occurrence(calendar_id, master_event_id, recurrence_id) -> None` — creates a cancellation exception (`is_cancelled=True`, no `modified_event`) without touching the master or its rule.

Authorization reuses Phase 0a's `_public_token_may_write` seam (org-wide **and** owner-scoped tokens; cross-owner/cross-tenant → `PermissionDenied("Calendar matching query does not exist.")`, not-found parity). Deliberately built **without** routing through `create_event` (which blocks org-wide tokens). BUNDLE calendars rejected for scoped tokens (parity with Phase 0a). Non-recurring master → `ValueError`.

## Plan reference
`ai-plans/2026-06-22-OWNER_SCOPED_RESCHEDULE_CANCEL_MUTATIONS_IMPLEMENTATION_PLAN.md` — **Phase 0b** (stacked on Phase 0a / PR #161). No feature flag. No migrations.

## Test plan
[...]

PR #163, opened 2026-06-23
feat(public_api): add rescheduleCalendarEvent mutation

hugobessa:
## Summary
Phase 1 of the **Owner-Scoped Public Reschedule / Cancel Mutations** plan — the first GraphQL surface. Adds the authenticated public mutation `rescheduleCalendarEvent`, a thin owner-scoped wrapper over the Phase 0a/0b service methods.

Three modes via one resolver:
- **Whole event** (`recurrenceId` omitted, `rruleString` omitted): moves start/end/timezone, preserving every non-time field (title, description, attendances, external attendances, resource allocations) AND the recurrence rule.
- **Series with new rule** (`recurrenceId` omitted, `rruleString` set): moves the event and replaces the recurrence pattern.
- **Single occurrence** (`recurrenceId` set): reschedules exactly that occurrence via `CalendarService.reschedule_event_occurrence` (Phase 0b) — a modified-occurrence exception, master untouched.

Owner-scope guard mirrors `scheduleEvent`: an owner-scoped token may only reschedule events on calendars its owner owns (cross-owner → the identical `"Calendar not found."` as a genuinely missing calendar — no existence leak); an org-wide token acts org-wide. Returns `CalendarEventGraphQLType` directly and raises `GraphQLError` on failure (mirrors `scheduleEvent`). `"rescheduleCalendarEvent"` maps to `CALENDAR_EVENT` (already in `PROVIDER_SCOPED_RESOURCES`).

The `*WithCode` reschedule mutations are untouched (negative scope).

## Plan reference
`ai-plans/2026-06-22-OWNER_SCOPED_RESCHEDULE_CANCEL_MUTATIONS_IMPLEMENTATION_PLAN.md` — **Phase 1** (stacked on Phase 0b / PR #162). No feature flag. No migrations. No REST/OpenAPI schema change (GraphQL schema is generated at runtime by Strawberry).

## Test plan
- `docker compose run --rm api uv run pytest public_api/tests/test_mutations.py -k Reschedule -p no:randomly` — `TestScopedTokenRescheduleCalendarEvent` (10): whole-event success + field/attendee/resource/description preservation, series rule preserved (no rrule) and replaced (explicit rrule), single-occurrence creates the exception with master untouched, non-recurring + `recurrenceId` → clean error (no 500), cross-owner not-found parity (no mutation), org-wide acts org-wide, missing event.
- Outer gate: full `pytest -n auto` → **3171 passed**.

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -27,7 +27,7 @@ from calendar_integration.graphql import (
     CalendarEventGraphQLType,
     CalendarGraphQLType,
 )
-from calendar_integration.models import Calendar
+from calendar_integration.models import Calendar, CalendarEvent
 from calendar_integration.mutations import (
     CalendarGroupMutations,
     ExternalEventChangeRequestMutations,
@@ -37,6 +37,7 @@ from calendar_integration.services.dataclasses import (
     EventAttendanceInputData,
     EventExternalAttendanceInputData,
     ExternalAttendeeInputData,
+    ResourceAllocationInputData,
 )
 from organizations.exceptions import NoServiceAccountConfiguredError, UserAlreadyHasMembershipError
 from organizations.models import Organization, OrganizationBranding, OrganizationMembership
@@ -622,6 +623,37 @@ class ScheduleEventInput:
         default_factory=list
     )
     rrule_string: str | None = None
+
+
+@strawberry.input
+class RescheduleCalendarEventInput:
+    """Input for rescheduling a single-calendar event via a public-API token.
+
+    Supports three modes:
+    - **Whole event / series** (``recurrence_id`` omitted): updates the event's time fields
+      and, optionally, its recurrence rule. The existing rule is preserved when
+      ``rrule_string`` is omitted — callers must not strip the series accidentally.
+    - **Series with new rule** (``recurrence_id`` omitted, ``rrule_string`` provided):
+      moves the event AND updates the recurrence pattern.
+    - **Single occurrence** (``recurrence_id`` provided): reschedules exactly the
+      occurrence whose original start equals ``recurrence_id`` without touching the
+      master or any other occurrence.
+
+    An owner-scoped token may only reschedule events on calendars its owner owns;
+    an org-wide token acts org-wide.
+    """
+
+    organization_id: int
+    calendar_id: int
+    event_id: int
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    timezone: str
+    # Optional: change the series' recurrence pattern. Omit to PRESERVE the existing rule.
+    rrule_string: str | None = None
+    # Optional: when set, reschedule ONLY this occurrence of a recurring series
+    # (the occurrence's original start == CalendarEvent.recurrence_id). Omit for whole event/series.
+    recurrence_id: datetime.datetime | None = None
 
 
 @strawberry.type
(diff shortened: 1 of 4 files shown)


public_api/mutations.py
```diff
@@ -2058,3 +2090,141 @@ def schedule_event(
...
+        - Org-wide token: ``assert_calendar_in_owner_scope`` is a no-op → acts org-wide.
+        - The service independently re-verifies ownership as defense-in-depth.
+        """
+        calendar_service, org = _get_org_and_init_calendar_service(info)
+        request: PublicApiHttpRequest = info.context.request
+
+        # Owner-scope guard: a scoped token may only target its owner's calendars.
+        # Cross-owner and missing calendars return the identical error — no existence leak.
+        try:
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, input.calendar_id)
```
hugobessa, line 2132: Owner-scope guard mirrors `scheduleEvent` exactly: `assert_calendar_in_owner_scope`
(no-op for org-wide, restricts owner-scoped) followed by the org-scoped Calendar
load, both mapped to the SAME `"Calendar not found."`. Ordered BEFORE the event
load so a foreign calendar never reaches the event branch and can't be
distinguished from a missing one. The service layer (Phase 0a/0b) independently
re-verifies as defense-in-depth.

public_api/mutations.py
```diff
@@ -2058,3 +2090,141 @@ def schedule_event(
...
+                    if ra.calendar_fk_id
+                ]
+
+                # Recurrence rule preservation: if the caller omits rrule_string, re-pass
+                # the existing rule string so update_event does NOT strip the series.
+                # (update_event deletes the rule when recurrence_rule=None.)
+                if input.rrule_string is not None:
+                    recurrence_rule = input.rrule_string
+                elif existing_event.is_recurring:
+                    recurrence_rule = existing_event.recurrence_rule.to_rrule_string()
```
hugobessa, line 2203: Rule preservation — the core correctness point. `update_event` deletes the
recurrence rule when `recurrence_rule=None`, so to avoid silently turning a series
into a one-off, a no-`rruleString` reschedule re-passes the master's existing rule
via `to_rrule_string()` (verified symmetric with `from_rrule_string`; `update_event`
keeps the existing rule id, so the series is preserved, not recreated). An explicit
`rruleString` replaces the rule; a non-recurring event passes `None`.

public_api/mutations.py
```diff
@@ -2058,3 +2090,141 @@ def schedule_event(
...
+
+        # Load the event — needed for the whole-event/series path (to preserve non-time
+        # fields) and to validate ownership independently of the calendar guard.
+        try:
+            existing_event = (
+                CalendarEvent.objects.filter_by_organization(org.id)
+                .select_related("calendar", "recurrence_rule")
+                .prefetch_related(
+                    "attendances",
+                    "external_attendances__external_attendee",
```
hugobessa, line 2145: When `recurrence_id` is set we route to `reschedule_event_occurrence` and
intentionally ignore the rrule/preserved-fields logic — a single-occurrence
reschedule only moves that one occurrence (a modified exception), never the series.
A non-recurring master here yields a `ValueError` → clean `GraphQLError` (tested),
not a 500.

public_api/permissions.py
```diff
@@ -71,6 +71,7 @@ class OrganizationResourceAccess(BasePermission):
         "updateBlockedTime": PublicAPIResources.UPDATE_BLOCKED_TIME,
         "deleteBlockedTime": PublicAPIResources.DELETE_BLOCKED_TIME,
         "scheduleEvent": PublicAPIResources.CALENDAR_EVENT,
+        "rescheduleCalendarEvent": PublicAPIResources.CALENDAR_EVENT,
```
hugobessa, line 74: Maps the GraphQL field (strawberry camelCases `reschedule_calendar_event` →
`rescheduleCalendarEvent`) to `CALENDAR_EVENT`, reusing the same resource as
`scheduleEvent`. `CALENDAR_EVENT` is already in `PROVIDER_SCOPED_RESOURCES`, so
provider/owner-scoped tokens can be granted it with no constants change. A wrong
key here would silently leave the mutation ungated — covered by a grant-denied test.

PR #164, opened 2026-06-23
feat(public_api): add rescheduleCalendarGroupEvent mutation

hugobessa:
## Summary
Phase 2 of the **Owner-Scoped Public Reschedule / Cancel Mutations** plan. Adds `rescheduleCalendarGroupEvent` — reschedules a grouped event (the primary `CalendarEvent` plus the linked non-primary `BlockedTime` rows, `group-event-{event_id}-cal-{cid}`) to a new time, all moving together. Whole-event only (group events aren't recurring in v1).

The resolver takes only `event_id` (no `calendar_id` — the primary calendar is derived server-side), loads the grouped event org-scoped, owner-scope-checks its **primary** calendar, wires the public-token-initialized `calendar_service` into the group service, and delegates to `CalendarGroupService.reschedule_grouped_event`. Owner-scoped tokens may only reschedule grouped events whose primary calendar their owner owns; org-wide acts org-wide.

### Security: uniform not-found (no existence leak)
Because the input is **event-addressed**, every not-found/forbidden case returns the **identical** `"Event not found."`:
- missing `event_id`
- non-grouped event
- cross-owner grouped event (owner-scope guard)
- cross-owner service-layer race (`PermissionDenied("Calendar matching query does not exist.")` sentinel)

This deliberately supersedes the plan body's per-phase prescription (which named `"Calendar not found."` for the scope failure — correct for Phase 1's calendar-addressed input, but a discriminating oracle here). A scoped token cannot tell "a grouped event I don't own exists" from "doesn't exist."

The `*WithCode` group reschedule mutation is untouched (negative scope).

## Plan reference
`ai-plans/2026-06-22-OWNER_SCOPED_RESCHEDULE_CANCEL_MUTATIONS_IMPLEMENTATION_PLAN.md` — **Phase 2** (stacked on Phase 1 / PR #163). No feature flag. No migrations.

## Test plan
- `docker compose run --rm api uv run pytest public_api/tests/test_mutations.py -k RescheduleCalendarGroup -p no:randomly` — `TestScopedTokenRescheduleCalendarGroupEvent` (6): success (primary event + linked BlockedTimes move), cross-owner uniform not-found (== missing message, no mutation), org-wide acts org-wide, non-grouped → "Event not found.", PermissionDenied-sentinel → "Event not found." (defense-in-depth), service validation error → clean GraphQLError (no 500).
- Outer gate: full `pytest -n auto` → **3177 passed**.

### DIFF
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -2267,8 +2267,8 @@ class Mutation(ExternalEventChangeRequestMutations, CalendarGroupMutations):
 
         Authorization:
         - Owner-scoped token: restricted to grouped events whose primary calendar is
-          owned by the token's owner; cross-owner → ``"Calendar not found."`` (same
-          as missing — no existence leak).
+          owned by the token's owner; cross-owner → ``"Event not found."`` (same
+          as missing event — no existence leak).
         - Org-wide token: acts org-wide.
         - The service independently re-verifies ownership as defense-in-depth.
 
@@ -2276,8 +2276,16 @@ class Mutation(ExternalEventChangeRequestMutations, CalendarGroupMutations):
         """
         from calendar_integration.exceptions import CalendarGroupValidationError
 
+        # Sentinel emitted by CalendarEventService.update_event when a SystemUser's
+        # scoped calendar can't be found (racing cross-owner path). Map it to the
+        # uniform not-found message to prevent a discriminating oracle.
+        calendar_not_found_sentinel = "Calendar matching query does not exist."
+
         calendar_service, org = _get_org_and_init_calendar_service(info)
         request: PublicApiHttpRequest = info.context.request
+
+        # input.organization_id is intentionally ignored: org is derived from the
+        # authenticated request context, not from caller-supplied input.
 
         # Load the grouped event scoped to the organization to derive the primary
         # calendar and validate that it is actually a grouped event.
@@ -2297,13 +2305,15 @@ class Mutation(ExternalEventChangeRequestMutations, CalendarGroupMutations):
         primary_calendar_id: int = grouped_event.calendar_fk_id  # type: ignore[assignment]
 
         # Owner-scope guard on the PRIMARY calendar: a scoped token may only target
-        # grouped events whose primary calendar its owner owns. Cross-owner and missing
-        # calendars return the identical error — no existence leak.
+        # grouped events whose primary calendar its owner owns. The grouped event was
+        # already loaded org-scoped with select_related("calendar"), so the derived
+        # primary calendar provably exists — only the ownership check can fail here.
+        # Map Calendar.DoesNotExist (cross-owner) to "Event not found." — uniform
+        # not-found, no existence leak.
         try:
             assert_calendar_in_owner_scope(request.public_api_system_user, org, primary_calendar_id)
-            Calendar.objects.filter_by_organization(org.id).get(id=primary_calendar_id)
         except Calendar.DoesNotExist as exc:
-            raise GraphQLError("Calendar not found.") from exc
+            raise GraphQLError("Event not found.") from exc
 
         # Wire the group service: inject the already-initialized calendar_service so
         # that the public-token auth context flows into reschedule_grouped_event →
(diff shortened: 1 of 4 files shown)


public_api/mutations.py
```diff
@@ -2228,3 +2249,104 @@ def reschedule_calendar_event(
...
+            # Non-grouped event — uniform not-found, no existence leak.
+            raise GraphQLError("Event not found.")
+
+        primary_calendar_id: int = grouped_event.calendar_fk_id  # type: ignore[assignment]
+
+        # Owner-scope guard on the PRIMARY calendar: a scoped token may only target
+        # grouped events whose primary calendar its owner owns. The grouped event was
+        # already loaded org-scoped with select_related("calendar"), so the derived
+        # primary calendar provably exists — only the ownership check can fail here.
+        # Map Calendar.DoesNotExist (cross-owner) to "Event not found." — uniform
```
hugobessa, line 2311: Existence-leak fix (the headline of this phase). The input is event-addressed
(no calendar_id), so the grouped event must be loaded org-scoped BEFORE the
primary calendar can be scope-checked — which means a scoped token can load an
event it doesn't own. To avoid a discriminating oracle, the cross-owner scope
failure maps to the SAME `"Event not found."` as a missing/non-grouped event,
not `"Calendar not found."`. (Phase 1 keeps `"Calendar not found."` because it
takes calendar_id as the addressing key — different shape, different correct answer.)

public_api/mutations.py
```diff
@@ -2228,3 +2249,104 @@ def reschedule_calendar_event(
...
+        - The service independently re-verifies ownership as defense-in-depth.
+
+        Returns the updated primary ``CalendarEvent``.
+        """
+        from calendar_integration.exceptions import CalendarGroupValidationError
+
+        # Sentinel emitted by CalendarEventService.update_event when a SystemUser's
+        # scoped calendar can't be found (racing cross-owner path). Map it to the
+        # uniform not-found message to prevent a discriminating oracle.
+        calendar_not_found_sentinel = "Calendar matching query does not exist."
```
hugobessa, line 2282: Third leak string closed: the service layer (Phase 0a `update_event`) raises
`PermissionDenied("Calendar matching query does not exist.")` for a racing
cross-owner token. The `except PermissionDenied` branch detects that exact sentinel
and remaps it to `"Event not found."`, so even a race that bypasses the GraphQL
guard cannot leak existence. Other PermissionDenied messages pass through normally.

public_api/mutations.py
```diff
@@ -2228,3 +2249,104 @@ def reschedule_calendar_event(
...
+            assert_calendar_in_owner_scope(request.public_api_system_user, org, primary_calendar_id)
+        except Calendar.DoesNotExist as exc:
+            raise GraphQLError("Event not found.") from exc
+
+        # Wire the group service: inject the already-initialized calendar_service so
+        # that the public-token auth context flows into reschedule_grouped_event →
+        # update_event. This mirrors exactly how rescheduleCalendarGroupEventWithCode
+        # wires its deps (deps.calendar_group_service.calendar_service = deps.calendar_service).
+        group_deps = get_calendar_mutation_dependencies()
+        group_deps.calendar_group_service.calendar_service = calendar_service
```
hugobessa, line 2323: DI wiring: `get_calendar_mutation_dependencies()` returns FRESH `Factory` instances,
so the group service from a second call is NOT the one `_get_org_and_init_calendar_service`
initialized. We inject the already-initialized `calendar_service` into the group
service, then `initialize(org)` — exactly the `*WithCode` group-reschedule wiring —
so the public-token auth context flows into `reschedule_grouped_event` → `update_event`
and Phase 0a's allowance applies. The success test proves this end-to-end.

public_api/permissions.py
```diff
@@ -72,6 +72,7 @@ class OrganizationResourceAccess(BasePermission):
         "deleteBlockedTime": PublicAPIResources.DELETE_BLOCKED_TIME,
         "scheduleEvent": PublicAPIResources.CALENDAR_EVENT,
         "rescheduleCalendarEvent": PublicAPIResources.CALENDAR_EVENT,
+        "rescheduleCalendarGroupEvent": PublicAPIResources.CALENDAR_EVENT,
```
hugobessa, line 75: `rescheduleCalendarGroupEvent` → `CALENDAR_EVENT` (already in `PROVIDER_SCOPED_RESOURCES`),
per the plan's "reuse CALENDAR_EVENT" decision — provider/owner-scoped tokens can be
granted it with no constants change.

PR #166, opened 2026-06-27
feat(calendar): add BookingPolicy model with single-target constraint

hugobessa:
## Summary
Phase 1 of the **Bookable Slots for Single Calendars & Bundles, with Booking Policies** plan — the foundation phase. Pure scaffolding: a new `BookingPolicy` table the later phases consume. No behavior change, no reachable surface yet.

`BookingPolicy(OrganizationModel)` is a flat set of booking guardrails attached to **exactly one** target — a calendar, an owning membership, a calendar group, or the organization default (`is_organization_default=True`). Four `PositiveIntegerField` second-counts (`lead_time_seconds`, `max_horizon_seconds`, `buffer_before_seconds`, `buffer_after_seconds`); **zero = "no constraint for that field"** (lead 0 = bookable now, horizon 0 = unbounded, buffer 0 = flush booking). `PositiveIntegerField` rejects negatives.

Integrity:
- A `bookingpolicy_exactly_one_target` **check constraint** (a 4-way disjunction over the concrete columns) enforces exactly-one-target.
- Per-target **partial unique indexes** (`bookingpolicy_uniq_calendar/membership/group/org_default`) keep resolution unambiguous: at most one policy per calendar / membership / group, and at most one org-default per organization.
- `calendar` + `calendar_group` are real `OrganizationForeignKey`s (concrete FK, `ON DELETE CASCADE`). `membership` is an `OrganizationMembershipForeignKey` (a `ForeignObject` with no DB constraint), so a follow-up migration `0040` adds the raw-SQL composite **PROTECT** FK (`ON DELETE NO ACTION DEFERRABLE INITIALLY DEFERRED`, `NOT VALID` + `VALIDATE`, `atomic=False`), mirroring the `CalendarOwnership`/`EventAttendance`/`CalendarManagementToken`/`ExternalEventChangeRequest` pattern.

Also: `BookingPolicyManager` + `BookingPolicyQuerySet` (org-scoped `for_target` / `for_calendar` / `for_membership` / `for_calendar_group` / `org_default` — the lookups the Phase 2 resolver will call), admin registration, a `create_booking_policy` factory, and 23 unit tests.

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 1**. First phase, branches off `main`. No feature flag (the feature is gated by data-presence: no policy ⇒ unchanged behavior). Migrations: `0039` (CreateModel) + `0040` (membership composite PROTECT FK).

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/test_booking_policy_models.py -p no:randomly` — factory builds valid rows; the check constraint rejects zero-target and multi-target rows; each partial unique index rejects a duplicate (calendar / membership / group / org-default); negatives rejected; the membership composite FK rejects a non-existent membership and blocks deleting a membership with a live policy (committed via `transaction=True`); org-scoped manager lookups return the right row / `None`.
- Outer gate: `makemigrations --check` (clean), `manage.py check --deploy` (0 errors), full `pytest -n auto` (**3244 passed**). Migration round-trips forward → reverse to 0038 → forward.
- Note: the shared compose `db` is contended by concurrent worktree suites and the repo uses `pytest-randomly`; isolated order-flakes (audit admin, calendar views) confirmed pre-existing — pass in isolation and on rerun.


calendar_integration/models.py
```diff
@@ -2029,3 +2030,146 @@ def __str__(self) -> str:
...
+            ),
+        ]
+        constraints: ClassVar = [
+            # Exactly one target set. Each of the four targets is mutually
+            # exclusive: the only valid rows are those matching exactly one of
+            # the four single-target shapes (the other three columns NULL /
+            # False). Expressed as an OR of the four mutually-exclusive shapes so
+            # the constraint reads as a disjunction Postgres can evaluate per row.
+            models.CheckConstraint(
+                name="bookingpolicy_exactly_one_target",
```
hugobessa, line 2111: Exactly-one-target is enforced as an explicit **disjunction of the four
single-target shapes** (each asserts one column set and the other three
NULL/false), not an integer-sum-of-booleans. Reasons: it compiles to plain
`(... OR ... OR ... OR ...)` SQL Postgres evaluates per row, needs no
`Case/When/Cast`, and there was no existing `CheckConstraint` in the repo to
mirror. It rejects both the all-empty row and any two-targets-set row.

calendar_integration/models.py
```diff
@@ -2029,3 +2030,146 @@ def __str__(self) -> str:
...
+        CalendarGroup,
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="booking_policies",
+    )
+    is_organization_default = models.BooleanField(default=False)
+
+    lead_time_seconds = models.PositiveIntegerField(default=0)
+    max_horizon_seconds = models.PositiveIntegerField(default=0)  # 0 = unbounded
```
hugobessa, line 2083: `max_horizon_seconds = 0` means **unbounded**, NOT "cannot book" — consistent
with the other three fields where 0 = "no constraint". The Phase 2 resolver
will read 0 here as `None`/unbounded. Kept as a plain `PositiveIntegerField`
(no nullable column) so all four guardrails share one encoding.

calendar_integration/models.py
```diff
@@ -2029,3 +2030,146 @@ def __str__(self) -> str:
...
+
+    calendar = OrganizationForeignKey(
+        Calendar,
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="booking_policies",
+    )
+    membership = OrganizationMembershipForeignKey(
+        on_delete=models.PROTECT,
```
hugobessa, line 2068: Deliberate asymmetry: `membership` is `PROTECT`, while `calendar` and
`calendar_group` are `CASCADE`. The `ForeignObject` ignores this kwarg at the
DB level (it's documentary), so the real enforcement is migration 0040's
composite FK (`ON DELETE NO ACTION` = PROTECT-at-commit), matching every other
membership reference in the repo. Net effect: deleting a calendar/group
removes its policy; deleting a membership with a live policy is blocked until
the policy is removed (or both go in one org-cascade transaction).

calendar_integration/migrations/0040_bookingpolicy_membership_protect_fk.py
```diff
@@ -0,0 +1,118 @@
...
+DROP_CONSTRAINT = f"""
+ALTER TABLE calendar_integration_bookingpolicy
+  DROP CONSTRAINT IF EXISTS {CONSTRAINT_NAME};
+"""
+
+
+class Migration(migrations.Migration):
+    """Add the raw-SQL composite PROTECT FK on BookingPolicy.membership."""
+
+    atomic = False
```
hugobessa, line 100: Why `DEFERRABLE INITIALLY DEFERRED` rather than a plain `RESTRICT`: Django's
Python cascade collector is blind to the `ForeignObject` (it carries no DB
constraint Django can see), so on `Organization` delete it removes the
membership and the policy in an arbitrary order. A deferred check fires at
COMMIT, so a same-transaction cascade that drops both succeeds, while a
membership-only delete (policy still live) still fails. `NOT VALID` + separate
`VALIDATE` keeps locks brief; `atomic=False` lets them run in separate txns.

calendar_integration/managers.py
```diff
@@ -337,3 +338,61 @@ def resolvable_by(
...
+    A policy is attached to exactly one target (calendar / membership / calendar
+    group / organization default); these helpers return the single matching row
+    (or ``None``) for a given target, all scoped through the inherited
+    organization filter.
+    """
+
+    def get_queryset(self) -> BookingPolicyQuerySet:
+        return BookingPolicyQuerySet(self.model, using=self._db)
+
+    def for_target(
```
hugobessa, line 355: `for_target` and `org_default` take a required `organization_id` and chain
`filter_by_organization(...)` before the per-target filter. This is
load-bearing: without it, `.first()` trips the tenant-scope guard
(`_check_required_tenant_filter`) and raises `ImproperlyConfigured` at runtime.
The Phase 2 resolver consumes exactly these methods, so they must be org-scoped
and callable — covered by the new manager-level tests.

PR #167, opened 2026-06-28
feat(calendar): add EffectivePolicy resolver + BookingPolicy write service

hugobessa:
## Summary
Phase 2 of the **Bookable Slots for Single Calendars & Bundles, with Booking Policies** plan — the resolver the slot surfaces (Phase 5/6/7) and the booking-time enforcement (Phase 8a/8b) all consume. Stacked on Phase 1 (the `BookingPolicy` model). No GraphQL/REST surface yet, no migration.

Adds:
- **`EffectivePolicy`** (frozen dataclass in `calendar_integration/services/dataclasses.py`) — `lead_time`, `max_horizon: timedelta | None` (None = unbounded), `buffer_before`, `buffer_after`. `from_model` maps a stored `max_horizon_seconds=0` to `None`; `most_restrictive` combines `max(lead)`, `min(finite horizons)`, `max(buffer_*)`; empty → `unconstrained()`.
- **`BookingPolicyService`** (`booking_policy_service.py`):
  - `resolve_for_calendar`: calendar → owning-membership (lone owner, else `is_default` owner when several, else skip) → org-default → unconstrained.
  - `resolve_for_bundle` / `resolve_for_group`: explicit policy on the bundle calendar / group → `most_restrictive` across all participants (each resolved via `resolve_for_calendar`) → unconstrained.
  - `create` / `update` / `delete` (+ per-target convenience deleters and `get_all_policies`) for Phases 3/4: exactly-one-target validation, create-uniqueness (`DuplicateBookingPolicyError`), immutable targets on update, idempotent delete-absent, and `AuditService` records (diffs on update).
- DI wiring: `booking_policy_service` Factory in `di_core/containers.py` (injects `audit_service`).

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 2**. Stacked on Phase 1 / PR #166. No feature flag (data-presence gate). No migration.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_booking_policy_service.py -p no:randomly` — full precedence matrix (calendar / membership / org-default present & absent, fallthrough to unconstrained); owning-membership ambiguity (zero → skip, one → use, multiple → `is_default`); `most_restrictive` field-by-field incl. None-horizon; bundle/group explicit-override beats per-participant combination; bundle inherits org-default via participants; bundle unconstrained when children have owners but no policies; create-uniqueness rejection; update diff (and `diff is None` on no-change); delete-absent no-op for every target; audit emission. **63 passed.**
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), full `pytest -n auto` → **3307 passed**.

### DIFF
--- a/calendar_integration/querysets.py
+++ b/calendar_integration/querysets.py
@@ -436,7 +436,7 @@ def _owning_membership_uid_expression(calendar_id_ref: OuterRef, org_id_ref: Out
 
 
 def _winning_calendar_policy_id_expression(
-    calendar_id_ref: OuterRef, org_id_ref: OuterRef, membership_uid_ref
+    calendar_id_ref: OuterRef, org_id_ref: OuterRef, membership_uid_ref: OuterRef
 ) -> Coalesce:
     """Resolve, in SQL, the id of the BookingPolicy that governs a single calendar.
 
--- a/calendar_integration/querysets.py
+++ b/calendar_integration/querysets.py
@@ -2,7 +2,22 @@ import datetime
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
-from django.db.models import Count, Exists, F, OuterRef, Prefetch, Q, Subquery
+from django.db.models import (
+    Case,
+    Count,
+    Exists,
+    F,
+    IntegerField,
+    Max,
+    Min,
+    OuterRef,
+    Prefetch,
+    Q,
+    Subquery,
+    Value,
+    When,
+)
+from django.db.models.functions import Coalesce
 from django.utils import timezone
 
 from calendar_integration.constants import (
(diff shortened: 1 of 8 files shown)


calendar_integration/services/booking_policy_service.py
```diff
@@ -0,0 +1,494 @@
...
+        3. ``EffectivePolicy.unconstrained()``.
+        """
+        self._assert_initialized()
+
+        # 1. Explicit bundle-calendar policy.
+        bundle_policy = self._policy_for_calendar(bundle_calendar.id)
+        if bundle_policy is not None:
+            return EffectivePolicy.from_model(bundle_policy)
+
+        # 2. Most-restrictive across all child calendars.
```
hugobessa: Subtle, intended coupling that Phases 5/7 rely on: bundle/group resolution
combines participants via `resolve_for_calendar`, which itself falls through
to the **org-default** policy. So when no explicit bundle/group/calendar/
membership policy exists but an org-default does, the bundle/group inherits
the org-default (most-restrictive of N identical org-defaults = the
org-default). Consequence: a bundle/group is never truly unconstrained while
an org-default exists. This matches the plan's "never offer a slot a
participant would reject" and is asserted by
`test_child_inherits_org_default_if_no_direct_policy`.

calendar_integration/services/booking_policy_service.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def _org_default_policy(self) -> BookingPolicy | None:
+        """Return the organization-default policy, or ``None``."""
+        return BookingPolicy.objects.org_default(self.organization.id)  # type: ignore[union-attr]
+
+    # ------------------------------------------------------------------
+    # Owning-membership resolution
+    # ------------------------------------------------------------------
+
+    def _resolve_owning_membership_user_id(self, calendar: Calendar) -> int | None:
```
hugobessa, line 163: Owning-membership ambiguity rule (from Guiding Decisions): exactly one
non-NULL `membership_user_id` ownership → use it; multiple → the
`is_default=True` one; multiple-with-no-default OR zero → return None (skip
the membership layer — resource / shared calendars). Orphan ownerships
(`membership_user_id IS NULL`) are excluded from the count, so they never
short-circuit the lone-owner branch.

calendar_integration/services/dataclasses.py
```diff
@@ -345,3 +349,86 @@ class BookableSlotProposal:
...
+        """
+        policy_list = list(policies)
+        if not policy_list:
+            return EffectivePolicy.unconstrained()
+
+        max_lead = max(p.lead_time for p in policy_list)
+        max_buffer_before = max(p.buffer_before for p in policy_list)
+        max_buffer_after = max(p.buffer_after for p in policy_list)
+
+        # Finite horizons only; None (unbounded) acts as +∞ and is skipped.
```
hugobessa, line 470: `None` horizon = unbounded = +∞, so it's excluded from the `min`; the result
is `None` only when EVERY input horizon is None. A single finite horizon
among unbounded participants wins. lead/buffers take `max` (strictest). This
is what makes a bundle/group "as strict as its strictest participant".

calendar_integration/services/booking_policy_service.py
```diff
@@ -0,0 +1,494 @@
...
+            is_organization_default=is_organization_default,
+            lead_time_seconds=lead_time_seconds,
+            max_horizon_seconds=max_horizon_seconds,
+            buffer_before_seconds=buffer_before_seconds,
+            buffer_after_seconds=buffer_after_seconds,
+        )
+        self._audit_write(AuditAction.CREATE, policy)
+        return policy
+
+    def update_booking_policy(
```
hugobessa, line 371: Targets (calendar / membership / calendar_group / is_organization_default)
are intentionally NOT updatable — only the four rule-fields are. Changing a
target means delete + re-create, which keeps the per-target partial unique
indexes and the exactly-one-target check constraint unambiguous. The audit
UPDATE diff therefore only ever covers rule-field changes.

calendar_integration/querysets.py
```diff
@@ -344,6 +359,154 @@ def _only_calendars_available_in_ranges(
 
         return self.filter(combined_query)
 
+    def annotate_effective_policy(self) -> "CalendarQuerySet":
```
hugobessa, line 362: Whole-policy precedence in SQL: the winning BookingPolicy **id** is resolved
first (COALESCE of the calendar-level, owning-membership, and org-default
subqueries), then all four `effective_*_seconds` columns are read from THAT
single policy. This deliberately avoids a per-field COALESCE across layers —
a winning policy with `max_horizon_seconds=0` stays unbounded rather than
falling through to a lower layer's horizon, matching the original Python
`resolve_for_calendar` exactly. Every inner subquery is org-scoped by the
row's own `organization_id`, so no cross-tenant policy can leak in.

calendar_integration/querysets.py
```diff
@@ -344,6 +359,154 @@ def _only_calendars_available_in_ranges(
...
+            )
+        )
+        return qs.annotate(
+            **_policy_field_subqueries(
+                OuterRef("_effective_policy_id"), OuterRef("organization_id")
+            )
+        )
+
+
+def _owning_membership_uid_expression(calendar_id_ref: OuterRef, org_id_ref: OuterRef) -> Coalesce:
```
hugobessa, line 394: The owning-membership tiebreak reproduced in SQL: `COALESCE(default_owner_uid,
lone_owner_uid)` where `lone_owner_uid` is gated by `HAVING COUNT(*) = 1` over
non-NULL `membership_user_id` ownerships. This yields exactly the Python rule —
0 owners → NULL (skip the membership layer), 1 → that owner, multiple-with-a-
default → the `is_default=True` one, **multiple-with-no-default → NULL (skip)**,
never an arbitrary pick. Orphan ownerships (`membership_user_id IS NULL`) are
excluded from the count so they can't short-circuit the lone-owner branch.

calendar_integration/querysets.py
```diff
@@ -501,6 +664,131 @@ def _only_groups_bookable_in_ranges(
 
         return qs
 
+    def annotate_effective_policy(self) -> "CalendarGroupQuerySet":
```
hugobessa, line 667: Group resolution: an explicit group policy short-circuits via `Case/When`;
otherwise the `most_restrictive` aggregate runs over the distinct participant
calendars (traversed by `slot_fk__group_fk_id`), each resolved through the same
single-calendar chain — `MAX(lead/buffer_before/buffer_after)` and `MIN` over
**positive** horizons only (0/unbounded excluded, so all-unbounded → NULL →
unbounded, any finite binds). MAX/MIN make a calendar that appears in multiple
slots idempotent, matching the Python `.distinct()`.

calendar_integration/services/booking_policy_service.py
```diff
@@ -0,0 +1,473 @@
...
+        self._assert_initialized()
+
+        row = (
+            Calendar.objects.filter_by_organization(self.organization.id)  # type: ignore[union-attr]
+            .annotate_effective_policy()
+            .get(pk=calendar.id)
+        )
+        return EffectivePolicy.from_annotation(row)
+
+    def resolve_for_bundle(self, bundle_calendar: Calendar) -> EffectivePolicy:
```
hugobessa, line 228: `resolve_for_bundle` keeps its explicit-bundle-policy short-circuit (a policy
attached to the bundle calendar is a calendar-level policy and wins), then
annotates ALL children in a SINGLE query and combines them via
`EffectivePolicy.most_restrictive`. This is the N+1 the amendment removed —
the old path issued one resolution query per child. The query-count test locks
in that bundle/group resolution no longer scales with participant count.

calendar_integration/services/dataclasses.py
```diff
@@ -345,3 +364,116 @@ class BookableSlotProposal:
...
+                datetime.timedelta(seconds=policy.max_horizon_seconds)
+                if policy.max_horizon_seconds > 0
+                else None
+            ),
+            buffer_before=datetime.timedelta(seconds=policy.buffer_before_seconds),
+            buffer_after=datetime.timedelta(seconds=policy.buffer_after_seconds),
+        )
+
+    @classmethod
+    def from_annotation(cls, row: "_EffectivePolicyRow") -> "EffectivePolicy":
```
hugobessa, line 419: `from_annotation` is the only decode step left in Python: the SQL annotation
resolved the entire chain, so this just maps the four columns into the dataclass.
A `0` or `NULL` horizon → `max_horizon=None` (unbounded, mirroring `from_model`);
`0`/`NULL` lead and buffers → `timedelta(0)`. When no policy resolved every
column is NULL → `unconstrained()`. Typed against the `_EffectivePolicyRow`
Protocol so any annotated `Calendar`/`CalendarGroup` row satisfies it.

PR #168, opened 2026-06-28
feat(calendar): add BookingPolicy REST CRUD endpoint

hugobessa:
## Summary
Phase 3 of the **Bookable Slots for Single Calendars & Bundles, with Booking Policies** plan — the private REST CRUD surface for `BookingPolicy` (spec use-case 3, REST half). Stacked on Phase 2 (the resolver/write service). No migration; `schema.yml` regenerated.

Adds:
- **`BookingPolicySerializer`** (`serializers.py`) — exactly-one-target validation on create, immutable targets on update (stripped), org-scoped `calendar`/`calendar_group` querysets, `validate_membership_user_id` (rejects bogus/cross-org ids with 400 instead of a DB 500), `min_value=0` on the four rule fields, and `DuplicateBookingPolicyError → 400`. Writes delegate to `BookingPolicyService`.
- **`BookingPolicyViewSet(VintaScheduleModelViewSet)`** (`views.py`) — org-scoped `get_queryset`; `create`/`update`/`destroy` resolve the DI `booking_policy_service`, `initialize(org)` + `set_actor(...)`, then delegate; idempotent `destroy` (204 even when absent).
- **`BookingPolicyPermission`** (`permissions.py`) — reads = any authenticated member; **writes = organization admin** (mirrors `IsOrganizationAdmin`), per spec use-case 3.
- Route `booking-policies` in `routes.py`; regenerated `schema.yml`.

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 3**. Stacked on Phase 2 / PR #167. No feature flag (data-presence gate). No migration.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/test_booking_policy_api.py -p no:randomly` — list/retrieve; create for all four target types; duplicate-target → 400; exactly-one-target (zero/multiple) → 400; negative on all four rule fields → 400; cross-org isolation (retrieve/update 404, delete 204 no-op); delete-absent → 204; **member write → 403, admin write → 2xx**; bogus `membership_user_id` → 400; audit emission on create/update/delete. **35 tests.**
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), `schema.yml` drift-free, full `pytest -n auto` → **3342 passed**.

### DIFF
--- a/calendar_integration/permissions.py
+++ b/calendar_integration/permissions.py
@@ -6,6 +6,29 @@ from calendar_integration.services.calendar_permission_service import CalendarPe
 from organizations.models import get_active_organization_membership
 
 
+class BookingPolicyPermission(BasePermission):
+    """Permission for ``BookingPolicyViewSet``.
+
+    Requires an authenticated user.  Membership-less (gated) users are allowed
+    through: ``get_queryset()`` returns an empty queryset for them so the list
+    action returns 200+[] rather than 403, which is the consistent pattern used
+    by ``CalendarEventViewSet`` and ``BlockedTimeViewSet``.
+
+    Write operations (create, update, destroy) that reach a membership-less user
+    will fail gracefully because ``_build_service()`` gets ``None`` from
+    ``get_active_organization_membership`` and will not initialize the service
+    with a tenant — but in practice the ``TenantScopedViewMixin`` will have
+    already returned 400/403 if the user has no org context.
+
+    Org-admin gating is intentionally **not** applied here — the plan does not
+    restrict policy management to admins only.
+    """
+
+    def has_permission(self, request, view) -> bool:
+        """Allow any authenticated user through (membership check happens in get_queryset)."""
+        return bool(request.user.is_authenticated)
+
+
 class ExternalEventChangeRequestPermission(BasePermission):
     """Permission for ``ExternalEventChangeRequestViewSet``.
 
--- a/calendar_integration/permissions.py
+++ b/calendar_integration/permissions.py
@@ -1,5 +1,5 @@
 from dependency_injector.wiring import Provide, inject
-from rest_framework.permissions import BasePermission
+from rest_framework.permissions import SAFE_METHODS, BasePermission
 
 from calendar_integration.models import CalendarOwnership
 from calendar_integration.services.calendar_permission_service import CalendarPermissionService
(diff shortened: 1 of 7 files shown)


calendar_integration/permissions.py
```diff
@@ -1,11 +1,36 @@
...
+    Membership-less (gated) users are allowed through on safe methods: the
+    queryset returns [] rather than 403, which is the consistent pattern used by
+    ``CalendarEventViewSet`` and ``BlockedTimeViewSet``.
+    """
+
+    def has_permission(self, request, view) -> bool:
+        """Safe methods: any authenticated user. Unsafe methods: org admin only."""
+        if not request.user or not request.user.is_authenticated:
+            return False
+        if request.method in SAFE_METHODS:
```
hugobessa, line 28: Reads (SAFE_METHODS) are visible to any authenticated member — the org-scoped
`get_queryset` already restricts what they see. **Writes require an org
admin** (`membership.is_admin`, mirroring `IsOrganizationAdmin`), because
booking policies include org-wide config (the org-default and group
policies) and SPEC use-case 3's actor is the organization admin. A non-admin
member write returns 403.

calendar_integration/serializers.py
```diff
@@ -2776,6 +2782,207 @@ class BookableSlotProposalSerializer(serializers.Serializer):
...
+            queryset=(
+                CalendarGroup.objects.filter_by_organization(org_id)
+                if org_id is not None
+                else CalendarGroup.original_manager.none()
+            ),
+            required=False,
+            allow_null=True,
+        )
+
+    def validate_membership_user_id(self, value):
```
hugobessa, line 2871: `membership_user_id` is a denormalized int, not a model FK, so DRF won't
validate it. Without this check a bogus / cross-org id would only fail at
COMMIT against Phase 1's composite PROTECT FK → HTTP 500. We validate
existence (org-scoped) here so the client gets a clean 400. calendar /
calendar_group are already org-scoped via the `__init__` querysets.

calendar_integration/views.py
```diff
@@ -1997,6 +2002,179 @@ def bookable_slots(
...
+    @extend_schema(
+        summary="Delete a booking policy",
+        description=(
+            "Delete a booking policy by id. "
+            "Returns 204 even when the policy does not exist (idempotent no-op)."
+        ),
+        responses={204: None},
+    )
+    @inject
+    def destroy(
```
hugobessa, line 2142: Idempotent destroy: resolve the policy org-scoped but DON'T 404 when absent —
pass the row (or None) to `delete_booking_policy`, which no-ops on None and
emits no spurious audit DELETE. Returns 204 either way, matching the plan's
delete-absent = idempotent-no-op-success contract (also why a cross-org pk is
a 204, not a 404 — it's simply not in this org's queryset).

calendar_integration/serializers.py
```diff
@@ -2776,6 +2782,207 @@ class BookableSlotProposalSerializer(serializers.Serializer):
...
+                    "non_field_errors": [
+                        "Exactly one of 'calendar', 'membership_user_id', 'calendar_group', "
+                        "or 'is_organization_default' must be set."
+                    ]
+                }
+            )
+
+        return attrs
+
+    def create(self, validated_data: dict) -> BookingPolicy:
```
hugobessa, line 2945: create/update delegate to `BookingPolicyService` (injected via serializer
context by the viewset) so validation, uniqueness, and audit all live in one
place. `DuplicateBookingPolicyError` (raised by the service on a second
policy for the same target) is caught and re-raised as a DRF ValidationError
→ 400 with the conflict named, rather than surfacing the DB partial-unique
index error.

PR #169, opened 2026-06-28
feat(public_api): expose BookingPolicy CRUD on the public GraphQL API

hugobessa:
## Summary
Phase 4 of the **Bookable Slots for Single Calendars & Bundles, with Booking Policies** plan — the public GraphQL CRUD surface for `BookingPolicy` (spec use-case 3, GraphQL half). Stacked on Phase 3 (REST CRUD). No migration.

Adds:
- **`BookingPolicyGraphQLType`** + `CreateBookingPolicyInput` / `UpdateBookingPolicyInput` / `DeleteBookingPolicyInput` / `BookingPolicyResult` / `DeleteBookingPolicyResult` (`calendar_integration/graphql.py`).
- **`booking_policies`** query (`public_api/queries.py`) — org-scoped via `BookingPolicyService.get_all_policies()`, filterable by target.
- **`create_booking_policy` / `update_booking_policy` / `delete_booking_policy`** mutations (`public_api/mutations.py`) — delegate to the shared `BookingPolicyService` (`set_actor(actor_from_system_user)`), resolve targets org-scoped, map `DuplicateBookingPolicyError`/`ValueError` → `GraphQLError`, idempotent delete.
- **`PublicAPIResources.BOOKING_POLICY`** + the four `FIELD_TO_RESOURCE_MAPPING` entries (`public_api/constants.py`, `public_api/permissions.py`) — a SystemUser token must hold the `BOOKING_POLICY` resource.

Parity fix (review): membership-existence validation was moved into the **shared service** `create_booking_policy` (raising `ValueError`), so GraphQL and REST behave identically — a bogus/cross-org `membership_user_id` now returns a clean error on both surfaces instead of a DB 500 on GraphQL.

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 4**. Stacked on Phase 3 / PR #168. No feature flag (data-presence gate). No migration.

## Test plan
- `docker compose run --rm api uv run pytest public_api/tests/test_booking_policy_graphql.py -p no:randomly` — authorized CRUD per target; missing-resource → permission error; cross-org isolation; duplicate-target (calendar / org-default / membership / group) rejection; exactly-one-target (zero / multiple) rejection; bogus & cross-org `membershipUserId` → clean error (not 500); idempotent delete; audit emission. Plus a service-level unit test for the membership `ValueError`.
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), `schema.yml` drift-free, full `pytest -n auto` → **3377 passed**.

### DIFF
--- a/calendar_integration/services/booking_policy_service.py
+++ b/calendar_integration/services/booking_policy_service.py
@@ -39,7 +39,7 @@ from calendar_integration.models import (
     ChildrenCalendarRelationship,
 )
 from calendar_integration.services.dataclasses import EffectivePolicy
-from organizations.models import Organization
+from organizations.models import Organization, OrganizationMembership
 
 
 if TYPE_CHECKING:
@@ -329,6 +329,13 @@ class BookingPolicyService:
 
         org_id = self.organization.id  # type: ignore[union-attr]
         org = self.organization  # type: ignore[assignment]
+
+        # Validate that membership_user_id references a real membership in the bound org.
+        if membership_user_id is not None:
+            if not OrganizationMembership.objects.filter(
+                organization_id=org_id, user_id=membership_user_id
+            ).exists():
+                raise ValueError("No membership with this user id in your organization.")
 
         # Check uniqueness before hitting the DB constraint to provide a clear error.
         if calendar is not None and self._policy_for_calendar(calendar.id) is not None:
--- a/public_api/mutations.py
+++ b/public_api/mutations.py
@@ -18,16 +18,22 @@ from graphql import GraphQLError
 from calendar_integration.constants import CalendarType
 from calendar_integration.exceptions import (
     CalendarIntegrationError,
+    DuplicateBookingPolicyError,
     NoAvailableTimeWindowsError,
 )
 from calendar_integration.graphql import (
     AvailableTimeGraphQLType,
     BlockedTimeGraphQLType,
+    BookingPolicyResult,
     CalendarBundleGraphQLType,
     CalendarEventGraphQLType,
     CalendarGraphQLType,
+    CreateBookingPolicyInput,
+    DeleteBookingPolicyInput,
+    DeleteBookingPolicyResult,
+    UpdateBookingPolicyInput,
 )
-from calendar_integration.models import Calendar, CalendarEvent
+from calendar_integration.models import BookingPolicy, Calendar, CalendarEvent, CalendarGroup
 from calendar_integration.mutations import (
     CalendarGroupMutations,
     ExternalEventChangeRequestMutations,
@@ -74,6 +80,7 @@ if TYPE_CHECKING:
 
 
 if TYPE_CHECKING:
+    from calendar_integration.services.booking_policy_service import BookingPolicyService
     from calendar_integration.services.calendar_group_service import CalendarGroupService
     from calendar_integration.services.calendar_service import CalendarService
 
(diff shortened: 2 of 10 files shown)


calendar_integration/services/booking_policy_service.py
```diff
@@ -351,6 +351,13 @@ def create_booking_policy(
         org_id = self.organization.id  # type: ignore[union-attr]
         org = self.organization  # type: ignore[assignment]
 
+        # Validate that membership_user_id references a real membership in the bound org.
+        if membership_user_id is not None:
+            if not OrganizationMembership.objects.filter(
+                organization_id=org_id, user_id=membership_user_id
+            ).exists():
+                raise ValueError("No membership with this user id in your organization.")
```
hugobessa, line 338: Membership-existence validation lives in the SHARED service (not the REST
serializer alone) so both surfaces get it in one place. `membership_user_id`
is a denormalized int, not a model FK — without this check a bogus/cross-org
id would only fail at COMMIT against Phase 1's composite PROTECT FK → 500.
Raising `ValueError` here means the GraphQL mutation's existing
`except ValueError → GraphQLError` returns a clean error, matching REST.

public_api/mutations.py
```diff
@@ -2600,3 +2619,149 @@ def cancel_event(
             raise GraphQLError(str(exc)) from exc
 
         return CancelEventResult(success=True)
+
+    # ------------------------------------------------------------------
+    # BookingPolicy mutations (Phase 4)
+    # ------------------------------------------------------------------
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def create_booking_policy(
```
hugobessa, line 2628: Targets are resolved org-scoped (`Calendar`/`CalendarGroup`/`BookingPolicy`
via `filter_by_organization(org.id)`), so a cross-org id is a clean
not-found GraphQLError — never a leak and never a 500. Org comes from the
authenticated SystemUser (`public_api_organization`); the audit actor is
`actor_from_system_user`. `DuplicateBookingPolicyError` and the
exactly-one-target / bad-membership `ValueError` both map to `GraphQLError`.

public_api/permissions.py
```diff
@@ -87,6 +87,10 @@ class OrganizationResourceAccess(BasePermission):
         "externalEventChangeRequests": PublicAPIResources.EXTERNAL_EVENT_CHANGE_REQUEST,
         "approveExternalEventChangeRequest": PublicAPIResources.EXTERNAL_EVENT_CHANGE_REQUEST,
         "rejectExternalEventChangeRequest": PublicAPIResources.EXTERNAL_EVENT_CHANGE_REQUEST,
+        "bookingPolicies": PublicAPIResources.BOOKING_POLICY,
```
hugobessa, line 90: All four GraphQL fields map to `PublicAPIResources.BOOKING_POLICY` (camelCase
keys, matching strawberry's field naming). A SystemUser token without that
resource is rejected by `OrganizationResourceAccess` before the resolver
runs — this is the public-API equivalent of Phase 3's admin gate (public API
is token-resource-scoped, not session-role-scoped).

public_api/queries.py
```diff
@@ -1065,6 +1079,44 @@ def external_event_change_requests(
             list(_slice_qs(qs, offset, limit)),
         )
 
+    # ------------------------------------------------------------------
+    # BookingPolicy queries (Phase 4)
+    # ------------------------------------------------------------------
+
+    @strawberry_django.field(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def booking_policies(
```
hugobessa, line 1087: `booking_policies` is org-scoped through `get_all_policies()` (which filters
by the bound org) plus the optional target filters — there is no path to read
another org's policies. Pagination via offset/limit on a stable `pk` order.

PR #170, opened 2026-06-28
feat(calendar): add policy-aware calendar_bookable_slots query (single + bundle)

hugobessa:
## Summary
Phase 5 — the keystone read surface. A new `calendar_bookable_slots` GraphQL query returns policy-compliant discretized slots for any single calendar id, auto-expanding bundles. Stacked on Phase 4. No migration. Spec use-cases #1 (single) + #2 (bundle).

- **`slot_engine`** (new `calendar_integration/services/slot_engine.py`) — the group walker's primitives extracted into pure, org-scoped, module-level functions (`split_calendars_by_management`, `fetch_available_spans`, `fetch_blocking_spans`, `calendar_free_for_window`, `intervals_overlap`) **plus** `apply_policy_filter`. `CalendarGroupService.find_bookable_slots` now delegates to these — a **pure refactor** (the group query's output, incl. `required_count` per-slot counting, is unchanged; existing group tests stay green).
- **`BookableSlotsService.find_bookable_slots_for_calendar`** (new `bookable_slots_service.py`) — detects personal vs bundle (`calendar_type`); for bundles requires **every** `bundle_children` calendar free; resolves the `EffectivePolicy` (`resolve_for_calendar` / `resolve_for_bundle`); walks candidates then applies the policy filter. DI-wired as `bookable_slots_service`.
- **`calendar_bookable_slots`** query (`public_api/queries.py`) + `PublicAPIResources.BOOKABLE_SLOTS` + mapping; `[IsAuthenticated, OrganizationResourceAccess]`; owner-scope check before resolving.

**Buffer semantics (review fix):** the buffer rule is the **event-envelope / dead-zone-around-the-event** model — a candidate `[start, end]` is dropped if it overlaps a blocking span expanded to `[span_start − buffer_before, span_end + buffer_after]`. (The original candidate-envelope phrasing in the spec/plan was inverted vs acceptance scenario #4 and was corrected in both the code and the contract docs in this PR.)

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 5**. Stacked on Phase 4 / PR #169. No feature flag (data-presence gate). No migration. The spec + plan buffer-rule wording was corrected in this PR.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_bookable_slots_service.py public_api/tests/test_bookable_slots_query.py -p no:randomly` — single managed/unmanaged == one-calendar-group; bundle two-free yields slot + busy-child suppresses; empty window / step≥window / empty bundle → []; lead / horizon / buffer-before / buffer-after each drop the right candidates; **scenario #4 asymmetric (before 10m / after 20m, event 14:00–15:00 → first post-event slot 15:20)**; no-policy == `CalendarGroupService` reference (byte-for-byte); GraphQL auth + resource-gating + owner-scope.
- Group regression: `test_calendar_group_service.py` 44 passed; group query tests green (refactor faithful).
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), full `pytest -n auto` → **3398 passed**, mypy no new errors.


calendar_integration/services/slot_engine.py
```diff
@@ -0,0 +1,264 @@
...
+        if horizon_cutoff is not None and proposal.start_time > horizon_cutoff:
+            continue
+        if has_buffer:
+            # Event-envelope: expand each blocking span to its dead zone
+            # [bs - buffer_before, be + buffer_after) and test the BARE candidate.
+            candidate = (proposal.start_time, proposal.end_time)
+            if any(
+                intervals_overlap(
+                    candidate,
+                    (bs - policy.buffer_before, be + policy.buffer_after),
```
hugobessa, line 258: Buffer = dead-zone-AROUND-THE-EVENT (event-envelope model). A candidate
`[start, end]` is dropped iff it overlaps a blocking span expanded to
`[bs - buffer_before, be + buffer_after]`. This was a review fix: the original
code (and the spec/plan wording, also corrected in this PR) expanded the
CANDIDATE instead, which silently SWAPPED before/after and put the first
post-event slot at 15:10 instead of 15:20 for scenario #4. Half-open overlap
(touching ≠ overlap) keeps flush/zero-gap booking allowed.

calendar_integration/services/bookable_slots_service.py
```diff
@@ -0,0 +1,250 @@
...
+            organization.id,
+            target_calendar_ids,
+            search_window_start,
+            search_window_end,
+            duration,
+            slot_step,
+            with_bulk_modifications=with_bulk_modifications,
+        )
+
+        if policy == EffectivePolicy.unconstrained():
```
hugobessa, line 140: The byte-for-byte no-policy guarantee: when the resolved policy is
`unconstrained()` we return the raw walked candidates and skip ALL policy
work — no managed-calendar blocking-span fetch, no envelope math. Proven by
`test_no_policy_matches_one_calendar_group`, which compares against the real
pre-feature `CalendarGroupService.find_bookable_slots` for the same calendar.

calendar_integration/services/bookable_slots_service.py
```diff
@@ -0,0 +1,250 @@
...
+            raise BookableSlotsValidationError("slot_step must be a positive timedelta.")
+        if duration <= datetime.timedelta(0):
+            raise BookableSlotsValidationError("duration must be a positive timedelta.")
+
+        if now is None:
+            now = timezone.now()
+
+        calendar = Calendar.objects.filter_by_organization(organization.id).get(id=calendar_id)
+
+        if calendar.calendar_type == CalendarType.BUNDLE:
```
hugobessa, line 120: Bundle detection + the all-children-free predicate: a bundle window is
offered only when EVERY `bundle_children` calendar is free (`is_primary` gets
no availability special-casing). The bundle's policy comes from
`resolve_for_bundle` (explicit override → most-restrictive across children),
NOT `resolve_for_calendar`. Empty bundle → [].

calendar_integration/services/bookable_slots_service.py
```diff
@@ -0,0 +1,250 @@
...
+                    blocking_spans,
+                )
+                for cid in target_calendar_ids
+            )
+            if all_free:
+                proposals.append(BookableSlotProposal(start_time=window_start, end_time=window_end))
+            cursor = cursor + slot_step
+        return proposals
+
+    def _buffer_blocking_spans(
```
hugobessa, line 216: Managed-calendar blocking spans are fetched ONLY when a buffer applies
(`buffer_before>0 or buffer_after>0`) — the existing group walker checks
managed calendars against AvailableTime coverage only, so fetching them
unconditionally would change no-buffer output. The fetch window is widened by
`buffer_after` on the START side and `buffer_before` on the END side (the
mirror of the envelope expansion) so a span just outside the search window
can still clip a candidate's dead zone.

public_api/queries.py
```diff
@@ -872,6 +885,49 @@ def calendar_group_bookable_slots(
             for p in proposals
         ]
 
+    @strawberry.field(permission_classes=[IsAuthenticated, OrganizationResourceAccess])
+    def calendar_bookable_slots(
```
hugobessa, line 889: Owner-scope check before the service call: a scoped token may only target its
owner's calendars — a cross-scope `calendar_id` raises
`Calendar.DoesNotExist("Calendar matching query does not exist.")`, the same
not-found message a genuinely missing calendar gets (no existence disclosure).
This mirrors the established `calendars`-query pattern and `scoped_calendar_ids`.

PR #171, opened 2026-06-28
feat(public_api): add code-gated calendar_bookable_slots_with_code query

hugobessa:
## Summary
Phase 6 — the anonymous/code-gated variant of the single+bundle slot query (spec use-case #4). Stacked on Phase 5. No migration.

A new unauthenticated `calendar_bookable_slots_with_code(code, search_window_start, search_window_end, duration_seconds, slot_step_seconds=900)` GraphQL field (`public_api/queries.py`) that mirrors the existing `*_with_code` fields exactly:
- validates the range against `MAX_CODE_GATED_RANGE` **before** any expensive work;
- resolves the `CalendarManagementToken` via `_resolve_code_from_deps` (all token failures → the uniform `"Invalid or expired code."`);
- derives the bound calendar from `token.calendar` (fallback `token.event.calendar`, the established sibling convention) and **rejects group-scoped codes** (single/bundle only) with the uniform error;
- delegates to the Phase-5 `BookableSlotsService.find_bookable_slots_for_calendar` (so a bundle code returns bundle slots, and any policy is applied through the code path);
- returns **only** slots (`start`/`end`) — no policy rule values disclosed to anonymous bookers.

No auth/resource permission classes (code-gated, like its siblings).

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 6**. Stacked on Phase 5 / PR #170. No feature flag (data-presence gate). No migration.

## Test plan
- `docker compose run --rm api uv run pytest public_api/tests/test_code_gated_reads.py -p no:randomly` — invalid / **expired** / **revoked** / **used** code → uniform error; group-scoped code rejected; backwards range + range > `MAX_CODE_GATED_RANGE` rejected; code never consumed by reads; **real-service** happy path (busy span excluded); **booking-policy lead-time filters slots through the code path**; **equivalence** with the authenticated `calendarBookableSlots`; **bundle code** (busy child suppresses the slot); **no policy-field disclosure** (requesting a policy field → GraphQL validation error).
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), schema drift-free, full `pytest -n auto` → **3412 passed**.

### DIFF
--- a/public_api/queries.py
+++ b/public_api/queries.py
@@ -1328,6 +1328,51 @@ class Query:
         ]
 
     @strawberry.field()
+    def calendar_bookable_slots_with_code(
+        self,
+        code: str,
+        search_window_start: datetime.datetime,
+        search_window_end: datetime.datetime,
+        duration_seconds: int,
+        slot_step_seconds: int = 15 * 60,
+    ) -> list[BookableSlotProposalGraphQLType]:
+        """Return policy-compliant bookable slot windows for a calendar via booking code.
+
+        No org token required.  The code gates access to its bound calendar or
+        calendar bundle only. Reads are repeatable: the code is never consumed by
+        this query. A group-scoped code is rejected (single/bundle calendars only).
+
+        The response omits policy rule values — slots only.
+        """
+        _validate_code_gated_range(search_window_start, search_window_end)
+        deps = get_query_dependencies()
+        token = _resolve_code_from_deps(deps, code)
+
+        # Resolve the bound calendar (calendar-scope or event.calendar fallback).
+        # Reject group-scoped codes (single/bundle only).
+        calendar = token.calendar
+        if calendar is None and token.event is not None:
+            calendar = token.event.calendar
+        if calendar is None:
+            raise GraphQLError(_CODE_GATED_ERROR_MESSAGE)
+
+        org = _get_org_from_token(token)
+        service = get_bookable_slots_service()
+        service.initialize(organization=org)
+
+        proposals = service.find_bookable_slots_for_calendar(
+            calendar_id=calendar.id,
+            search_window_start=search_window_start,
+            search_window_end=search_window_end,
+            duration=datetime.timedelta(seconds=duration_seconds),
+            slot_step=datetime.timedelta(seconds=slot_step_seconds),
+        )
+        return [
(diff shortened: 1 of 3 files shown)


public_api/queries.py
```diff
@@ -1327,6 +1327,51 @@ def calendar_group_bookable_slots_with_code(
...
+        _validate_code_gated_range(search_window_start, search_window_end)
+        deps = get_query_dependencies()
+        token = _resolve_code_from_deps(deps, code)
+
+        # Resolve the bound calendar (calendar-scope or event.calendar fallback).
+        # Reject group-scoped codes (single/bundle only).
+        calendar = token.calendar
+        if calendar is None and token.event is not None:
+            calendar = token.event.calendar
+        if calendar is None:
```
hugobessa, line 1356: Group-scoped codes are rejected with the SAME uniform "Invalid or expired
code." used for every other failure — no disclosure of why. The
`token.calendar` → `token.event.calendar` fallback is the established
convention across all `*_with_code` fields (a bundle calendar is a Calendar,
so a bundle code resolves here too); a code whose only scope is a
`calendar_group` leaves `calendar` None and is refused. Single/bundle only,
per the plan.

public_api/queries.py
```diff
@@ -1327,6 +1327,51 @@ def calendar_group_bookable_slots_with_code(
             for p in proposals
         ]
 
+    @strawberry.field()
+    def calendar_bookable_slots_with_code(
```
hugobessa, line 1331: Unauthenticated (`@strawberry.field()` — no IsAuthenticated/
OrganizationResourceAccess), mirroring the sibling code-gated reads. Order is
load-bearing: `_validate_code_gated_range` runs BEFORE `_resolve_code_from_deps`
and the service call (cheap-before-expensive / amplification guard). The
response maps only start/end via `BookableSlotProposalGraphQLType`, so policy
rule values are never exposed to anonymous bookers — asserted by the
no-policy-field-disclosure test. Policy filtering still happens (it's inside
the shared Phase-5 service), proven by the lead-time-through-code test.

PR #172, opened 2026-06-28
feat(calendar): make group bookable-slots query policy-aware

hugobessa:
## Summary
Phase 7 — the one change to an **existing** public surface. `CalendarGroupService.find_bookable_slots` (behind both `calendar_group_bookable_slots` and its `_with_code` variant) now applies the resolved group `BookingPolicy`. Stacked on Phase 6. No migration. No GraphQL signature change.

**Data-presence gate (no feature flag):** after building the candidate proposals (unchanged `required_count` walk), the method resolves the group's `EffectivePolicy` via `BookingPolicyService.resolve_for_group` and applies `slot_engine.apply_policy_filter`. When no `BookingPolicy` resolves (`unconstrained()`) — or no `booking_policy_service` is wired (legacy direct instantiation) — it returns the raw proposals untouched, so the output is **byte-for-byte the pre-feature result** (SPEC acceptance scenario #5). The existing 44 group-service tests stay green unchanged — that *is* the backward-compat proof.

- Buffer reuses the corrected **event-envelope** model (Phase 5); when a buffer applies, blocking spans are fetched for ALL participant calendars (managed included), widening the fetch window by `buffer_after`/`buffer_before` exactly as Phase 5 does.
- Conservative group buffer semantics (documented + tested): a candidate is dropped if **any** participant calendar — across all pools, regardless of `required_count` — has an event in the buffer dead zone ("never offer a slot a participant would reject").
- DI: `booking_policy_service` injected into the `calendar_group_service` provider; a `now` param (default `timezone.now()`) added for lead/horizon cutoffs (no GraphQL-layer change).

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 7** / **Objectives #3** / SPEC **Acceptance scenarios #5**. Stacked on Phase 6 / PR #171. No feature flag (data-presence gate). No migration.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_find_bookable_slots_group_policy.py public_api/tests/test_bookable_slots_query.py -p no:randomly` — **no-policy == pre-feature** (vs the no-service path + concrete expected set); group-override lead/horizon/buffer filtering; per-participant most-restrictive; **explicit group policy beats participant policies**; managed-calendar buffer-width honored (30m); non-required-calendar event in dead zone drops the slot (conservative semantics); GraphQL-resolver-level group policy filtering (+ no-policy returns all).
- Existing group regression: `test_calendar_group_service.py` 44 passed (byte-for-byte).
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), full `pytest -n auto` → **3424 passed**, mypy 297 (no new errors).

### DIFF
--- a/calendar_integration/services/calendar_group_service.py
+++ b/calendar_integration/services/calendar_group_service.py
@@ -55,6 +55,7 @@ from users.models import User
 
 if TYPE_CHECKING:
     from audit.services import AuditService
+    from calendar_integration.services.booking_policy_service import BookingPolicyService
     from calendar_integration.services.calendar_service import CalendarService
 
 
@@ -69,11 +70,15 @@ class CalendarGroupService:
             "CalendarPermissionService | None", Provide["calendar_permission_service"]
         ] = None,
         audit_service: Annotated["AuditService | None", Provide["audit_service"]] = None,
+        booking_policy_service: Annotated[
+            "BookingPolicyService | None", Provide["booking_policy_service"]
+        ] = None,
     ) -> None:
         self.organization = None
         self.calendar_service = calendar_service
         self.calendar_permission_service = calendar_permission_service
         self.audit_service = audit_service
+        self.booking_policy_service = booking_policy_service
 
     def _audit_group_write(
         self,
@@ -112,8 +117,13 @@ class CalendarGroupService:
         or authenticate `self.calendar_service` with the same organization — the
         grouped-event flow needs external-provider adapters if any of the
         selected calendars is backed by one.
+
+        Also propagates the org context to ``booking_policy_service`` when present,
+        so the policy resolver is bound to the same tenant.
         """
         self.organization = organization
+        if self.booking_policy_service is not None:
+            self.booking_policy_service.initialize(organization)
 
     def _assert_initialized(self) -> None:
         if self.organization is None:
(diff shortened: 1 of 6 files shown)


calendar_integration/services/calendar_group_service.py
```diff
@@ -1040,4 +1073,42 @@ def find_bookable_slots(
...
+
+        # ------------------------------------------------------------------
+        # Policy filter — gated by data-presence
+        # ------------------------------------------------------------------
+        # When no booking_policy_service is injected (e.g. legacy test fixtures
+        # that instantiate CalendarGroupService directly without DI), or when the
+        # resolved policy is unconstrained (no BookingPolicy anywhere for this
+        # group), skip ALL policy work so the output is byte-for-byte the
+        # pre-feature engine result.
+        if self.booking_policy_service is None:
```
hugobessa, line 1085: The data-presence gate (this feature has no flag). Two short-circuits return
the raw pre-feature proposals: (1) `booking_policy_service is None` — a legacy
direct instantiation without DI; in production the provider always wires it,
so this only protects old test fixtures; (2) `policy == unconstrained()` — no
BookingPolicy resolves for the group. Both return BEFORE any buffer fetch or
filter, so an org with no policy gets byte-for-byte identical output. The 44
existing group tests (no policy) staying green is the backward-compat proof.

calendar_integration/services/calendar_group_service.py
```diff
@@ -1040,4 +1073,42 @@ def find_bookable_slots(
...
+        no_buffer = policy.buffer_before <= datetime.timedelta(0) and policy.buffer_after <= (
+            datetime.timedelta(0)
+        )
+        if no_buffer:
+            buffer_blocking_spans: slot_engine.SpansByCalendarId = {}
+        else:
+            buffer_blocking_spans = slot_engine.fetch_blocking_spans(
+                org_id,
+                all_calendar_ids,
+                search_window_start - policy.buffer_after,
```
hugobessa, line 1109: Buffer fetch mirrors Phase 5's `BookableSlotsService._buffer_blocking_spans`
exactly: fetched only when a buffer applies, across ALL participant calendars
(managed included — the group walk otherwise checks managed calendars only
against AvailableTime), and the window is widened by `buffer_after` on the
START side / `buffer_before` on the END side (mirror of the event-envelope
expansion) so a span just outside the search window can still clip a
candidate's dead zone. `apply_policy_filter` flattens all participants' spans
and drops a candidate overlapping any dead zone (conservative, documented).

PR #173, opened 2026-06-28
feat(calendar): enforce booking policy on single/bundle event creation

hugobessa:
## Summary
Phase 8a — booking-time enforcement for the single-calendar, bundle, and code-gated-single write paths (spec use-case #5). Stacked on Phase 7. No migration. Discovery and the write path now AGREE: a time `calendar_bookable_slots` would not offer cannot be booked.

- New `BookingPolicyViolationError(CalendarIntegrationError)` (`calendar_integration/exceptions.py`).
- A single enforcement gate `_check_booking_policy` + call in `CalendarService.create_event` (the facade all three paths funnel through). It resolves the `EffectivePolicy` (`resolve_for_bundle` for a bundle, `resolve_for_calendar` otherwise), builds a `BookableSlotProposal(start, end)`, fetches buffer blocking spans the **same way Phase 5 does**, and calls `slot_engine.apply_policy_filter([proposal], …)` — empty result ⇒ violation. So enforcement is byte-identical to discovery (lead inclusive, horizon inclusive, event-envelope half-open buffer).
- **Data-presence gate**: `unconstrained()` ⇒ skip the check ⇒ pre-feature write behavior, byte-for-byte.
- **Enforce-once on bundles**: the bundle fan-out passes a private `_enforce_policy=False` for the per-child creates, so the policy is checked exactly once at the top-level entry using `resolve_for_bundle`. This prevents a child's *individual* policy from rejecting a booking the bundle policy (and discovery) allows — a discovery/enforcement divergence, now tested.
- The check runs inside the existing `create_event` transaction (`ATOMIC_REQUESTS` / the code path's `transaction.atomic()`), so a violation rolls back the whole write — no event or blocked time (incl. bundle children) is persisted.
- Error mapping: `schedule_event` → explanatory GraphQL error; `create_calendar_event_with_code` → `SLOT_UNAVAILABLE` (no policy internals leaked to anonymous bookers).
- DI: `booking_policy_service` injected into `CalendarService`. The facade fetch now populates `_calendar_cache` so the event service reuses it (no duplicate query).

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 8a** / **Use-cases #5** / SPEC **Acceptance #7**. Stacked on Phase 7 / PR #172. No feature flag (data-presence gate). No migration.

Follow-up note (out of scope here): the internal `create_recurring_event` shortcut delegates to the event service directly and bypasses the facade gate — the public `schedule_event` rrule path IS enforced (on the first occurrence via the facade). Recurring-series enforcement beyond the first occurrence is not part of this plan.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_booking_policy_enforcement.py public_api/tests/test_booking_policy_enforcement_mutations.py -p no:randomly` — lead / horizon / buffer rejection on single, bundle, and code-gated paths; compliant bookings succeed (incl. a real bundle success creating primary + child rows); **no-policy ⇒ unchanged**; **zero rows persisted on a rejected bundle booking** (no orphan children); event-envelope buffer; discovery/enforcement agreement (negative + positive, lead/horizon/buffer); **bundle explicit-policy-allows-when-child-stricter** (proves enforce-once matches discovery).
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), full `pytest -n auto` → **3455 passed**, mypy 297 (no new errors).

### DIFF
--- a/calendar_integration/services/calendar_bundle_service.py
+++ b/calendar_integration/services/calendar_bundle_service.py
@@ -110,7 +110,11 @@ class BundleServiceHost(Protocol):
     ) -> Iterable[AvailableTimeWindow]: ...
 
     def create_event(
-        self, calendar_id: int, event_data: CalendarEventInputData
+        self,
+        calendar_id: int,
+        event_data: CalendarEventInputData,
+        *,
+        _enforce_policy: bool = True,
     ) -> CalendarEvent: ...
 
     def update_event(
@@ -434,7 +438,14 @@ class CalendarBundleService:
             recurrence_rule=event_data.recurrence_rule,
         )
 
-        primary_event = self._host.create_event(primary_calendar.id, primary_event_data)
+        # Policy was already enforced once at the top-level CalendarService.create_event
+        # entry (using resolve_for_bundle). Skip re-enforcement for all child creates to
+        # avoid: (a) N redundant policy resolutions and buffer fetches, and (b) false
+        # rejections when a child has its own stricter individual policy that would block
+        # a booking the bundle policy (and Phase-5 discovery) correctly permits.
+        primary_event = self._host.create_event(
+            primary_calendar.id, primary_event_data, _enforce_policy=False
+        )
 
         # Mark primary event as part of bundle
         primary_event.bundle_calendar = bundle_calendar
@@ -459,7 +470,9 @@ class CalendarBundleService:
                     resource_allocations=[],
                 )
 
-                child_event = self._host.create_event(child_calendar.id, child_event_data)
+                child_event = self._host.create_event(
+                    child_calendar.id, child_event_data, _enforce_policy=False
+                )
 
                 # Link to primary event and bundle
                 child_event.bundle_calendar = bundle_calendar
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -1298,14 +1298,25 @@ class CalendarService(BaseCalendarService):
         if not allowed:
             raise BookingPolicyViolationError()
 
-    def create_event(self, calendar_id: int, event_data: CalendarEventInputData) -> CalendarEvent:
+    def create_event(
+        self,
+        calendar_id: int,
+        event_data: CalendarEventInputData,
+        *,
+        _enforce_policy: bool = True,
+    ) -> CalendarEvent:
         """
         Create a new event in the calendar.
         :param calendar_id: Internal ID of the calendar
         :param event_data: Dictionary containing event details.
+        :param _enforce_policy: Internal flag; callers must NOT pass this. Set to
+            ``False`` by the bundle fan-out so policy is enforced exactly once at the
+            top-level entry point (using ``resolve_for_bundle``), not again for each
+            child create (which would use ``resolve_for_calendar`` on the child and
+            could falsely reject bookings the bundle policy permits).
         :return: Response from the calendar client.
         """
-        if self.organization is not None:
+        if _enforce_policy and self.organization is not None:
             # Enforcement runs inside the existing transaction (ATOMIC_REQUESTS) so any
             # violation rolls back the entire write — no event or blocked time is created.
             try:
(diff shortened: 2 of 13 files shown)


calendar_integration/services/calendar_service.py
```diff
@@ -1214,13 +1225,117 @@ def _serialize_event_data_input(
...
         :param calendar_id: Internal ID of the calendar
         :param event_data: Dictionary containing event details.
+        :param _enforce_policy: Internal flag; callers must NOT pass this. Set to
+            ``False`` by the bundle fan-out so policy is enforced exactly once at the
+            top-level entry point (using ``resolve_for_bundle``), not again for each
+            child create (which would use ``resolve_for_calendar`` on the child and
+            could falsely reject bookings the bundle policy permits).
         :return: Response from the calendar client.
         """
+        if _enforce_policy and self.organization is not None:
```
hugobessa, line 1319: The single enforcement gate for all three write paths (single, bundle,
code-gated — they all funnel through this facade `create_event`). Gated by
`_enforce_policy` (private; the bundle fan-out passes False for child creates)
so the policy is checked exactly once. Runs inside the request/atomic
transaction, so a `BookingPolicyViolationError` rolls back the entire write —
no event/blocked-time persisted. The fetched calendar is cached so the event
service doesn't re-query it.

calendar_integration/services/calendar_service.py
```diff
@@ -1214,13 +1225,117 @@ def _serialize_event_data_input(
...
+            buffer_blocking_spans = slot_engine.fetch_blocking_spans(
+                self.organization.id,
+                target_calendar_ids,
+                start_time - policy.buffer_after,
+                end_time + policy.buffer_before,
+                with_bulk_modifications=False,
+            )
+
+        proposal = BookableSlotProposal(start_time=start_time, end_time=end_time)
+        allowed = slot_engine.apply_policy_filter([proposal], policy, now, buffer_blocking_spans)
```
hugobessa, line 1297: Enforcement REUSES the discovery engine: a single `BookableSlotProposal(start,
end)` run through `slot_engine.apply_policy_filter` with the same buffer-fetch
widening as Phase 5. Empty result ⇒ the booking is a slot discovery would not
offer ⇒ reject. This guarantees discovery and the write path agree by
construction (same lead/horizon/buffer semantics, same `now` source). When the
policy is `unconstrained()` the method returns earlier — the data-presence
gate, so no-policy bookings are byte-for-byte unchanged.

calendar_integration/services/calendar_bundle_service.py
```diff
@@ -434,7 +438,14 @@ def create_bundle_event(
...
         )
 
-        primary_event = self._host.create_event(primary_calendar.id, primary_event_data)
+        # Policy was already enforced once at the top-level CalendarService.create_event
+        # entry (using resolve_for_bundle). Skip re-enforcement for all child creates to
+        # avoid: (a) N redundant policy resolutions and buffer fetches, and (b) false
+        # rejections when a child has its own stricter individual policy that would block
+        # a booking the bundle policy (and Phase-5 discovery) correctly permits.
+        primary_event = self._host.create_event(
+            primary_calendar.id, primary_event_data, _enforce_policy=False
```
hugobessa, line 447: Enforce-once: the bundle fan-out creates the primary + child events via the
facade `create_event` with `_enforce_policy=False`, because the top-level
bundle entry already enforced the bundle's `resolve_for_bundle` policy. Without
this, each child create would re-check its OWN `resolve_for_calendar` policy —
and a child with a stricter individual policy would reject a booking the bundle
policy (and Phase-5 discovery, which only uses `resolve_for_bundle`) allows.
The `test_bundle_explicit_policy_allows_when_child_individual_policy_is_stricter`
test locks this in.

PR #174, opened 2026-06-28
feat(calendar): enforce booking policy on group event creation

hugobessa:
## Summary
Phase 8b — the final phase: booking-time policy enforcement for the **group** write path (spec use-case #5, group). Stacked on Phase 8a. No migration. With this, every slot surface (single / bundle / group, authed + code-gated) and every write path enforce the same policy, and discovery + booking agree everywhere.

- `CalendarGroupService._check_group_booking_policy` + a call in `create_grouped_event`, placed AFTER `_assert_calendars_available` and BEFORE any event / blocked-time / `CalendarEventGroupSelection` write. The method is `@transaction.atomic()`, so a `BookingPolicyViolationError` rolls back the entire group write — zero rows persisted.
- Mirrors Phase 8a/7 exactly: `resolve_for_group` (data-presence skip on `unconstrained()`), the same `all_calendar_ids` participant set as Phase-7 discovery, event-envelope buffer (widen start by `buffer_after`, end by `buffer_before`), and `slot_engine.apply_policy_filter([proposal], …)` — empty ⇒ raise. So group discovery (`calendar_group_bookable_slots`) and group booking agree by construction.
- `BookingPolicyViolationError` (Phase 8a) reused — subclasses `CalendarIntegrationError` only, so it is NOT swallowed by the broader `except CalendarGroupError` / `except (EventManagementError, CalendarGroupError)` at either mutation site.
- Error mapping at both `CalendarGroupMutations` call sites in `calendar_integration/mutations.py`: authed → `CalendarGroupEventResult(success=False, error_message=…)`; code-gated → `CodeEventResult(success=False, error_code=SLOT_UNAVAILABLE)` with the code **not consumed** (rolled back, the booker may retry).
- No new DI: `CalendarGroupService` already had `booking_policy_service` injected (Phase 7).

## Plan reference
`ai-plans/2026-06-26-BOOKABLE_SLOTS_SINGLE_CALENDAR_AND_BOOKING_POLICY_IMPLEMENTATION_PLAN.md` — **Phase 8b** / **Use-cases #5** / SPEC **Acceptance #7**. Stacked on Phase 8a / PR #173. No feature flag (data-presence gate). No migration. Final executable phase of the plan.

## Test plan
- `docker compose run --rm api uv run pytest calendar_integration/tests/services/test_booking_policy_enforcement_group.py public_api/tests/test_booking_policy_enforcement_group_mutations.py -p no:randomly` — group booking inside lead / beyond horizon / inside buffer rejected; compliant group booking succeeds (event + selections persisted); **no-policy ⇒ unchanged**; **zero rows persisted on a rejected booking** (events + blocked-times + selections); event-envelope buffer; discovery/enforcement agreement (lead / horizon / positive); both mutation paths incl. code-not-consumed-on-violation.
- Existing group-booking regression: `test_calendar_group_service.py` 44 passed.
- Outer gate: `makemigrations --check` (no changes), `check --deploy` (0 errors), full `pytest -n auto` → **3478 passed**, mypy 297 (no new errors).


calendar_integration/services/calendar_group_service.py
```diff
@@ -532,6 +611,16 @@ def create_grouped_event(self, data: CalendarGroupEventInputData) -> CalendarEve
         all_selected_ids = {cid for sel in data.slot_selections for cid in sel.calendar_ids}
         self._assert_calendars_available(all_selected_ids, data.start_time, data.end_time)
 
+        # --- Booking policy enforcement ---
+        # Runs inside the @transaction.atomic() so any violation rolls back the
+        # entire write — no events or blocked times are persisted on rejection.
+        self._check_group_booking_policy(
```
hugobessa, line 617: The enforcement gate sits AFTER availability validation but BEFORE the
primary event create, the `CalendarEventGroupSelection` rows, and the child
blocked-times — all inside the method's `@transaction.atomic()`. So a
`BookingPolicyViolationError` rolls back EVERY row: no partial group booking
is ever persisted (asserted by the no-rows-on-violation tests, counting
events + blocked-times + selections). For the code-gated path the consume
happens after this in the same transaction, so a violation also leaves the
code unconsumed.

calendar_integration/services/calendar_group_service.py
```diff
@@ -453,6 +454,84 @@ def check_group_availability(
             )
         return results
 
+    # ------------------------------------------------------------------
+    # Booking policy enforcement (group path)
+    # ------------------------------------------------------------------
+
+    def _check_group_booking_policy(
```
hugobessa, line 461: Enforcement == discovery by construction: same `resolve_for_group` (NOT
per-participant), the same all-pool `all_calendar_ids` participant set Phase-7
`find_bookable_slots` uses, the same event-envelope buffer widening, the same
`slot_engine.apply_policy_filter`. So a time `calendar_group_bookable_slots`
would not offer is rejected here, and one it would offer is accepted. Using
all POOL calendars (not just the booking's selected ones) is deliberate — it
matches discovery's conservative set; a selected-only set could accept a
booking discovery wouldn't offer. `unconstrained()` ⇒ skip (data-presence gate).

calendar_integration/mutations.py
```diff
@@ -1429,6 +1437,15 @@ def create_calendar_group_event_with_code(
                 error_code=BookingCodeErrorCode.NOT_PERMITTED,
                 error_message="This code does not permit booking on this calendar.",
             )
+        except BookingPolicyViolationError:
```
hugobessa, line 1440: Code-gated group booking: a policy violation maps to `SLOT_UNAVAILABLE` with
no policy internals leaked, and — because the check raises before
`consume_code` inside the same `transaction.atomic()` — the booking code is
NOT consumed, so an anonymous booker can retry a different slot. Ordered
before the broader `except (EventManagementError, CalendarGroupError)`, which
is safe because `BookingPolicyViolationError` subclasses neither.

PR #177, opened 2026-07-04
feat(legal): scaffold legal app with immutably-versioned PolicyDocument + admin

hugobessa:
## Summary
Phase 1 of the **SMS MFA Consent** plan. Scaffolds a new `legal` Django app and the `PolicyDocument` model — the storage for the privacy policy, terms of use, and SMS-messaging consent text that later phases capture consent against and expose over REST. Additive only: no existing code path changes, no signup/MFA behavior touched.

`PolicyDocument` is a **global** model (plain `common.models.BaseModel`, not `OrganizationModel`) because v1 policies are Vinta-owned, not per-organization (see the plan's **Guiding Decisions → Policy scope**). It is **immutably versioned**: each publish is a new row keyed by `(document_type, version)`; the admin locks every field on existing rows and blocks deletion, so a published version a user may later accept can never be silently rewritten or removed.

Adds:
- `PolicyDocument` model — `document_type` enum (`PRIVACY_POLICY` / `TERMS_OF_USE` / `SMS_CONSENT`), monotonic `version`, `title`, `body_markdown` (raw markdown, rendered client-side — no markdown lib in the repo), `published_at`. Unique `(document_type, version)`, index `(document_type, -version)`.
- `PolicyDocumentManager` / `PolicyDocumentQuerySet` — `latest_for(document_type)` (highest version, or `None`) and `latest_per_type()` (one latest row per type via Postgres `DISTINCT ON`).
- `PolicyDocumentAdmin` — published rows read-only on the change form, deletion blocked, list-filtered by type, add-form suggests the next version per type.
- `PolicyDocumentFactory`, migration `0001_initial`, model + admin tests.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 1** (first phase; branches off `main`). No feature flag. One additive migration (`legal.0001_initial`).

## Test plan
- `docker compose run --rm api uv run pytest legal/tests/ -n auto` — 12 passed: `latest_for` / `latest_per_type` selection, unique-constraint `IntegrityError`, admin field-lock on POST (not just GET), add-form version suggestion, and delete-permission denial (GET forbidden + POST rejected, row survives).
- Outer gate: `python manage.py check --deploy` clean; `makemigrations --check` clean; migration reverse/reapply clean; full `pytest -n auto` → **3537 passed**.


legal/models.py
```diff
@@ -0,0 +1,49 @@
...
+from legal.managers import PolicyDocumentManager
+
+
+class PolicyDocumentType(models.TextChoices):
+    PRIVACY_POLICY = "privacy_policy", "Privacy Policy"
+    TERMS_OF_USE = "terms_of_use", "Terms of Use"
+    SMS_CONSENT = "sms_consent", "SMS Messaging Consent"
+
+
+class PolicyDocument(BaseModel):
```
hugobessa, line 16: Deliberately a plain `BaseModel`, not an `OrganizationModel` — no `organization`
FK. v1 policy documents are global (Vinta-owned), per the plan's
**Guiding Decisions → Policy scope**. Per-reseller policies are an explicit
non-goal; if they ever land, that's a new tenant-scoped model, not a column here.

legal/models.py
```diff
@@ -0,0 +1,49 @@
...
+    publish (enforced in `legal/admin.py`, not just convention). "Latest" is
+    the highest `version` for a given `document_type` among published rows.
+
+    Global — not tenant-scoped. `users.User` is a global model and policy
+    documents are Vinta-owned (not per-organization) in v1, so this is a plain
+    `BaseModel`, not an `OrganizationModel`; it carries no `organization` FK.
+    """
+
+    document_type = models.CharField(max_length=32, choices=PolicyDocumentType)
+    version = models.PositiveIntegerField(help_text="Monotonically increasing per document_type.")
```
hugobessa, line 31: Immutable-versioning invariant: a "publish" is a new row, never an edit.
`version` is monotonic per `document_type`, enforced by the `(document_type,
version)` unique constraint. Consent records in Phase 3 FK to a specific version
so we can prove the exact text a user accepted — which only holds if rows are
never mutated (see the admin lock below).

legal/querysets.py
```diff
@@ -0,0 +1,18 @@
...
+        return self.filter(document_type=document_type)
+
+    def latest_per_type(self) -> "PolicyDocumentQuerySet":
+        """Return one row per document_type: the highest-version row.
+
+        Implemented with Postgres ``DISTINCT ON`` via ``QuerySet.distinct(*fields)``
+        (Postgres-only API). The project's sole supported database is Postgres, so
+        this avoids a less efficient group-by-then-fetch round trip.
+        """
+        return self.order_by("document_type", "-version").distinct("document_type")
```
hugobessa, line 18: `latest_per_type()` uses Postgres `DISTINCT ON` (`.distinct("document_type")`
after ordering by `document_type, -version`) — one query, one row per type. This
is a Postgres-only API; fine because Postgres is the project's only supported DB.
The `order_by` fields must lead with the `distinct` field, hence this exact ordering.

legal/admin.py
```diff
@@ -0,0 +1,75 @@
...
+    already seen/accepted. New versions are authored by adding a new row.
+    """
+
+    form = PolicyDocumentAdminForm
+    list_display = ("title", "document_type", "version", "published_at")
+    list_filter = ("document_type",)
+    search_fields = ("title",)
+    ordering = ("-version",)
+
+    def get_readonly_fields(
```
hugobessa, line 63: `get_readonly_fields` locks every field once a row has a pk — the add form stays
editable, published rows are frozen. Paired with `has_delete_permission` below
(returns `False`, mirroring `audit/admin.py`), this is what makes the "immutable,
audit-grade proof" goal real at the admin layer, ahead of the `UserConsent`
`on_delete=PROTECT` FK that reinforces it in Phase 3.

PR #178, opened 2026-07-04
feat(legal): add read-only policy document REST API

hugobessa:
## Summary
Phase 2 of the **SMS MFA Consent** plan (stacked on Phase 1 / PR #177). Exposes the `PolicyDocument` versions from Phase 1 over four read-only GET endpoints so the frontend can render policy text at the consent step:

- `GET /policy-documents/latest/` — **public**, one latest published version per `document_type`.
- `GET /policy-documents/latest/{document_type}/` — **public**, latest version of one type; 404 on unknown enum value or a type with no published rows.
- `GET /policy-documents/{id}/` — **authenticated**, exact version by pk.
- `GET /policy-documents/` — **authenticated**, full version history (newest-first per type) with optional `?document_type=` filter.

The two `latest` endpoints are public because the frontend must render policy text before a session exists (mid-signup / pre-OAuth-completion, resolving the plan's **Open Questions** on doc-read auth); the full-history + by-id endpoints stay behind auth. No write surface is exposed anywhere (`http_method_names = ("get","head","options")`, fully read-only serializer).

Built on a plain DRF `ReadOnlyModelViewSet` rather than the `*VintaScheduleModelViewSet` base — that family mixes in tenant scoping (irrelevant; `PolicyDocument` is global) and a virtual-model requirement this flat 5-field serializer doesn't need. Mirrors the existing `organizations.views.ServiceAccountViewSet` precedent.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 2** (stacked on Phase 1). No feature flag. No migrations. `schema.yml` regenerated (drf-spectacular).

## Test plan
- `docker compose run --rm api uv run pytest legal/tests/test_policy_document_api.py -n auto` — latest-list one-per-type, latest-by-type highest version, by-id exact, history newest-first + `?document_type=` filter, unknown type → 404, invalid filter value → 400, and the auth split (401 on history/by-id unauth, 200 on public `latest`).
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; schema byte-identical to a fresh `spectacular` run; full `pytest -n auto` → **3549 passed**.

Note (follow-up, not in this PR): the two public endpoints have no rate-limit — deliberately, since the repo has no `throttle_classes` precedent to follow; content is non-secret and static-ish. Worth revisiting if we add a project-wide anon-throttle policy.

### DIFF
--- a/legal/views.py
+++ b/legal/views.py
@@ -1,3 +1,5 @@
+from typing import cast
+
 from drf_spectacular.utils import OpenApiParameter, OpenApiResponse, extend_schema
 from rest_framework import status
 from rest_framework.decorators import action
@@ -8,6 +10,7 @@ from rest_framework.viewsets import ReadOnlyModelViewSet
 
 from legal.filtersets import PolicyDocumentFilterSet
 from legal.models import PolicyDocument, PolicyDocumentType
+from legal.querysets import PolicyDocumentQuerySet
 from legal.serializers import PolicyDocumentSerializer
 
 
@@ -56,7 +59,7 @@ class PolicyDocumentViewSet(ReadOnlyModelViewSet):
 
         Public — no authentication required.
         """
-        queryset = PolicyDocument.objects.latest_per_type()
+        queryset = self.get_queryset().latest_per_type()
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data, status=status.HTTP_200_OK)
 
@@ -86,7 +89,8 @@ class PolicyDocumentViewSet(ReadOnlyModelViewSet):
         if document_type not in valid_types:
             raise NotFound(detail=f"Unknown document_type '{document_type}'.")
 
-        document = PolicyDocument.objects.latest_for(document_type)
+        queryset = cast(PolicyDocumentQuerySet, self.get_queryset())
+        document = queryset.of_type(document_type).order_by("-version").first()
         if document is None:
             raise NotFound(detail=f"No published document of type '{document_type}'.")
 
--- a/legal/views.py
+++ b/legal/views.py
@@ -1,0 +1,94 @@
+from drf_spectacular.utils import OpenApiParameter, OpenApiResponse, extend_schema
+from rest_framework import status
+from rest_framework.decorators import action
+from rest_framework.exceptions import NotFound
+from rest_framework.permissions import AllowAny, IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.viewsets import ReadOnlyModelViewSet
+
+from legal.filtersets import PolicyDocumentFilterSet
+from legal.models import PolicyDocument, PolicyDocumentType
+from legal.serializers import PolicyDocumentSerializer
+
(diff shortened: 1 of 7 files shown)


legal/views.py
```diff
@@ -0,0 +1,98 @@
...
+    No write surface is exposed anywhere on this viewset.
+    """
+
+    queryset = PolicyDocument.objects.all()
+    serializer_class = PolicyDocumentSerializer
+    permission_classes = (IsAuthenticated,)
+    filterset_class = PolicyDocumentFilterSet
+    http_method_names = ("get", "head", "options")
+
+    def get_permissions(self):
```
hugobessa, line 47: Per-action auth split: `latest` / `latest_by_type` are public (`AllowAny`);
`list` (full history) and `retrieve` (by id) keep the class-level
`IsAuthenticated`. This resolves the plan's **Open Questions** on doc-read auth —
the frontend needs the current policy text before a session exists (mid-signup /
pre-OAuth), but the full version history stays gated. Fails closed: if an action
name here drifts, it falls back to authenticated.

legal/views.py
```diff
@@ -0,0 +1,98 @@
...
+                type=str,
+                description="One of PolicyDocumentType's values (e.g. sms_consent).",
+            ),
+        ],
+        responses={
+            200: PolicyDocumentSerializer,
+            404: OpenApiResponse(description="Unknown document_type, or no published version yet"),
+        },
+    )
+    @action(detail=False, methods=["get"], url_path="latest/<str:document_type>")
```
hugobessa, line 81: `url_path="latest/<str:document_type>"` is collision-safe with the `<str:pk>`
detail route: the router registers dynamic list-actions (`latest`,
`latest/{type}`) *before* the `{pk}` retrieve route, so `/latest/` and
`/latest/sms_consent/` never fall through to by-id retrieve. Verified by the
routing test + `router.urls` order.

legal/views.py
```diff
@@ -0,0 +1,98 @@
...
+from rest_framework.viewsets import ReadOnlyModelViewSet
+
+from legal.filtersets import PolicyDocumentFilterSet
+from legal.models import PolicyDocument, PolicyDocumentType
+from legal.querysets import PolicyDocumentQuerySet
+from legal.serializers import PolicyDocumentSerializer
+
+
+@extend_schema(tags=["Legal"])
+class PolicyDocumentViewSet(ReadOnlyModelViewSet):
```
hugobessa, line 18: Deliberately a plain `ReadOnlyModelViewSet`, not `ReadOnlyVintaScheduleModelViewSet`:
the Vinta base pulls in `TenantScopedViewMixin` (no org on this global model) and
a `VirtualModelSerializer` requirement this flat, no-N+1 serializer doesn't warrant.
Precedent: `organizations.views.ServiceAccountViewSet`. `http_method_names`
hard-limits to safe verbs so no write surface can leak in.

PR #179, opened 2026-07-04
feat(legal): add UserConsent model + ConsentService with audit

hugobessa:
## Summary
Phase 3 of the **SMS MFA Consent** plan (stacked on Phase 2 / PR #178). Adds the durable consent record and the service that writes/queries it — the foundation the signup capture (Phase 4) and the SMS gate (Phase 5) build on. Not yet wired to any request path.

- `UserConsent` — append-only, one row per `(user, policy_document version)` accepted. Global (plain `BaseModel`, no `organization` FK — consent is per-user). Audit-grade proof fields: `accepted_at`, `ip_address`, `user_agent`, `source` (`ConsentSource`: SIGNUP_FORM / OAUTH_STEP / API). FK to `users.User` is `CASCADE`; FK to `PolicyDocument` is **`PROTECT`** so an accepted version can never be deleted (proof integrity, reinforcing the Phase 1 admin lock).
- `UserConsentManager.has_sms_consent(user)` — the consent-once-ever gate predicate: `exists()` of any `SMS_CONSENT` consent row, regardless of version.
- `ConsentService` (DI-registered, receives `audit_service`) — `record_consent(user, document_type, *, source, ip, user_agent)` version-pins the latest `PolicyDocument` (raises `NoPolicyDocumentError` if none published) and creates the row; `has_sms_consent(user)`.
- Migration `0002_userconsent` (additive, reverse/reapply clean).

## Design call — pre-org consent & the audit trail
`Audit` is tenant-scoped (`OrganizationModel`, needs `organization_id`), but consent is captured during signup **before** the user has an organization (email path provisions the org on email confirmation; OAuth in the adapter). So `ConsentService` emits an `AuditService` CREATE record only when the user already has an active membership, and skips it (INFO log) otherwise — which is the common signup case. The `UserConsent` row itself is the audit-grade proof (timestamp, exact version, IP, user-agent, source), so consent proof does not depend on the org-scoped audit trail. Recorded as an accepted **Open Question** in the plan; if a global (non-tenant) audit stream is later wanted for pre-org events, add it then.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 3** (stacked on Phase 2). No feature flag. One additive migration (`legal.0002_userconsent`).

## Test plan
- `docker compose run --rm api uv run pytest legal/tests/test_consent_service.py legal/tests/test_userconsent_model.py -n auto` — `record_consent` version-pins + stores ip/ua/source; raises when no published doc; audit emitted when the user has a membership and skipped when not (both branches); `has_sms_consent` true only for an SMS_CONSENT row (any version); `PROTECT` blocks deleting a consented `PolicyDocument`.
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; migration reverse/reapply clean; full `pytest -n auto` → **3564 passed**.


legal/models.py
```diff
@@ -47,3 +47,47 @@ class Meta:
...
+    Re-consent policy (consent-once-ever): the SMS-consent gate is satisfied by
+    ANY `UserConsent` row whose document is of type `SMS_CONSENT`, regardless
+    of version — see `UserConsentManager.has_sms_consent`.
+    """
+
+    user = models.ForeignKey("users.User", on_delete=models.CASCADE, related_name="consents")
+    # PROTECT: proof integrity. An accepted policy version must never be
+    # deletable while a UserConsent row still references it as what the user saw.
+    policy_document = models.ForeignKey(
+        PolicyDocument, on_delete=models.PROTECT, related_name="consents"
```
hugobessa, line 80: `on_delete=PROTECT` on the policy_document FK is load-bearing: an accepted
version must never disappear, or the consent proof it anchors becomes
unverifiable. This is the DB-level backstop behind the Phase 1 admin lock
(published rows read-only + delete blocked). User FK stays CASCADE — deleting
a user removes their consent rows, which is correct.

legal/services.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def has_sms_consent(self, user: User) -> bool:
+        """Return True if `user` has ever accepted any version of SMS_CONSENT.
+
+        Delegates to ``UserConsentManager.has_sms_consent`` (consent-once-ever:
+        any prior SMS-consent row satisfies the gate, regardless of version).
+        """
+        return UserConsent.objects.has_sms_consent(user)
+
+    def _audit_consent_created(self, consent: UserConsent) -> None:
```
hugobessa, line 105: Pre-org audit skip (deliberate, see the PR description + plan Open Questions):
`Audit` needs an `organization_id`, but consent is captured mid-signup before the
user has an org, so we audit only when an active membership exists and skip
otherwise. The `UserConsent` row is the audit-grade proof of record; the
org-scoped `Audit` entry is a best-effort secondary log for post-org grants.

legal/services.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+
+    @inject
+    def __init__(
+        self,
+        audit_service: Annotated[AuditService, Provide["audit_service"]],
+    ) -> None:
+        self.audit_service = audit_service
+
+    def record_consent(
```
hugobessa, line 53: `record_consent` resolves and pins the *latest* published version at call time
and raises `NoPolicyDocumentError` when nothing is published — so a consent row
can never reference a nonexistent document, and always names the exact text the
user saw. Consent-once-ever means later version bumps don't invalidate this row
(`has_sms_consent` matches any SMS_CONSENT version).

PR #180, opened 2026-07-04
feat(accounts,legal): capture policy consent in email + OAuth signup

hugobessa:
## Summary
Phase 4 of the **SMS MFA Consent** plan (stacked on Phase 3 / PR #179). Collects policy consent during onboarding on both signup paths, persisting it through the Phase 3 `ConsentService`. Phone verification is still OFF (Phase 6), so this capture triggers no SMS yet — it just records the consent the Phase 5 gate will require.

**Email/password path** — `accounts/base_forms.py` `BaseVintaScheduleSignupForm`:
- New **required** `accepted_policies` boolean acknowledging the privacy policy, terms of use, and SMS-messaging consent. A missing/false value makes the headless signup invalid.
- `signup()` records `SIGNUP_FORM`-sourced consent for **every published** policy document type (all three per the plan's Open Questions; only `SMS_CONSENT` gates SMS). A type with no published version yet is logged and skipped so signup never breaks — the Phase 5 gate independently fails closed.

**OAuth2 path** — new authenticated `POST /consents/` (`ConsentViewSet`): OAuth signups use `SOCIALACCOUNT_AUTO_SIGNUP` and collect no consent at signup, so the frontend calls this post-signup (before requesting phone verification). Records `OAUTH_STEP` consent for `request.user` only (payload is just `document_type` — no user id, no IDOR), capturing client IP + user-agent. Unknown/unpublished type → 400, unauthenticated → 401.

Both service call sites are DI-injected (`@inject` + `Provide["consent_service"]`, per AGENTS.md), and the client-IP/user-agent extraction is shared via `common/utils/request_utils.py`. The OAuth/social signup adapter (`SocialAccountAdapter.save_user`) does **not** go through this form, so the new required field does not regress social login.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 4** (stacked on Phase 3). No feature flag. No migrations. `schema.yml` / `schema-auth.yml` regenerated (headless `BaseSignup` now requires `accepted_policies`).

## Test plan
- `docker compose run --rm api uv run pytest accounts/tests/test_signup_consent.py accounts/tests/test_signup_consent_headless.py legal/tests/test_consent_endpoint.py -n auto` — form-level: acknowledgement records a version-pinned `SMS_CONSENT` row (source SIGNUP_FORM), missing acknowledgement → invalid, signup survives an unpublished doc type; **HTTP-level**: real headless signup endpoint with `accepted_policies: true` creates the consent row, without it → 400 no user/consent; endpoint: OAUTH_STEP consent with IP/UA, 401 unauth, 400 invalid type.
- Two existing signup fixtures updated to send `accepted_policies: true`.
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; schema matches a fresh `spectacular` run; full `pytest -n auto` → **3576 passed**.

### DIFF
--- a/accounts/base_forms.py
+++ b/accounts/base_forms.py
@@ -1,17 +1,46 @@
+import logging
+
 from django import forms
 from django.utils import timezone
 
+from legal.exceptions import NoPolicyDocumentError
+from legal.models import ConsentSource, PolicyDocumentType
+from legal.services import ConsentService
 from organizations.models import OrganizationInvitation
 from users.models import Profile
+
+
+logger = logging.getLogger(__name__)
+
+
+def _client_ip_from_request(request: object) -> str | None:
+    """Extract the client IP address from a Django request for audit logging.
+
+    Prefers the first entry of ``X-Forwarded-For`` (set by load balancers /
+    proxies); falls back to ``REMOTE_ADDR``. Robust to a missing/``None``
+    request (allauth's ``signup()`` hook is invoked with ``request=None`` in
+    some tests) — returns ``None`` rather than raising.
+    """
+    meta = getattr(request, "META", {})
+    forwarded_for = meta.get("HTTP_X_FORWARDED_FOR", "")
+    if forwarded_for:
+        return forwarded_for.split(",")[0].strip()
+    return meta.get("REMOTE_ADDR") or None
+
+
+def _user_agent_from_request(request: object) -> str:
+    """Extract the client User-Agent header. Robust to a missing/``None`` request."""
+    return getattr(request, "META", {}).get("HTTP_USER_AGENT", "")
 
 
 class BaseVintaScheduleSignupForm(forms.Form):
     """
     Base form for user signup.
 
-    Captures first_name, last_name, and an optional organization_name. At
-    signup time, the intended org name is persisted on Profile.pending_organization_name
-    so it can be consumed during email-confirmation provisioning (Phase 3).
+    Captures first_name, last_name, an optional organization_name, and a
+    required policy-acceptance acknowledgement. At signup time, the intended
+    org name is persisted on Profile.pending_organization_name so it can be
+    consumed during email-confirmation provisioning (Phase 3).
 
     When a non-expired, unaccepted OrganizationInvitation exists for the
     signup email, the org name is left blank — the user will auto-join the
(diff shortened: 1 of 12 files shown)


accounts/base_forms.py
```diff
@@ -25,6 +38,16 @@ class BaseVintaScheduleSignupForm(forms.Form):
         required=False,
         label="Organization Name",
     )
+    accepted_policies = forms.BooleanField(
```
hugobessa, line 41: Contract change: `accepted_policies` is now a REQUIRED field on the one
production email/password signup surface (`HEADLESS_ONLY=True`). The frontend
must send it (documented in the Phase 7 handoff). It gates only the email path —
social/OAuth signup goes through `SocialAccountAdapter.save_user`, not this form,
so it's unaffected (verified). A dedicated headless-HTTP test proves the round-trip.

accounts/base_forms.py
```diff
@@ -35,17 +58,65 @@ def _has_pending_invitation(self, email: str) -> bool:
...
+            except NoPolicyDocumentError:
+                logger.warning(
+                    "Skipping signup consent capture for document_type=%s, user=%s: "
+                    "no published PolicyDocument exists yet.",
+                    document_type,
+                    user.pk,
+                )
+
+    @inject
+    def signup(
```
hugobessa, line 95: `@inject` on `signup` with a `Provide["consent_service"]` default keeps allauth's
positional `form.signup(request, user)` call working while satisfying AGENTS.md's
"inject services, don't instantiate" rule. Consent is recorded for every *published*
doc type; an unpublished type raises `NoPolicyDocumentError`, which is logged and
swallowed per-type so signup always succeeds — the SMS gate (Phase 5) is the real
enforcement and fails closed if no `SMS_CONSENT` row exists.

legal/views.py
```diff
@@ -96,3 +103,61 @@ def latest_by_type(self, request, document_type: str | None = None):
...
+        request=ConsentCreateSerializer,
+        responses={
+            201: UserConsentSerializer,
+            400: OpenApiResponse(
+                description="Invalid document_type, or no published document of that type yet"
+            ),
+        },
+    )
+    @inject
+    def create(
```
hugobessa, line 139: Write-only OAuth-step endpoint. Consent always writes to `request.user`
(`ConsentCreateSerializer` accepts only `document_type` — no user id, no
mass-assignment/IDOR). An unpublished type surfaces as a 400 here (not swallowed
like the signup path) because the caller explicitly named that one type.
Repeat POSTs create additional rows — intentional; `UserConsent` is append-only.

PR #181, opened 2026-07-04
feat(accounts): gate verification SMS on recorded SMS consent

hugobessa:
## Summary
Phase 5 of the **SMS MFA Consent** plan (stacked on Phase 4 / PR #180). The authoritative, server-side guarantee: no verification SMS is ever dispatched to a user without a recorded `SMS_CONSENT` consent. The Phase 4 capture UX is convenience; **this gate is the enforcement**, and it fails closed.

- `AccountAdapter.send_verification_code_sms` now checks `consent_service.has_sms_consent(user)` (DI-injected) before dispatching. No consent → log + raise `ConsentRequiredError`, **zero** notification/Twilio calls. A consented user's OTP path is unchanged.
- This adapter method is the single choke point every headless phone-verification path funnels through (signup stage, resend, authenticated change-phone), so one guard covers them all.
- `ConsentRequiredError` (`accounts/exceptions.py`) multi-inherits `AccountError` + allauth's `ImmediateHttpResponse`, so allauth's already-installed `AccountMiddleware.process_exception` turns it into a clean **403** carrying `code="consent_required"` (allauth headless `APIResponse` envelope when a request context exists; plain `JsonResponse` fallback for direct/unit calls) — never an unhandled 500. The frontend routes to the consent step on that code.

## ⚠️ Compliance gap surfaced (blocks Phase 6, not fixed here)
Phase 5 gates **verification-code** SMS only. Two other adapter methods — `send_unknown_account_sms` and `send_account_already_exists_sms` (anti-enumeration messages) — still send SMS with **no consent check** and fire when there's no `user` to check against (a phone merely typed into a form). They go live the moment `ACCOUNT_PHONE_VERIFICATION_ENABLED` flips (Phase 6). That is the exact Twilio/TCPA concern this feature exists for, and it conflicts with Goal 4. Gating/removing them is a **product decision** (it changes enumeration-prevention UX), so it's recorded as a hard **Phase 6 prerequisite** + **Open Question** in the plan rather than silently shipped. **Do not flip the setting until it's resolved.**

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 5** (stacked on Phase 4). No feature flag. No migrations. `MFA_SUPPORTED_TYPES` unchanged (not an MFA factor).

## Test plan
- `docker compose run --rm api uv run pytest accounts/tests/test_sms_consent_gate.py accounts/tests/test_account_adapters.py -n auto` — unit: sends when consent present, refuses (asserts `create_notification` never called) when absent, **fails closed when `has_sms_consent` raises** (still zero dispatch); integration: real headless `POST .../account/phone` for a consent-less user → 403 `consent_required` + zero notifications, consented user → OTP dispatched once. `ACCOUNT_PHONE_VERIFICATION_ENABLED=True` is applied via `@override_settings` **in-test only** (production stays False until Phase 6).
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; full `pytest -n auto` → **3583 passed**.

### DIFF
--- a/accounts/account_adapters.py
+++ b/accounts/account_adapters.py
@@ -18,6 +18,8 @@ from vintasend.constants import NotificationTypes
 from vintasend.services.dataclasses import NotificationContextDict
 from vintasend.services.notification_service import NotificationService
 
+from accounts.exceptions import ConsentRequiredError
+from legal.services import ConsentService
 from organizations.exceptions import UserAlreadyHasMembershipError
 from organizations.models import get_active_organization_membership
 from organizations.services import OrganizationService
@@ -258,10 +260,12 @@ class AccountAdapter(DefaultAccountAdapter):
         *args,
         notification_service: Annotated[NotificationService, Provide["notification_service"]],
         organization_service: Annotated[OrganizationService, Provide["organization_service"]],
+        consent_service: Annotated[ConsentService, Provide["consent_service"]],
         **kwargs,
     ):
         self.notification_service = notification_service
         self.organization_service = organization_service
+        self.consent_service = consent_service
         super().__init__(*args, **kwargs)
 
     def send_password_reset_mail(self, user, email, context):
@@ -429,6 +433,12 @@ class AccountAdapter(DefaultAccountAdapter):
     def send_verification_code_sms(self, user, phone: str, code: str, **kwargs):
         """
         Sends a verification code.
+
+        Security-critical gate: no verification SMS is ever dispatched for a
+        user without a recorded SMS_CONSENT ``UserConsent`` row (fail closed).
+        This is the authoritative, server-side guarantee behind the consent
+        capture UX — see ``legal.services.ConsentService.has_sms_consent`` and
+        ``accounts.exceptions.ConsentRequiredError``.
         """
         if not user:
             logger.warning("No user provided for sending verification code SMS.")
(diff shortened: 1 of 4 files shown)


accounts/account_adapters.py
```diff
@@ -438,6 +448,13 @@ def send_verification_code_sms(self, user, phone: str, code: str, **kwargs):
             logger.warning("No phone number provided for sending verification code SMS.")
             return
 
+        if not self.consent_service.has_sms_consent(user):
```
hugobessa, line 451: The security guarantee. Placed AFTER the user/phone presence checks and BEFORE any
dispatch, with no try/except around it — so a DB/wiring error in `has_sms_consent`
propagates and blocks the send (fail-closed), never falls through to
`create_notification`. A dedicated test asserts zero dispatch even when the check
raises. `has_sms_consent` is consent-once-ever (any SMS_CONSENT row).

accounts/exceptions.py
```diff
@@ -7,3 +16,43 @@ class AccountError(Exception):
 class UserNotAuthenticatedError(AccountError):
     def __init__(self, message="User must be authenticated to generate an access token."):
         super().__init__(message)
+
+
+class ConsentRequiredError(AccountError, ImmediateHttpResponse):
```
hugobessa, line 21: Dual-inherits `AccountError` (domain-level, catchable in tests) + allauth
`ImmediateHttpResponse` (allauth's documented mid-adapter abort hook). allauth's
`AccountMiddleware.process_exception` catches it and returns the 403 regardless of
which headless view/stage triggered the SMS send — so one raise is both the
hard backstop and the clean-4xx seam; no per-view override needed.

accounts/account_adapters.py
```diff
@@ -429,6 +433,12 @@ def get_user_by_phone(self, phone: str):
     def send_verification_code_sms(self, user, phone: str, code: str, **kwargs):
```
hugobessa, line 433: NOTE for reviewers (see PR description + plan Open Questions): this gate covers
verification SMS only. `send_unknown_account_sms` / `send_account_already_exists_sms`
below still send anti-enumeration SMS to a raw phone with no consent check — a
compliance gap that goes live with Phase 6. Flagged as a hard Phase 6 prerequisite;
a product/legal decision, deliberately not silently changed here.

PR #183, opened 2026-07-04
feat(accounts,legal): phone-keyed SMS consent gating all SMS types

hugobessa:
## Summary
Phase 8 of the **SMS MFA Consent** plan (stacked on Phase 7 / PR #182). Closes the compliance gap flagged in Phase 5: the two **anti-enumeration** SMS (`send_unknown_account_sms`, `send_account_already_exists_sms`) had no consent gate because they carry no `user`. This makes SMS consent **phone-keyed** so all three SMS types are covered.

- `UserConsent.phone_number` (indexed) — recorded at signup (from `user.phone_number`) and via `POST /consents/` (new optional `phone_number`). Phones are **normalized** (E.164-ish, matching allauth's `PhoneField`) on both write and read so stored/queried values compare equal.
- **Verification SMS** (`send_verification_code_sms`) → `has_sms_consent_for_phone_and_user(phone, user)`: refuse with `403 consent_required` unless the requesting user recorded SMS consent for that phone.
- **Anti-enumeration SMS** → `has_sms_consent_for_phone(phone)`: **silently send nothing** (log + return, no error — preserves allauth's uniform enumeration-safe response) unless the phone's **owning account** consented.

## Security design (post-review, important for reviewers)
The first cut let any authenticated user POST an arbitrary `phone_number` to `/consents/` and thereby "consent" a victim's number. Fixed by tying consent to **phone ownership**:
- Anti-enumeration gate (`has_sms_consent_for_phone`) requires the consent row's own `user.phone_number == phone` — a fabricated row (attacker's user, victim's phone) can't satisfy it. Consequently `send_unknown_account_sms` is now a **structural no-op** (it only fires when no account owns the phone, so no owning-account consent can exist) — i.e. we never SMS a phone with no account, the safest outcome. `already_exists` sends only to a phone whose owning account previously consented.
- Verification gate is tied to the **same acting user** (not a strict `user.phone_number == phone` join) because allauth does **not** set `user.phone_number` before the gated send on add/change-phone (and rejects re-submitting the current phone) — a strict join would permanently 403 every legitimate add-phone. Verified empirically against the allauth flow.
- **Residual, accepted:** an authenticated user can self-consent a typed phone and receive one OTP to it — inherent to *any* phone verification (you must send one code to prove ownership), bounded by allauth's per-phone rate limit. Full closure (CAPTCHA / per-phone throttle on `/consents/`) is a noted follow-up; the repo has no `throttle_classes` precedent to follow.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 8** (stacked on Phase 7). Resolves the "anti-enumeration SMS are ungated" Open Question; clears the Phase 6 blocker. One additive migration (`legal.0003`). Frontend handoff doc updated.

## Test plan
- `docker compose run --rm api uv run pytest accounts/tests/ legal/tests/ -n auto` — ownership: a consent row by user A for phone Y does **not** satisfy the gate; format: a `"+1 555-555-0102"` consent then permits `send_verification_code_sms("+15555550102")`; anti-enumeration real-flow: unconsented phone → zero dispatch + no error, consented+owned → dispatch; verification gate per-(user,phone).
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; migration reverse/reapply clean; full `pytest -n auto` → **3612 passed**.


legal/managers.py
```diff
@@ -59,3 +59,59 @@ def has_sms_consent(self, user: "User") -> bool:
             .for_document_type(PolicyDocumentType.SMS_CONSENT)
             .exists()
         )
+
+    def has_sms_consent_for_phone(self, phone: str) -> bool:
```
hugobessa, line 63: Anti-enumeration gate: requires the consent row's own `user.phone_number == phone`
(`user__phone_number=phone`), so a `/consents/` row fabricated by an attacker for a
victim's number never satisfies it. A side effect: `send_unknown_account_sms` becomes
a structural no-op (a phone with no account has no owning-account consent), i.e. we
never SMS a phone with no account — the safest compliance outcome.

legal/managers.py
```diff
@@ -59,3 +59,59 @@ def has_sms_consent(self, user: "User") -> bool:
...
+
+        return (
+            self.get_queryset()
+            .for_phone(phone)
+            .filter(user__phone_number=phone)
+            .for_document_type(PolicyDocumentType.SMS_CONSENT)
+            .exists()
+        )
+
+    def has_sms_consent_for_phone_and_user(self, phone: str, user: "User") -> bool:
```
hugobessa, line 89: Verification gate is tied to the SAME acting user, NOT a strict `user.phone_number ==
phone` join. allauth doesn't set `user.phone_number` before the gated send on
add/change-phone (and rejects re-submitting the current phone), so a strict join would
permanently 403 legitimate add-phone. Empirically verified. Residual (accepted): a user
can self-consent + OTP a typed phone — inherent to phone verification, rate-limited.

accounts/account_adapters.py
```diff
@@ -479,11 +487,26 @@ def send_unknown_account_sms(self, phone: str | None, **kwargs) -> None:
...
+        gate. If `phone` has no recorded SMS_CONSENT, this is a silent no-op —
+        no SMS, no raised error — which preserves allauth's uniform,
+        enumeration-safe response (the caller can't distinguish "no consent"
+        from "SMS sent") while never sending an unconsented SMS.
         """
         if not phone:
             logger.warning("No phone number provided for sending unknown account SMS.")
             return
 
+        if not self.consent_service.has_sms_consent_for_phone(phone):
```
hugobessa, line 502: Anti-enumeration send: silent no-op (log + return, no exception) when the phone lacks
owning-account consent — deliberately silent so allauth's enumeration-prevention
response stays uniform (an error here would itself leak account existence). The phone
is redacted to last-4 in the log line (PII).

legal/models.py
```diff
@@ -83,11 +95,15 @@ class UserConsent(BaseModel):
     ip_address = models.GenericIPAddressField(null=True, blank=True)
     user_agent = models.TextField(blank=True, default="")
     source = models.CharField(max_length=16, choices=ConsentSource)
+    phone_number = models.CharField(max_length=20, blank=True, default="")
```
hugobessa, line 98: `phone_number` is normalized (spaces/hyphens stripped, matching allauth's PhoneField)
on write (`record_consent`) and read (`for_phone`) via `common/utils/phone_utils.py`,
so a human-formatted number posted to `/consents/` still matches the E.164 value
allauth hands the gate. `user` stays required; the phone check joins on it for ownership.

PR #184, opened 2026-07-04
feat(accounts): make ACCOUNT_PHONE_VERIFICATION_ENABLED env-driven (default off)

hugobessa:
## Summary
Phase 6 of the **SMS MFA Consent** plan (stacked on Phase 8 / PR #183). Converts the hardcoded `ACCOUNT_PHONE_VERIFICATION_ENABLED = False` (set in commit `62a3968` while Twilio validated the messaging profile) into an **environment variable**: `config("ACCOUNT_PHONE_VERIFICATION_ENABLED", cast=bool, default=False)`.

This is the rollout control the reactivation needs: **default `False`, so a plain deploy changes nothing**, and each environment (staging, prod) enables SMS phone verification independently by setting the env var — **no code change to flip**. The Phase 5/8 consent gate protects all SMS sends, so enabling per-environment is safe once Twilio approves that environment.

Wired through every layer (per the `add-env-var` skill): `vinta_schedule_api/settings/base.py`, `.env.example`, `.env.docker.example`, `render.yaml` (`integrations-credentials` group, `value: false` — operators flip in the Render dashboard), the CI workflow `env:` blocks (`'false'`, keeps CI off), and `ai-tools/AGENTS.md`.

## Behavioral nuance (reviewers please note)
`ACCOUNT_PHONE_VERIFICATION_ENABLED` gates allauth's **signup/login phone-verification stage**. allauth's authenticated `ManagePhoneView` (add/change phone) does **not** consult this flag — it always attempts to send, gated only by the Phase 5/8 SMS-consent check. So with the flag OFF, a *consented, authenticated* user could still receive an OTP via change-phone. This is pre-existing allauth behavior (not introduced here), and unconsented sends are still blocked by the consent gate — but the flag is **not a total SMS kill-switch**. If a full kill-switch is wanted, gating `ManagePhoneView`/the adapter on the flag too is a small follow-up (deliberately not scoped into this phase).

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 6** (stacked on Phase 8). No migrations. Remaining real-world gate to actually enable in an environment: Twilio approving that environment's messaging profile.

## Test plan
- `docker compose run --rm api uv run pytest accounts/tests/test_account_phone_verification_env_var.py accounts/tests/test_sms_consent_gate.py -n auto` — settings resolves the env var (unset → False; truthy strings → True; falsy → False); with the flag at its default False the phone endpoint stays consent-gated (consent-less → `403 consent_required`, zero dispatch); enabled-path integration (consented → OTP) already covered from Phase 5/8.
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; full `pytest -n auto` → **3626 passed**.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -168,6 +168,11 @@ envVarGroups:
         sync: false
       - key: TWILIO_DEFAULT_BROADCAST_NUMBERS
         sync: false
+      # Per-environment rollout gate for SMS phone verification. Default off;
+      # flip to true in the Render dashboard once Twilio approves this
+      # environment's messaging profile.
+      - key: ACCOUNT_PHONE_VERIFICATION_ENABLED
+        value: false
 
   # AWS S3 + CloudFront storage. Values come from the Terragrunt/Scalr stack
   # under infrastructure/ — see `terragrunt output`. Secrets are sync:false so
--- a/vinta_schedule_api/settings/base.py
+++ b/vinta_schedule_api/settings/base.py
@@ -347,7 +347,12 @@ HEADLESS_TOKEN_STRATEGY = "accounts.token_strategies.AccessAndRefreshTokenStrate
 ACCESS_TOKEN_EXPIRY_MINUTES = config("ACCESS_TOKEN_EXPIRY_MINUTES", cast=int, default=15)
 REFRESH_TOKEN_EXPIRY_DAYS = config("REFRESH_TOKEN_EXPIRY_DAYS", cast=int, default=30)
 ACCOUNT_LOGIN_METHODS = {"phone", "email"}
-ACCOUNT_PHONE_VERIFICATION_ENABLED = False
+# Rollout gate: default off while Twilio validates our messaging profile per
+# environment. An operator flips this in the environment (Render dashboard /
+# .env) once approved there — no code change required.
+ACCOUNT_PHONE_VERIFICATION_ENABLED = config(
+    "ACCOUNT_PHONE_VERIFICATION_ENABLED", cast=bool, default=False
+)
 ACCOUNT_PHONE_VERIFICATION_MAX_ATTEMPTS = 3
 ACCOUNT_PHONE_VERIFICATION_SUPPORTS_RESEND = True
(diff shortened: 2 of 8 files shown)


vinta_schedule_api/settings/base.py
```diff
@@ -347,7 +347,12 @@ def base_dir_join(*args):
 ACCESS_TOKEN_EXPIRY_MINUTES = config("ACCESS_TOKEN_EXPIRY_MINUTES", cast=int, default=15)
 REFRESH_TOKEN_EXPIRY_DAYS = config("REFRESH_TOKEN_EXPIRY_DAYS", cast=int, default=30)
 ACCOUNT_LOGIN_METHODS = {"phone", "email"}
-ACCOUNT_PHONE_VERIFICATION_ENABLED = False
+# Rollout gate: default off while Twilio validates our messaging profile per
+# environment. An operator flips this in the environment (Render dashboard /
+# .env) once approved there — no code change required.
+ACCOUNT_PHONE_VERIFICATION_ENABLED = config(
```
hugobessa, line 353: The rollout control. Default `False` means a plain deploy is byte-for-byte the prior
behavior (verification off); an operator enables it per environment by setting the env
var (Render dashboard / `.env`) once Twilio approves that environment — no code change,
no redeploy of new code. `cast=bool` accepts true/false/1/0/yes/no/on/off.

render.yaml
```diff
@@ -168,6 +168,11 @@ envVarGroups:
         sync: false
       - key: TWILIO_DEFAULT_BROADCAST_NUMBERS
         sync: false
+      # Per-environment rollout gate for SMS phone verification. Default off;
+      # flip to true in the Render dashboard once Twilio approves this
+      # environment's messaging profile.
+      - key: ACCOUNT_PHONE_VERIFICATION_ENABLED
```
hugobessa, line 174: Declared in the `integrations-credentials` envVarGroup (attached to web/worker/beat)
at `value: false`, matching the file's existing bool convention. Operators flip it to
`true` in the Render dashboard for the environment being turned on; it is not a secret
(no `sync: false`).

PR #185, opened 2026-07-04
feat(accounts): split signup consent into separate terms + SMS checkboxes

hugobessa:
## Summary
Phase 9 of the **SMS MFA Consent** plan (stacked on Phase 6 / PR #184). Twilio (and TCPA) require SMS consent to be its **own explicit, separate, unchecked** opt-in — not bundled with Terms/Privacy acceptance. Phase 4 shipped a single combined `accepted_policies` checkbox; this splits it into two **required** signup checkboxes:

- `accepted_terms` — "I agree to the Privacy Policy and Terms of Use." → records `PRIVACY_POLICY` + `TERMS_OF_USE`.
- `accepted_sms_consent` — "I agree to receive SMS text messages (e.g. verification codes) at the phone number I provide. Msg & data rates may apply." → records `SMS_CONSENT`.

Both are required today (phone MFA is mandatory), but the SMS recording is driven **specifically** by `accepted_sms_consent` (via an explicit `_CONSENT_FIELD_DOCUMENT_TYPES` field→doc-type map), so SMS could later become optional without touching terms recording. `schema-auth.yml` regenerated (headless `BaseSignup`: `accepted_policies` → `accepted_terms` + `accepted_sms_consent`, both required); handoff §3a updated with the two-checkbox copy.

**Frontend contract change:** the signup POST body now sends `accepted_terms: true` and `accepted_sms_consent: true` (was `accepted_policies: true`). Documented in the handoff.

## Plan reference
`ai-plans/2026-07-03-SMS_MFA_CONSENT_IMPLEMENTATION_PLAN.md` — **Phase 9** (stacked on Phase 6). No migrations. No behavior change to the consent gate — only how consent is collected at signup.

## Test plan
- `docker compose run --rm api uv run pytest accounts/tests/test_signup_consent.py accounts/tests/test_signup_consent_headless.py -n auto` — both checkboxes required (missing either → invalid/400, no user/consent created); both checked → all three doc types recorded; **directed tests prove SMS recording is driven by `accepted_sms_consent`, not `accepted_terms`** (would fail under an "any consent → record all" bug). Headless HTTP-level covered. All prior signup fixtures updated to the two fields.
- Outer gate: `check --deploy` clean; `makemigrations --check` clean; schema regenerated; full `pytest -n auto` → **3632 passed**.

### DIFF
--- a/accounts/base_forms.py
+++ b/accounts/base_forms.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Annotated
+from typing import Annotated, ClassVar
 
 from django import forms
 from django.utils import timezone
@@ -21,14 +21,18 @@ class BaseVintaScheduleSignupForm(forms.Form):
     """
     Base form for user signup.
 
-    Captures first_name, last_name, an optional organization_name, and a
-    required policy-acceptance acknowledgement. At signup time, the intended
+    Captures first_name, last_name, an optional organization_name, and two
+    required policy-acceptance acknowledgements. At signup time, the intended
     org name is persisted on Profile.pending_organization_name so it can be
     consumed during email-confirmation provisioning (Phase 3).
 
     When a non-expired, unaccepted OrganizationInvitation exists for the
     signup email, the org name is left blank — the user will auto-join the
     inviting org instead of creating a new one.
+
+    Consent acceptance is split into two separate fields (Phase 9) rather
+    than one combined checkbox: Twilio / TCPA require SMS consent to be its
+    own explicit, distinct opt-in, not bundled with Terms/Privacy acceptance.
     """
 
     first_name = forms.CharField(max_length=255, required=True, label="First Name")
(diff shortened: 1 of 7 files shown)


accounts/base_forms.py
```diff
@@ -38,17 +42,39 @@ class BaseVintaScheduleSignupForm(forms.Form):
...
-    accepted_policies = forms.BooleanField(
+    accepted_terms = forms.BooleanField(
         required=True,
-        label="I agree to the Privacy Policy, Terms of Use, and SMS messaging consent.",
+        label="I agree to the Privacy Policy and Terms of Use.",
+        error_messages={
+            "required": "You must accept the Privacy Policy and Terms of Use to create an account.",
+        },
+    )
+    accepted_sms_consent = forms.BooleanField(
```
hugobessa, line 52: The Twilio/TCPA requirement: SMS consent is its own explicit, separate, unchecked
opt-in with its own label + error message — not bundled with Terms/Privacy. Distinct
from `accepted_terms` above.

accounts/base_forms.py
```diff
@@ -38,17 +42,39 @@ class BaseVintaScheduleSignupForm(forms.Form):
...
+                "provide to create an account."
             ),
         },
     )
 
+    # Maps each consent checkbox field to the PolicyDocumentType values it
+    # covers. Keeping this as an explicit mapping (rather than iterating
+    # PolicyDocumentType.values directly) means SMS consent could later
+    # become optional without touching how terms/privacy are recorded.
+    _CONSENT_FIELD_DOCUMENT_TYPES: ClassVar[dict[str, list[str]]] = {
```
hugobessa, line 70: Explicit field→doc-type map (not inline conditionals): `accepted_terms` →
[PRIVACY_POLICY, TERMS_OF_USE], `accepted_sms_consent` → [SMS_CONSENT]. The recording
loop reads each field's `cleaned_data` value before recording its docs, so SMS is
recorded ONLY when its own box is checked. This structure lets SMS consent become
optional later without touching terms recording — a directed test locks the wiring in.

PR #189, opened 2026-07-18
feat(payments): move billing ownership from user to organization (billing plans and limits, phase 1)

hugobessa:
Phase 1 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md), implementing the spec at [ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_SPEC.md](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_SPEC.md).

This is the **one-way door** the spec names: billing moves from being owned by a `User` to being owned by an `Organization`. It lands first, deliberately, before any money can flow. Nothing else in the plan can be built until this is correct.

## Why this is safe to do destructively

`payments_billingprofile` and `payments_subscription` are rebuilt rather than migrated in place. That is sanctioned because both tables — along with `organizations_organizationtier` and `organizations_subscriptionplan` — are empty in every environment: nothing assigns `Organization.tier`, and the only writes to those models live in tests.

**Reviewer action:** please confirm emptiness in staging and production before this deploys. The migration fails loudly rather than losing rows if the assumption is wrong (`preserve_default=False` on a non-null re-added column), but the check is worth doing by hand.

## What changed

**Ownership move**
- `BillingProfile.user` (a `OneToOneField` primary key) becomes `BillingProfile.organization`.
- `Subscription` gains real `organization`, `plan`, `billing_state`, `billing_interval`, `current_period_start`/`_end`, `grace_period_ends_at`, `plan_external_id`, and `payment_provider` fields.
- `BillingProfileViewSet` mixes in `TenantScopedViewMixin` and resolves by active organization.

**Dead-seam repair.** These were pre-existing defects, not regressions introduced here:
- `Subscription.membership` was a bare type annotation with no field and no migration, so `Subscription.plan` raised `AttributeError` — which the `except ObjectDoesNotExist` around it never caught. Both removed; `plan` is now a real FK.
- `RefundStatusUpdate.status` used `PaymentStatuses` instead of `RefundStatuses`, and its `__str__` referenced a nonexistent `self.payment`.
- Two modules did `from venv import logger` instead of importing the stdlib `logging`.

**Payer identity.** `BillingProfile` gains `contact_first_name`, `contact_last_name`, `contact_email`, `contact_phone`. With billing no longer user-owned there is no `User` to source a payer from, and MercadoPago hard-400s on a null `payer.email`. These are provider-payload data and are deliberately distinct from the future `is_billing_owner` membership flag, which governs *who may manage* billing rather than *what identity is sent to the gateway*. The plan's Data Model Changes section was amended to record this.

## Reviewer notes

**Breaking change for API clients.** The billing-profile `id` field changes source from `user_id` to `organization_id`. The field name and type are unchanged, so this does **not** show up as a schema diff — a client that persisted the old `id` will silently read a different entity. This needs a client handoff before merge.

**Permissions widened, then narrowed.** Moving the resource from "a user's own billing data" to "the organization's tax document number" meant any member — including one invited a minute ago — could read and overwrite it. Write actions are now gated on `IsOrganizationAdmin`. The plan's `IsBillingOwnerOrAdmin` is Phase 9's job and is intentionally not built here.

**Scope taken on deliberately.** A minimal `BillingPlan` model is pulled forward from Phase 3 because `Subscription.plan` cannot reference a model that does not exist. Phase 3 extends it with `PlanLimit`/`PlanEntitlement` and seeding; no destructive redo is required. Its one DB constraint ships with a test rather than untested.

**`Subscription` carries two status concepts.** `status` (`SubscriptionStatuses`) is the provider-reported state fed by `SubscriptionStatusUpdate`; `billing_state` (`BillingState`) is the internal billing lifecycle Phase 10's state machine will drive. They share member names (`active`, `cancelled`, `pending`), so the boundary is documented in the model docstring and `status` is dropped from the serializer until a consumer needs it.

## Verification

Every gate was run and independently re-run by the orchestrator, not taken from the implementing agent's report:

- Full suite: **3693 passed**, 0 failed (baseline before this phase: 3679 — 14 new tests).
- `ruff check` / `ruff format --check`: clean.
- `mypy .`: 322 errors in 59 files, exactly matching the pre-existing baseline; zero in `payments/`.
- `makemigrations --check`: no changes detected.
- **Migration reverses**: verified forward → `migrate payments 0002` → forward again, both before and after the review fixes edited migration `0003` in place. The first implementation attempt failed on Postgres FK dependencies and had to be rewritten as drop-FK → `DeleteModel` → `CreateModel` → re-add-FK.

## Review findings addressed

An adversarial review raised three BLOCKERs, all fixed in `2368c27`:

1. **Broken provider payload** — the payer dataclass emitted a null email, which MercadoPago rejects with a hard 400. Every `create_payment`/`process_subscription` path would have failed against the live gateway while the test suite stayed green, because nothing tested the serialized payload.
2. **Webhook 500-loop** — an unguarded reverse 1:1 (`subscription.organization.billing_profile`) would raise `RelatedObjectDoesNotExist` on the unauthenticated provider webhook once Phase 4 gives every organization a subscription while most have no billing profile. Worth noting this is a Phase 1 change that only detonates on Phase 4's data — Phase 1's own tests could not have caught it.
3. **`request.organization` may be `None`** for an authenticated user with zero memberships, giving `IntegrityError` → 500 on write. A genuine regression introduced by this phase; the previous `request.user` was never `None`.

Plus: the cross-organization isolation test was vacuous (it passed for the trivial reason that no profile existed, and would have passed against the old user-keyed code) and has been replaced with real isolation, active-org-header switching, and non-member-org cases.

### DIFF
--- a/payments/migrations/0003_organization_owned_billing.py
+++ b/payments/migrations/0003_organization_owned_billing.py
@@ -9,6 +9,16 @@
 # an `organization` FK), and Postgres will not let a column backing a live FK constraint
 # be dropped without first dropping the dependent constraints. The reverse path recreates
 # the prior (`user`-owned) schema in the same way.
+#
+# The `billing_profile` FK on `payments_payment` and the `subscription` FK on
+# `payments_subscriptionstatusupdate` are dropped for the same reason (they point at
+# `payments_billingprofile` / `payments_subscription`) and re-added below. Both source
+# tables are transitively empty (they FK to the equally-empty tables being rebuilt), so
+# dropping and re-adding these columns loses no data either. The re-`AddField` calls at
+# `payment.billing_profile` and `subscriptionstatusupdate.subscription` below use
+# `preserve_default=False` on a non-null column: if either table were ever non-empty,
+# this makes the migration fail loudly (an `IntegrityError` on the implied backfill)
+# rather than silently defaulting existing rows to `NULL` and losing the association.
 import django.db.models.deletion
 import django.utils.timezone
 import model_utils.fields
@@ -127,6 +137,10 @@ class Migration(migrations.Migration):
                         to="organizations.organization",
                     ),
                 ),
+                ("contact_first_name", models.CharField(max_length=255)),
+                ("contact_last_name", models.CharField(blank=True, max_length=255)),
+                ("contact_email", models.EmailField(max_length=254)),
+                ("contact_phone", models.CharField(blank=True, max_length=50)),
                 ("document_type", models.CharField(max_length=50)),
                 ("document_number", models.CharField(max_length=50)),
                 (
@@ -281,7 +295,10 @@ class Migration(migrations.Migration):
             preserve_default=False,
         ),
         # --- Genuine pre-existing bug: RefundStatusUpdate.status used PaymentStatuses
-        #     instead of RefundStatuses. ---
+        #     instead of RefundStatuses. `payments_refundstatusupdate` is empty in every
+        #     environment (it FKs to `payments_refund`, which FKs to the equally-empty
+        #     `payments_payment`), so there is no pre-existing row whose status value
+        #     would fall outside the new (narrower) `RefundStatuses` choices. ---
         migrations.AlterField(
             model_name="refundstatusupdate",
             name="status",
--- a/payments/migrations/0003_organization_owned_billing.py
+++ b/payments/migrations/0003_organization_owned_billing.py
@@ -1,0 +1,300 @@
+# Phase 1 of billing plans and limits: move billing ownership from the user to the
+# organization, and repair the dead `Subscription.plan` seam.
+#
+# Destructive rebuild of `payments_billingprofile` and `payments_subscription` — both
+# tables are empty in every environment (no code writes to them today), so no data
+# migration is needed. Both tables (and the FK fields on `Payment` /
+# `SubscriptionStatusUpdate` that point at them) are dropped and recreated rather than
+# altered in place: `BillingProfile`'s primary key changes identity (from a `user` FK to
+# an `organization` FK), and Postgres will not let a column backing a live FK constraint
+# be dropped without first dropping the dependent constraints. The reverse path recreates
+# the prior (`user`-owned) schema in the same way.
+import django.db.models.deletion
+import django.utils.timezone
+import model_utils.fields
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("organizations", "0014_organization_external_event_update_policy"),
+        ("payments", "0002_initial"),
+    ]
+
+    operations = [
+        # --- Drop the FK fields that depend on the tables being rebuilt. ---
+        migrations.RemoveField(
+            model_name="payment",
+            name="billing_profile",
+        ),
+        migrations.RemoveField(
+            model_name="payment",
+            name="subscription",
+        ),
+        migrations.RemoveField(
+            model_name="subscriptionstatusupdate",
+            name="subscription",
+        ),
+        # --- Drop the tables being rebuilt. ---
+        migrations.DeleteModel(
+            name="Subscription",
+        ),
+        migrations.DeleteModel(
+            name="BillingProfile",
+        ),
+        # --- New catalog model the rebuilt Subscription.plan resolves to. ---
+        migrations.CreateModel(
+            name="BillingPlan",
+            fields=[
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
+                    ),
+                ),
+                (
+                    "created",
+                    model_utils.fields.AutoCreatedField(
+                        db_index=True,
+                        default=django.utils.timezone.now,
+                        editable=False,
+                        verbose_name="created",
+                    ),
+                ),
+                (
+                    "modified",
+                    model_utils.fields.AutoLastModifiedField(
+                        db_index=True,
+                        default=django.utils.timezone.now,
+                        editable=False,
+                        verbose_name="modified",
+                    ),
+                ),
+                ("meta", models.JSONField(blank=True, default=dict, verbose_name="meta")),
+                ("slug", models.SlugField(max_length=100, unique=True)),
+                ("name", models.CharField(max_length=255)),
+                ("is_active", models.BooleanField(db_index=True, default=True)),
+                ("is_default_for_new_organizations", models.BooleanField(default=False)),
+                ("monthly_price", models.DecimalField(decimal_places=2, max_digits=10)),
+                (
+                    "annual_price",
+                    models.DecimalField(blank=True, decimal_places=2, max_digits=10, null=True),
+                ),
+                ("currency", models.CharField(max_length=3)),
+                ("grace_period_days", models.PositiveIntegerField(blank=True, null=True)),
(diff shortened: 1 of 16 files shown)


payments/migrations/0003_organization_owned_billing.py
```diff
@@ -0,0 +1,317 @@
+# Phase 1 of billing plans and limits: move billing ownership from the user to the
```
hugobessa, line 1: This migration is destructive by design — it drops and recreates
`payments_billingprofile` and `payments_subscription` rather than altering
them in place. A straight field-level diff (what `makemigrations` proposes)
fails on Postgres because `Payment.billing_profile` and the subscription FKs
depend on the old primary-key column.

It also drops columns on `payments_payment` and
`payments_subscriptionstatusupdate`, which the header comment now documents.
Those tables are transitively empty. If that assumption is ever wrong, the
re-`AddField` uses `preserve_default=False` on a non-null column, so the
deploy fails loudly rather than silently losing rows.

Verified reversible: forward → `migrate payments 0002` → forward again.

payments/models.py
```diff
@@ -32,51 +30,109 @@ class BillingAddress(BaseModel):
...
+        "organizations.Organization",
+        primary_key=True,
+        on_delete=models.CASCADE,
+        related_name="billing_profile",
     )
+    # Payer identity sent to the payment gateway. Distinct from the future
+    # `OrganizationMembership.is_billing_owner` (Phase 9), which is about who may
+    # *manage* billing — these fields are about what the gateway needs to charge
+    # the organization (e.g. MercadoPago rejects a payer with no email).
+    contact_first_name = models.CharField(max_length=255)
```
hugobessa, line 84: These four `contact_*` fields are the answer to "who is the payer" now that
billing is organization-owned. They are deliberately **not** the same concept
as the `is_billing_owner` membership flag arriving in a later phase:
`is_billing_owner` decides who may press the purchase button, while these
decide what identity is transmitted to the payment provider.

They landed in this phase specifically because the table is already being
recreated here — adding them later would cost a second migration on the same
table for no benefit.

payments/services/payment_service.py
```diff
@@ -228,9 +240,20 @@ def receive_subscription_payment_update(
         payment_external_id = subscription_payment_data.external_id
         payment = self.get_payment_by_external_id(payment_external_id)
         if not payment:
+            billing_profile = BillingProfileModel.objects.filter(
```
hugobessa, line 243: This filtered lookup replaces an unguarded `subscription.organization.billing_profile`.
The reverse 1:1 raises `RelatedObjectDoesNotExist`, and this method is reachable
from the **unauthenticated** provider webhook — so once every organization has a
subscription (Phase 4) while most have no billing profile, the old code would
500 and the provider would retry forever.

Returning `None` with a warning mirrors the other early returns in this method.

payments/services/payment_service.py
```diff
@@ -247,14 +270,24 @@ def receive_subscription_payment_update(
         )
 
     def _serialize_subscription(self, subscription: SubscriptionModel) -> Subscription:
+        organization_billing_profile = BillingProfileModel.objects.filter(
```
hugobessa, line 273: Deliberate asymmetry with the guard above: this one raises rather than
returning `None`, because `_serialize_subscription` is typed to return a
non-optional `Subscription` and its callers (`process_subscription`,
`cancel_subscription`) are not wired to any live view or task today.

Widening the return type would mean touching two more methods on a code path
with no caller. If a later phase wires those up, revisit this.

payments/views.py
```diff
@@ -81,12 +86,39 @@ class BillingProfileViewSet(
...
+    #: authenticated member (IsAuthenticated, above).
+    write_actions = (
+        "create_billing_profile",
+        "update_billing_profile",
+        "partial_update_billing_profile",
+    )
+
+    def get_permissions(self):
+        if self.action in self.write_actions:
+            return [IsAuthenticated(), IsOrganizationAdmin()]
```
hugobessa, line 100: Write actions are gated on `IsOrganizationAdmin` because this phase changed
what the resource *is*: it went from "my own billing data" to "the
organization's tax document number". Without this, any member — including
someone invited a minute ago — could read and overwrite it.

This is intentionally not `IsBillingOwnerOrAdmin`; that permission and the
`is_billing_owner` flag it depends on belong to a later phase.

payments/models.py
```diff
@@ -32,51 +30,109 @@ class BillingAddress(BaseModel):
...
+    contact_phone = models.CharField(max_length=50, blank=True)
     document_type = models.CharField(max_length=50)
     document_number = models.CharField(max_length=50)
     billing_address = models.OneToOneField(
         BillingAddress, on_delete=models.CASCADE, related_name="billing_profile"
     )
 
     def __str__(self):
-        return f"{self.pk} {self.user} - {self.document_type} - {self.document_number}"
+        return f"{self.pk} {self.organization} - {self.document_type} - {self.document_number}"
```
hugobessa, line 95: `Subscription` carries two status-shaped fields and they are not
interchangeable — they even share member names (`active`, `cancelled`,
`pending`), which is exactly why the boundary is written down here.

`status` is what the provider told us, fed by `SubscriptionStatusUpdate`.
`billing_state` is our own billing lifecycle and is what the dunning /
grace / restricted state machine will drive in a later phase. Read the wrong
one and you will get plausible-looking but incorrect behavior.

payments/tests/views/test_billing_profile_view_set.py
```diff
@@ -39,33 +64,137 @@ def test_retrieve_billing_profile_success(self, auth_client, user):
...
+    def test_retrieve_billing_profile_not_found(self, auth_client, membership):
         """Test retrieving billing profile when it doesn't exist."""
         url = reverse("api:BillingProfile-retrieve")
         response = auth_client.get(url)
 
         assert response.status_code == status.HTTP_404_NOT_FOUND
 
+    def test_retrieve_billing_profile_no_membership(self, auth_client):
+        """A user with no active organization membership cannot resolve a billing
+        profile at all — the active org resolves to None and the lookup 404s."""
```
hugobessa, line 82: The previous version of this test was vacuous: org A had no billing profile
at all, so the endpoint 404'd for the trivial reason that nothing existed —
it would have passed byte-for-byte against the old user-keyed implementation.

It now gives both organizations distinct profiles and asserts the caller sees
only its own, plus an active-org-header switching case and a non-member-org
403 case.

payments/models.py
```diff
@@ -32,51 +30,109 @@ class BillingAddress(BaseModel):
...
+        return f"{self.id} {self.organization} - {self.city} - {self.state} - {self.country} - {self.zip_code}"
 
     @property
-    def user(self):
-        return getattr(self, "billing_profile", None) and self.billing_profile.user
+    def organization(self):
+        return getattr(self, "billing_profile", None) and self.billing_profile.organization
+
+
+class BillingPlan(BaseModel):
```
hugobessa, line 40: `BillingPlan` is pulled forward from the plan-catalog phase because
`Subscription.plan` cannot reference a model that does not exist yet.

A later phase extends this with `PlanLimit`, `PlanEntitlement`, and seeding —
additive only, no destructive redo. The `uniq_default_billing_plan` partial
constraint ships with its test here rather than arriving untested.

PR #190, opened 2026-07-19
feat(payments): authenticate provider webhooks and make them idempotent (billing plans and limits, phase 2)

hugobessa:
Phase 2 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). **Stacked on [phase 1](https://github.com/vintasoftware/vinta-schedule-api/pull/189)** — review that one first.

## Why this lands before anything trusts a webhook

`payment_update` and `subscription_payment_update` had no `authentication_classes`, no `permission_classes`, and no signature verification. Later phases make organization billing state — lifting limits, clearing a *restricted* state — depend on these endpoints. This phase closes that before anything reads them.

Worth noting how this was survivable until now: the pre-existing handlers used per-method `@inject` without declaring the `pk`/`provider` URL kwargs the router supplies, so they raised `TypeError` on any real HTTP request. The endpoints were not merely unauthenticated — they were non-functional. That independently corroborates Phase 1's finding that this seam was dead code.

## What changed

- **Signature verification.** `verify_mercadopago_signature` implements MercadoPago's real `x-signature` scheme: HMAC-SHA256 over the manifest `id:{data.id};request-id:{x-request-id};ts:{ts};`. Constant-time comparison via `hmac.compare_digest`. An empty `MERCADOPAGO_WEBHOOK_SECRET` fails **closed**.
- **Idempotency ledger.** New `ProviderWebhookEvent`, unique on `(provider, route, external_event_id)`. `route` is in the key because MercadoPago's payment and preapproval notification-id spaces are not documented as disjoint.
- **Provider registry.** The `provider` URL kwarg now actually selects an adapter, replacing the hardcoded MercadoPago factories in `di_core/containers.py`. This is the seam Stripe plugs into in phase 2b.
- **Status mappings.** The four empty dicts in the payment adapter and the empty `SUBSCRIPTION_STATUS_MAPPING` are filled *and wired in*. Previously `check_status` and `create_status_update_from_payment_payload` wrote raw provider strings straight into `choices`-constrained columns.
- `MERCADOPAGO_WEBHOOK_SECRET` and `WEBHOOK_SIGNATURE_TOLERANCE_SECONDS` added end-to-end via the `add-env-var` checklist.

## The security property, stated precisely

**MercadoPago's signature does not authenticate the request body.** It covers only `data.id`, `x-request-id`, and `ts`. Anyone can forge the payload.

That is tolerable only because status is always re-fetched from MercadoPago's API by id and never trusted from the body. The review traced every path reachable from both views to confirm this: `is_payment_update`/`receive_update` read `type`/`action` from the body but only to *gate* execution; every value that gets persisted comes from an `sdk.*.get()` response.

**Anyone extending this code must preserve that invariant.** If a future change reads a decision out of the webhook body, an attacker controls it.

## Review findings — two BLOCKERs, both fixed in `38f9f70`

**1. The idempotency ledger key was attacker-controlled.** `get_event_id` returned the payload's **top-level `id`**, which the signature does not cover. An attacker could capture one valid signature, hold `data.id` fixed so the HMAC still verified, and mutate the top-level `id` on every replay — each replay yielding a different `external_event_id`, so the ledger never deduped and the handler re-ran unbounded: one status-update row and one outbound provider API call per request.

This is worth dwelling on. Signature verification was correct. The ledger was correct. Their *composition* was broken, because the key was derived from unsigned material. The key is now `f"{data_id}:{request_id}:{ts}"` — verified manifest components only.

**2. No timestamp tolerance.** `ts` was folded into the manifest but never checked for freshness, so a captured `(body, signature, request-id)` triple verified forever. The ledger did **not** mitigate this, precisely because of BLOCKER 1. Proof it was absent: the original tests signed with `ts="1700000000"` — November 2023, ~2.5 years stale — and passed. Now rejected outside a 300s window (Stripe's convention, so the interface generalizes to phase 2b).

Also fixed: no `transaction.atomic()` around the ledger write and its effect (safe today only via `ATOMIC_REQUESTS = True` in `production.py`, which `test.py` and dev do not set — it would have broken when phases 5-8 add Celery callers); full MercadoPago API responses being `json.dumps`'d into logs and Sentry, including `payer.email`, name, `identification.number` (CPF/CNPJ) and billing address; a `provider` parameter that defaulted rather than required, which is the same shape as the bug this phase fixed; and no throttling on two unauthenticated endpoints that each trigger an outbound provider call.

## Verification

Every gate re-run independently by the orchestrator, not relayed from the implementing agents:

- Full suite **3756 passed** (phase 1 baseline 3693 → 3726 after implementation → 3756 after fixes).
- `ruff check` / `ruff format --check` clean; `mypy .` at the exact 322-error/59-file baseline, zero in `payments/`; `makemigrations --check` clean; migration forward/reverse verified.
- **The BLOCKER 1 regression test was proven, not assumed.** The vulnerability was temporarily reintroduced and `test_replayed_signature_with_mutated_notification_id_is_rejected` **failed**; restored, it passes. Only the payment-endpoint variant failed (only that adapter was reverted), confirming the test is specific to the defect rather than incidentally green.


payments/services/mercadopago_signature.py
```diff
@@ -0,0 +1,143 @@
...
+    keep ``data.id`` fixed (so the HMAC still verifies) and vary the top-level
+    ``id`` freely across replays.
+    """
+
+    data_id: str
+    request_id: str
+    ts: str
+
+    @property
+    def event_id(self) -> str:
```
hugobessa, line 45: This is the fix for the phase's most serious finding, and the reasoning is
worth keeping in view.

MercadoPago's HMAC covers `data.id`, `x-request-id`, and `ts` — it does *not*
cover the notification payload's top-level `id`. The first implementation used
that top-level `id` as the idempotency ledger key, so an attacker holding one
captured valid signature could keep `data.id` fixed (HMAC still verifies) and
vary the top-level `id` on every replay. Each replay looked like a distinct
event, so the ledger never deduped and the handler re-ran without bound.

Signature verification was correct. The ledger was correct. The composition
was broken. Any future provider adapter must derive its ledger key from
signed material only.

payments/services/mercadopago_signature.py
```diff
@@ -0,0 +1,143 @@
...
+        logger.warning("MercadoPago webhook signature mismatch")
+        return None
+
+    try:
+        ts_seconds = int(ts)
+    except ValueError:
+        logger.warning("MercadoPago webhook x-signature ts is not a valid integer: %r", ts)
+        return None
+
+    tolerance_seconds = getattr(
```
hugobessa, line 134: Without this window a captured `(body, x-signature, x-request-id)` triple
verifies forever — one leaked delivery becomes a permanent forgery capability.

The idempotency ledger does not substitute for it: before the key was moved
onto signed material, the ledger key was attacker-mutable, so replays were not
deduped either. The two fixes are load-bearing together.

300s matches Stripe's default, which matters because phase 2b implements
Stripe against this same interface — a `verify_signature` with no clock input
would not have generalized.

payments/views.py
```diff
@@ -22,57 +25,158 @@
...
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.payment_service = payment_service
+
     @extend_schema(
         summary="Receive payment updates",
-        description="This endpoint is used to receive payment updates from MercadoPago.",
+        description="This endpoint is used to receive payment updates from a payment provider.",
         request=None,
```
hugobessa, line 67: `authentication_classes = ()` and `AllowAny` are deliberate — a provider
cannot present our credentials. Authenticity comes from the HMAC instead.

Verified during review that this does not widen anything else: this is a bare
`ViewSet` defining exactly these two methods, so DRF's router generates no
list/retrieve/create routes. `BillingProfileViewSet` is a separate class and
is unaffected.

Note `raw_body = request.body` is captured *before* `request.data` is touched.
Reversing that order raises `RawPostDataException`, and would also risk
verifying a re-serialized payload rather than the wire bytes.

payments/services/payment_service.py
```diff
@@ -269,6 +307,90 @@ def receive_subscription_payment_update(
...
+    def handle_payment_webhook(
+        self, provider: str, raw_body: bytes, headers: Mapping[str, str], payload: dict
+    ) -> PaymentStatusUpdateModel | None:
+        """Idempotently process an inbound ``payment-update`` webhook notification.
+
+        Callers must call ``verify_payment_webhook_signature`` first — this method
+        does not re-verify authenticity, only idempotency + dispatch. Safe to call
+        more than once with the same provider event: a redelivery of an
+        already-processed event is a no-op.
+
```
hugobessa, line 341: The `transaction.atomic()` here is not belt-and-braces. Without it, the ledger
row and the effect it guards were atomic only because `production.py` sets
`ATOMIC_REQUESTS = True` — an implicit dependency that `test.py` and local dev
do not share.

That would have held right up until phases 5-8 introduce non-request callers
(Celery retries, replay commands), at which point concurrent delivery
double-processes: `get_or_create` commits immediately, the concurrent caller
reads `processed_at is None`, and both run the handler. `get_or_create_pending`
now also `select_for_update()`s the row it finds.

payments/tests/views/test_payment_webhooks.py
```diff
@@ -0,0 +1,414 @@
...
+            data=payload,
+            content_type="application/json",
+            **sign("mp-payment-456"),
+        )
+
+        assert response.status_code == status.HTTP_200_OK
+        assert ProviderWebhookEvent.objects.count() == 1
+        assert ProviderWebhookEvent.objects.get().processed_at is not None
+
+    def test_replayed_signature_with_mutated_notification_id_is_rejected(
```
hugobessa, line 283: This test was verified to actually catch the bug rather than merely pass: the
vulnerability was temporarily reintroduced and this test failed; with the fix
restored it passes.

The distinction matters here. The pre-existing idempotency test replays a
byte-identical body — the cooperative-provider case. It stayed green
throughout the vulnerability, because a well-behaved provider never mutates
the field an attacker would. Testing the honest path told us nothing about
the adversarial one.

payments/services/payment_adapters/mercadopago_payment_adapter.py
```diff
@@ -88,11 +126,23 @@ def check_status(
         self, payment_external_id: str, update_id: str | None = None
     ) -> PaymentStatusUpdate:
         response = self.sdk.payment().get(payment_external_id)
+        original_status = response["response"]["status"]
+        mapped_status = PAYMENT_STATUS_MAPPING.get(original_status, PaymentStatuses.UNKNOWN)
+        if mapped_status == PaymentStatuses.UNKNOWN:
+            logger.error(
```
hugobessa, line 132: This previously did `logger.error(..., json.dumps(response))`, sending the
entire MercadoPago payment response to logs and Sentry — `payer.email`,
first/last name, `identification.number` (CPF/CNPJ), and the full billing
address.

Now logs only the external id and the unmapped status, which is all that is
needed to diagnose an unrecognized provider status.

payments/models.py
```diff
@@ -204,3 +205,38 @@ class RefundStatusUpdate(BaseModel):
 
     def __str__(self):
         return f"{self.id} {self.refund} - {self.status} - {self.created.isoformat()}"
+
+
+class ProviderWebhookEvent(BaseModel):
```
hugobessa, line 210: `ProviderWebhookEvent` is unique on `(provider, route, external_event_id)` rather than `(provider, external_event_id)`.

MercadoPago does not document its payment and preapproval notification-id spaces as disjoint, so keying without `route` risks one route's event suppressing the other's. Nothing legitimately delivers the same event to both routes, so this is not over-scoped either.

Note the ledger key itself (`external_event_id`) is now derived entirely from signed manifest material — see the comment on `mercadopago_signature.py`. Keying it off unsigned payload data was this phase's most serious finding.

PR #191, opened 2026-07-19
feat(payments): add a Stripe adapter behind the provider abstraction (billing plans and limits, phase 2b)

hugobessa:
Phase 2b of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main` directly, since [phase 1](https://github.com/vintasoftware/vinta-schedule-api/pull/189) and [phase 2](https://github.com/vintasoftware/vinta-schedule-api/pull/190) are already merged.

Nothing routes any organization to Stripe — that is phase 9. This phase exists to **validate the abstraction against a real second provider**, on the plan's explicit instruction: where Stripe does not fit the MercadoPago-derived interface, change the interface and bring MercadoPago along.

It worked. The abstraction was wrong in ways only a second provider could expose.

## The bug the second provider found

`BasePaymentAdapter.receive_update` called the adapter's own `get_payment_external_id_from_update` hook — and then **immediately overwrote the result** with a hardcoded `payload["data"]["id"]`. The subscription base was worse: its hook was never called at all.

That is not merely wasteful. The hook call sat inside a `try/except` returning `None`, while the hardcoded helper raised **outside** any handler. So for any provider whose hook succeeds while `data.id` is absent — exactly Stripe, whose ids live at `data.object.id` — the old code raised uncaught and 500'd the webhook. MercadoPago never hit it only because its hook was byte-identical to the deleted helper.

The base classes now use the hook's return value. Behavior-preserving for MercadoPago (pinned by a new test), correct for everyone else.

## Interface changes, and the MercadoPago consequence

| Change | Why Stripe forced it | MercadoPago impact |
|---|---|---|
| `Plan.billing_interval` required | Stripe's `Price.recurring.interval` is explicit | MercadoPago hardcoded `frequency=1, frequency_type="months"`; now derives it (annual → 12 months, since it has no `"years"`) |
| `refund() -> RefundResult` (was `-> str`) | Both providers return status synchronously with the id | Removed a pointless second round trip |
| `check_refund_status(refund)` (was `(refund_external_id)`) | MercadoPago has no refund-by-id endpoint, only `list_all(payment_id)` | **Fixed a real bug** — see below |
| `verifies_full_body: bool` | MercadoPago signs a narrow manifest; Stripe signs the whole body | Makes the asymmetry explicit and now load-bearing |

**The `check_refund_status` change surfaced a genuine pre-existing bug.** The old MercadoPago implementation called `self.sdk.payment().get(refund_external_id)` — a *payment* lookup using a *refund* id — then mapped the resulting payment status through `REFUND_STATUS_MAPPING`. Unreachable today, wrong whenever reached. Now uses `refund().list_all()` and filters by id.

## Review found the adapter was written against an obsolete Stripe API

Three BLOCKERs, one root cause. `stripe==15.3.1` pins `api_version = 2026-06-24.dahlia`, in which `Invoice.subscription`, `Invoice.payment_intent`, and `PaymentIntent.charges` no longer exist. The adapter read all three.

The consequence was not subtle: `is_payment_update` accepts only `invoice.paid`/`invoice.payment_failed`, so **every** Stripe subscription webhook would have returned `None`, bailed, and — because of the base-class change above — still been `mark_processed`-ed, permanently burning the delivery so a redelivery after any fix would be a no-op.

**The tests passed the whole time**, because every fixture was hand-written to the same obsolete shape the implementation assumed. A test that encodes the implementation's assumption rather than the provider's contract manufactures confidence instead of providing it. All fixtures were rewritten from SDK introspection, and every field path in this PR was verified by introspecting the installed SDK — then independently spot-checked by the orchestrator.

Also fixed: `mark_processed` no longer runs when a handler returns `None` (so a failed delivery stays redeliverable); `verifies_full_body` became enforced rather than decorative; the conformance suite gained public-method-parity checks and subscription-adapter coverage; and a `Mock` that made an untested error branch look tested.

## Known gap — please read before merging

**No Stripe credentials exist in this environment**, so fixtures are derived from the SDK's type annotations, not captured from live events. That is a real improvement over hand-written guesses and is still **not** the same as verification.

Specifically unconfirmed: whether `Invoice.payments` is populated by default or requires `expand=["latest_invoice.payments"]`. The `Optional[...]` typing and doc-link imply expand is needed, and that is what the code does — but nobody has seen a real response.

**What this phase does and does not establish.** It establishes the interface generalizes: "no interface method exists that only one provider can implement" is now enforced by the conformance suite. It does **not** establish that the Stripe adapter works. The first real test is phase 9's sandbox run.

## Verification

- Full suite **3908 passed** (2b baseline 3883 → 3908 after fixes; phase 2 was 3756).
- `ruff` clean; `mypy .` at the exact 322-error/59-file baseline; `makemigrations --check` clean; migration `0005` forward/reverse verified (initially inconclusive — the DB was still at `0004`, so it was re-run properly).
- `stripe` license verified **MIT** from installed package metadata, not from the agent's assertion. Not in the project's forbidden SPDX list.
- Every Stripe field path re-introspected against the installed SDK by the orchestrator.


payments/services/payment_adapters/base.py
```diff
@@ -20,6 +18,18 @@
 class BasePaymentAdapter:
     provider: str
```
hugobessa, line 20: This is the finding that justified naming a second provider up front.

The old `receive_update` called `get_payment_external_id_from_update` and then
overwrote the result on the very next line with a hardcoded
`payload["data"]["id"]`. The subscription base never called its hook at all.

The sharp edge: the hook call was inside a `try/except` that returns `None`,
but the hardcoded helper raised *outside* any handler. Any provider whose hook
succeeds while `data.id` is absent — Stripe, whose ids are at
`data.object.id` — got an uncaught raise and a 500. MercadoPago was immune
only by coincidence: its hook was byte-identical to the deleted helper.

No amount of reading this code with one provider in the tree would have shown
it. That is the argument for building the abstraction against something real.

payments/services/subscription_adapters/stripe_subscription_adapter.py
```diff
@@ -0,0 +1,330 @@
...
+                f"Cannot update payment token for subscription {subscription.id} "
+                "with no external_id"
+            )
+        stripe.Subscription.modify(
+            subscription.external_id,
+            default_payment_method=payment_token,
+            api_key=self.api_key,
+        )
+
+    def get_subscription_external_id_from_update(self, update_payload: dict) -> str | None:
```
hugobessa, line 213: `stripe==15.3.1` pins api_version `2026-06-24.dahlia`, where
`Invoice.subscription` no longer exists — the id moved to
`parent.subscription_details.subscription`.

The original implementation read the removed field, so every Stripe
subscription webhook would have returned `None` and bailed. Verified by
introspection rather than memory:

    'subscription' in stripe.Invoice.__annotations__  -> False
    'parent' in stripe.Invoice.__annotations__        -> True
    'subscription' in stripe.Invoice.Parent.SubscriptionDetails.__annotations__ -> True

The legacy fallback is retained for older API versions.

payments/services/payment_adapters/base.py
```diff
@@ -20,6 +18,18 @@
...
+    #: end to end (e.g. Stripe's ``Stripe-Signature`` covers
+    #: ``{timestamp}.{raw_body}`` in full) or only a narrower manifest carved out
+    #: of specific fields (e.g. MercadoPago's ``x-signature`` covers only
+    #: ``data.id`` + ``x-request-id`` + ``ts``, never the body as a whole). This is
+    #: load-bearing, not decorative: a passing ``verify_signature`` does **not** by
+    #: itself imply every field in the parsed payload is trustworthy — that only
+    #: holds when this is ``True``. Every concrete adapter must set this
+    #: explicitly rather than lean on a base default, so the answer is never left
+    #: to tribal knowledge about a specific provider's signing scheme.
+    verifies_full_body: bool
```
hugobessa, line 31: `verifies_full_body` started as a bare annotation nothing read at runtime — a
trust marker that could be set wrong with nothing to catch it, which is worse
than no marker because it reads as reassurance.

It is now load-bearing: `get_event_id` raises unless a `False` adapter
overrides it, so the idempotency ledger key can never be silently derived
from unsigned material. That is deliberately narrower than "refuse all
payload fields when False" — MercadoPago's `data.id` *is* signed (it is in
the manifest), so a blanket rule would misdescribe its actual trust boundary.

payments/tests/services/subscription_adapters/test_stripe_subscription_adapter.py
```diff
@@ -0,0 +1,526 @@
+import hashlib
```
hugobessa, line 1: Every fixture here was rewritten. The originals were hand-written to the
pre-dahlia invoice shape — the same obsolete shape the implementation
assumed — so they asserted the implementation's own mistake back at itself
and stayed green through three BLOCKERs.

Worth stating as a general rule: a fixture that encodes what the code expects,
rather than what the provider actually sends, manufactures confidence instead
of providing it. These are now derived from SDK type introspection.

**Still not live-captured** — there are no Stripe credentials in this
environment. See the known-gap section of the PR description.

payments/services/payment_adapters/mercadopago_payment_adapter.py
```diff
@@ -10,7 +10,7 @@
 
 from payments.constants import PaymentProviders, PaymentStatuses, RefundStatuses
 from payments.exceptions import ProviderWebhookEventIdMissingError
-from payments.services.dataclasses import Refund
+from payments.services.dataclasses import Refund, RefundResult
```
hugobessa, line 13: `check_refund_status` previously called `self.sdk.payment().get(refund_external_id)`
— a *payment* lookup using a *refund* id — and then mapped the resulting
payment status through `REFUND_STATUS_MAPPING`.

A real pre-existing bug, unreachable today, wrong the moment it were reached.
It surfaced only because Stripe forced the signature to change: MercadoPago
has no refund-by-id endpoint at all, only `refund().list_all(payment_id)`,
which is why the method now takes the whole `Refund` rather than a bare id.

payments/tests/services/test_provider_registry.py
```diff
@@ -0,0 +1,424 @@
+"""Adapter conformance suite.
```
hugobessa, line 1: The conformance suite is what enforces this phase's acceptance criterion: no
interface method exists that only one provider can implement.

It originally enumerated only `@abstractmethod`s, which missed
`update_subscription_payment_token` — a public method only MercadoPago had.
It now also asserts public-method parity across adapters of the same kind,
and runs the signature/idempotency assertions against subscription adapters,
not just payment adapters.

The adapter class lists stayed hand-maintained for a concrete reason:
`container.wire(...)` imports this module during `django.setup()` before the
container is assigned, so a module-scope registry access crashes startup.
A consistency test asserts the lists match the DI registrations instead, so
drift fails loudly.

payments/services/payment_service.py
```diff
@@ -200,18 +200,24 @@ def create_refund(
             description="Refund created in the database, will send to payment gateway",
         )
         try:
-            refund.external_id = self.payment_gateway.refund(
+            refund_result = self.payment_gateway.refund(
```
hugobessa, line 203: `mark_processed` now runs only when the handler returns a result.

This matters because of an interaction introduced earlier in this phase. The
base classes used to raise on a missing id; the raise propagated, the view
500'd, and `transaction.atomic()` rolled the ledger row back — so a provider
retry could still succeed. Changing that raise to `return None` silently
turned every failed delivery into a *permanently consumed* one: marked
processed, never redeliverable, invisible.

Combined with the Stripe field-path bugs, every Stripe subscription webhook
would have been burned on first delivery and unrecoverable even after a fix.

PR #192, opened 2026-07-19
feat(payments): add plan catalog, limits, and entitlements (billing plans and limits, phase 3)

hugobessa:
Phase 3 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main`, which already has [phase 1](https://github.com/vintasoftware/vinta-schedule-api/pull/189) and [phase 2](https://github.com/vintasoftware/vinta-schedule-api/pull/190). Independent of [phase 2b](https://github.com/vintasoftware/vinta-schedule-api/pull/191).

The catalog every later phase reads. No user-visible behavior.

## This migration is the feature's kill switch

There is no feature flag in this plan — that was a deliberate decision, recorded in the plan's **Guiding Decisions**. The project has no feature-flag module, and building one for this feature alone would duplicate what the catalog already provides.

Instead: the seeded **`unlimited`** plan carries a `PlanLimit` row for *every* `LimitedResource` with `limit_value = NULL`, every entitlement enabled, and `is_default_for_new_organizations = True`. Enforcement code from phase 6 onward will run but cannot block anyone, because no organization has a ceiling until it is deliberately migrated onto a real plan. Rollback for any organization is a plan change, not a deploy.

Everything about this PR follows from that. It is why the seeding is tested more heavily than a data migration usually would be, and why the review's one substantive finding mattered.

## What changed

- `PlanLimit` (keyed `(plan, resource_key)`, carrying `limit_value` / `kind` / `overage_unit_price`) and `PlanEntitlement` (keyed `(plan, entitlement_key)`), both hanging off the minimal `BillingPlan` that phase 1 pulled forward.
- `LimitedResource`, `LimitKind`, `Entitlement` added to the existing `payments/billing_constants.py`.
- Seed migration creating `unlimited` and `free`.
- `BillingPlanFactory` replacing `OrganizationSubscriptionPlanFactory`, wired into `di_core/containers.py`.
- **Deleted**: `OrganizationTier`, `SubscriptionPlan`, `Organization.tier`, and `organizations/organization_subscription_plan_factory.py`.

The deleted factory is worth a note: it implemented `create_subscription_plan`, while the protocol and `PaymentService` both call `make_plan_from_subscription`. It also queried `SubscriptionPlan.objects.filter(subscription=...)` on a field that never existed. That is why the path was dead rather than merely unused, and `BillingPlanFactory` actually satisfies the protocol.

## Review finding: the seeding had to converge, not just fill gaps

`get_or_create(defaults=...)` applies `defaults` **only on creation**. If a `BillingPlan` with slug `unlimited` already existed — from a partial deploy, a manual fix, or an earlier test run — with `is_default_for_new_organizations=False`, the migration would silently no-op on that field.

The failure mode is quiet and bad: new organizations would not land on the unlimited plan, enforcement would be live for them, and nothing would raise. For a migration whose entire job is to be the safety net, filling gaps is not enough — it has to converge. All three models now use `update_or_create`, with a test that corrupts an existing plan and asserts the seeding corrects it.

## Verification

Re-run independently by the orchestrator, not relayed from the agents:

- Full suite **3771 passed** (main baseline 3756; +15 new).
- `ruff` clean; `makemigrations --check` clean; `check --deploy` clean.
- `mypy` **308 errors / 58 files**, *down* from main's 322/59 — attributable to deleted dead code, with no `type: ignore` or `noqa` added anywhere in the diff (checked).
- **Acceptance verified directly against the database**: `unlimited` has 8 `PlanLimit` rows for 8 `LimitedResource` members, all `limit_value` NULL, `is_default_for_new_organizations=True`.
- **Round-trip verified**: `free 8/5/False` + `unlimited 8/5/True` → reverse deletes both → re-apply reproduces byte-identical state. Both the `payments` seed and the `organizations` deletion migration reverse and re-apply cleanly. This matters more than usual — with no feature flag, this migration is the rollback path.
- `grep` for `OrganizationTier` / `organizations.SubscriptionPlan` outside migrations: zero hits.

## Carried forward

- **`free` plan values are placeholders.** 5 members, 3 resource calendars, 2 groups, 1 bundle calendar, 5 availability windows, 1 webhook subscription, 0 partner-API system users, 50 event occurrences at $0.05 overage; only Google calendar entitlement enabled. Product supplies the real numbers before phase 14 (which is deferred for exactly this reason).
- **`billing_day` gap**, marked in-code for phase 9: it is derived from `current_period_start.day` because no standalone field survives, so a cycle starting on the 29th-31st yields a value most providers reject for monthly recurrence. Clamping is a phase 9 decision, not made here.


payments/migrations/0007_seed_billing_plans.py
```diff
@@ -0,0 +1,131 @@
+# Phase 3 of billing plans and limits: seed the plan catalog. There is no feature
```
hugobessa, line 1: This migration is the closest thing this feature has to a feature flag.

The plan deliberately does not add one — the project has no flag module, and
the catalog already provides the same lever. The `unlimited` plan seeded here
has a NULL `limit_value` for every limited resource and is the default for new
organizations, so every enforcement phase that follows runs but cannot block
anyone until an organization is deliberately moved off it.

That is also the rollback path: reverting enforcement for an organization is a
plan change, not a deploy. Which is why this file gets more test coverage than
a data migration normally would, and why the `update_or_create` change below
was the review's one substantive finding.

payments/migrations/0007_seed_billing_plans.py
```diff
@@ -0,0 +1,131 @@
...
+    Entitlement.ADVANCED_SCHEDULING: False,
+}
+
+
+def seed_billing_plans(apps, schema_editor):
+    BillingPlan = apps.get_model("payments", "BillingPlan")
+    PlanLimit = apps.get_model("payments", "PlanLimit")
+    PlanEntitlement = apps.get_model("payments", "PlanEntitlement")
+
+    unlimited_plan, _created = BillingPlan.objects.update_or_create(
```
hugobessa, line 53: `update_or_create`, not `get_or_create`, and deliberately so.

`get_or_create(defaults=...)` applies `defaults` only when it creates the row.
If a `BillingPlan` with this slug already existed — partial deploy, manual
fix, leftover test data — with `is_default_for_new_organizations=False`, the
original code silently no-opped on that field.

New organizations would then not land on the unlimited plan, enforcement would
be live for them, and nothing anywhere would raise. For a migration whose only
job is to be the safety net, filling gaps is not sufficient; it has to
converge on every run. Same reasoning applied to `PlanLimit` (a wrong
`limit_value` or `kind` breaks enforcement) and `PlanEntitlement`.

payments/tests/test_plan_seed_migration.py
```diff
@@ -0,0 +1,108 @@
...
+from payments.models import BillingPlan
+
+
+@pytest.mark.django_db
+class TestPlanSeedMigration:
+    def test_unlimited_plan_exists_and_is_default(self):
+        plan = BillingPlan.objects.get(slug="unlimited")
+
+        assert plan.is_active is True
+        assert plan.is_default_for_new_organizations is True
```
hugobessa, line 25: This test enumerates `LimitedResource.values` dynamically rather than checking
a hardcoded list, which is the point of it.

If a later phase adds a limited resource to the enum without adding a matching
`unlimited` seed row, that resource would have no NULL-limit row — meaning it
would be enforced from the moment its check landed, with no way to roll back
per-organization. This test fails on both the count and the missing key in
that case, so the gap cannot ship silently.

payments/services/subscription_plan_factory/billing_plan_factory.py
```diff
@@ -0,0 +1,37 @@
...
+            else plan.monthly_price
+        )
+        return CreatedPlan(
+            id=plan.pk,
+            name=plan.name,
+            value=value,
+            currency=plan.currency,
+            # The day-of-month the provider bills on. Derived from when the current
+            # period started rather than stored separately — `Subscription` carries
+            # no standalone `billing_day` field.
```
hugobessa, line 32: Known gap, marked here because this is where someone will hit it.

`billing_day` is derived from `current_period_start.day` because no standalone
field survives — that one lived on the deleted `organizations.SubscriptionPlan`.
A cycle starting on the 29th-31st therefore produces a billing day most
providers reject or mishandle for monthly recurrence.

Deliberately not clamped here: choosing between clamping, rejecting, and
normalizing is a phase 9 decision, and guessing now would bake in the wrong
one. Inert until phase 9 drives a real provider subscription through this.

organizations/models.py
```diff
@@ -81,29 +81,12 @@ class ExternalEventUpdatePolicy(models.TextChoices):
     FORBIDDEN = "forbidden", "Updates are forbidden"
```
hugobessa, line 83: `OrganizationTier`, `SubscriptionPlan`, and `Organization.tier` are deleted
here — the first time this plan reaches outside `payments/`.

Safe because they were never wired to anything: nothing ever assigned
`Organization.tier` (null on every row), and the only writes to these models
lived in tests. `organization_subscription_plan_factory.py` goes with them —
it implemented `create_subscription_plan` while the protocol and
`PaymentService` both call `make_plan_from_subscription`, and it queried a
`SubscriptionPlan.subscription` field that never existed. Dead, not merely
unused.

Grep for these names outside migrations now returns nothing, and review
separately checked the things a grep misses: `apps.get_model` strings, the
GraphQL surface, admin fieldsets, serializers, fixtures, and reverse
accessors.

PR #193, opened 2026-07-19
feat(payments): place every organization on a plan (billing plans and limits, phase 4)

hugobessa:
Phase 4 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main`, which now has phases [1](https://github.com/vintasoftware/vinta-schedule-api/pull/189), [2](https://github.com/vintasoftware/vinta-schedule-api/pull/190), [2b](https://github.com/vintasoftware/vinta-schedule-api/pull/191) and [3](https://github.com/vintasoftware/vinta-schedule-api/pull/192).

Establishes the plan's core invariant: **every organization always has exactly one active plan — there is no plan-less state.**

Every billing-root organization gets a `Subscription` with its own copy of the plan's limits and entitlements. Reseller children never hold one; they pool against their root. Existing organizations are backfilled onto `unlimited`, never `free`, because the rollout itself must not block anyone.

This PR took two review rounds and found six BLOCKERs. That is worth reading about, because most of them were invisible from any single vantage point.

## What changed

- `SubscriptionPlanLimit` / `SubscriptionEntitlement` — per-subscription copies written on creation and on plan change. Catalog edits never propagate; `is_overridden=True` rows survive plan changes untouched. This is the support lever, and it is *why* the plan has no `bypass_limits` escape hatch.
- `SubscriptionService` with `create_subscription_for_organization` and `change_plan`.
- `is_billing_root(org)` / `billing_root_filter()` — one predicate, used by the service, the backfill migration, and the acceptance query.
- `resolve_billing_root` — cycle-guarded walk up `parent` to the nearest reseller.
- `OrganizationMembership.is_billing_owner` (field only; the permission is phase 9).
- Backfill migration placing every existing billing root on `unlimited`.
- Admin: `SubscriptionPlanLimit` / `SubscriptionEntitlement` inlines with `is_overridden` handling.

## Round 1 — four BLOCKERs, all one root cause

**"Is this organization a billing root?" was defined twice, and the definitions disagreed.** The backfill said *parent is null*. `resolve_billing_root` said *parent is null OR `can_invite_organizations`*. A **nested reseller** — the flag set, with a parent — fell in the gap: its whole subtree resolved to a root with no subscription. Now one predicate serves all three call sites.

The other three followed from the same area:

- **Django admin was an unhooked fourth creation path.** The implementer's report said "exactly three"; `OrganizationAdmin` had no `save_model`, so adding an org through admin produced a plan-less one directly.
- **The cycle guard returned an arbitrary node.** `A→B→C→A` returned `C`, which has a parent and is therefore never backfilled. The test asserted `result.pk in (a, b)` — it passed while the invariant was broken. Now raises `BillingRootCycleError`.
- **This phase broke phase 3's rollback.** `Subscription.plan` is `on_delete=PROTECT`, so once the backfill created subscriptions, the seed migration's reverse raised `ProtectedError`. Fixed by stamping `meta={"backfilled_by": "payments.0009"}` and having the reverse delete exactly those rows.

## Round 2 — the fixes created a new BLOCKER

A verification review ran after the round-1 fixes. It found two more, and the first is the interesting one:

**Fixing BLOCKER 1 built the trigger for a gap BLOCKER 4's fix did not cover.** `is_billing_root` now includes `can_invite_organizations` — and that flag is toggleable on an *existing* organization through the admin's Reseller Capability fieldset. But `save_model` returned early on `change=True`. So flipping that toggle on a GraphQL-created child (correctly subscription-less) instantly produced a billing root with no subscription: the acceptance query goes non-zero, and phase 5's `get_effective_limit` receives a subscription-less root.

Neither review could have caught this alone. It existed only in the interaction between two fixes.

**And the `is_overridden` fix did not work.** Unchecking the box makes `form.has_changed()` true, so the row landed in `formset.save(commit=False)` and was re-stamped `True`. The docstring asserted it was clearable; it was provably one-way, and an admin unchecking it got a silent no-op. Nobody had exercised it. That matters beyond tidiness — a one-way override turns every temporary support grant permanent.

Also fixed: a `Subscription` with zero limit rows was reachable; `change_plan` was not atomic while `create` was; an admin DI failure silently left an org plan-less (the same shape as the round-1 BLOCKER that was fixed by raising); and a "concurrency" test that was byte-identical to the test above it and exercised no concurrency.

## Verification

- Full suite **3969 passed**.
- `ruff` clean; `mypy` at the 308-error/58-file baseline; `makemigrations --check` clean.
- **Migration chain verified from a fresh schema**, not an incrementally-migrated one: `DROP SCHEMA public CASCADE` → migrate to head → reverse `payments` to `0004` → forward again. All five billing migrations apply and reverse in order, including the seed reverse succeeding *after* the backfill removes its stamped rows.
- **Both round-2 BLOCKER fixes were proven, not assumed.** The `is_overridden` guard was neutralised and its test **failed**; restored, it passes. (This is the same fix that was previously claimed working and was not — hence proving it rather than trusting the report.)

## Migration renumbering

This phase's migrations moved `0007/0008` → `0008/0009` after the Stripe adapter merged and pushed the plan catalog to `0006/0007`. Worth noting one hazard that produced: `BACKFILL_META_VALUE` had to move with it. A stale constant would make the reverse silently delete nothing — and the seed migration's reverse would then fail on `PROTECT` again, quietly reintroducing a fixed BLOCKER.

## Carried forward

- `payments/services/payment_service.py` `create_subscription` is now an unreconciled second creation path — every billing root already has a `Subscription`, so its unconditional `objects.create` would raise `IntegrityError` on the OneToOne. Tests-only today, so latent. Marked in-code and left for phase 9 rather than rewritten here.
- `0009` imports `billing_root_filter` from a live service module rather than freezing it. The risk assessment holds (it is a pure `Q` over two fields present since `0001`), but a future rename of `can_invite_organizations` or `parent` would retroactively change a historical migration's behavior. Noted in AGENTS.md.


payments/services/subscription_service.py
```diff
@@ -0,0 +1,264 @@
...
+    Subscription,
+    SubscriptionEntitlement,
+    SubscriptionPlanLimit,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+def is_billing_root(organization: Organization) -> bool:
```
hugobessa, line 25: This predicate is the fix for the phase's first BLOCKER, and the reason it
exists as a single named function rather than an inline condition.

"Is this organization a billing root?" was previously answered in two places
with two different answers: the backfill migration checked `parent__isnull`,
while `resolve_billing_root` short-circuited on `can_invite_organizations`.
A **nested reseller** — the flag set, with a parent — was a billing root by
one definition and not the other, so it was never backfilled, and its entire
subtree resolved to a root with no subscription.

Three call sites now share this: the service, the backfill migration, and the
acceptance query. If a fourth appears, it uses this too — a divergent copy is
how the original bug happened.

organizations/admin.py
```diff
@@ -58,6 +127,37 @@ class OrganizationAdmin(admin.ModelAdmin):
         ),
```
hugobessa, line 127: `save_model` fires on **edit as well as create**, and that is deliberate.

The first round of fixes made `is_billing_root` include
`can_invite_organizations`. That flag is toggleable on an existing
organization through the Reseller Capability fieldset — so flipping it turns
a subscription-less child into a billing root with no subscription, breaking
the phase's invariant and handing phase 5 a root it cannot resolve limits for.

An earlier version returned early when `change=True` and missed exactly this.
The call is idempotent (`get_or_create`) and a no-op for non-roots, so running
it on every save is both safe and necessary.

Worth noting how this bug arose: fixing one BLOCKER created the precondition
for another's gap. Neither review round could have found it alone.

payments/admin.py
```diff
@@ -78,6 +116,40 @@ class SubscriptionAdmin(admin.ModelAdmin):
     list_filter = ("status", "billing_state", "billing_interval", "payment_provider")
     search_fields = ("organization__name", "external_id")
     readonly_fields = ("created", "modified")
```
hugobessa, line 118: This reads `formset.changed_objects` rather than stamping everything returned
by `formset.save(commit=False)`, because the previous version made
`is_overridden` impossible to clear.

Unchecking the box makes `form.has_changed()` true, so the row was returned by
`save(commit=False)` and immediately re-stamped `True`. The override was
one-way and the docstring claimed otherwise — an admin unchecking it got a
silent no-op with no feedback.

That is not cosmetic. Support raising limits is *why* this plan has no
`bypass_limits` escape hatch, so a one-way override turns every temporary
grant into a permanent entitlement that survives all future plan changes.

Verified by neutralising this guard and confirming the test fails, then
restoring it — the previous fix was claimed working and was not.

payments/services/subscription_service.py
```diff
@@ -0,0 +1,264 @@
...
+def billing_root_filter() -> Q:
+    """``Q``-object equivalent of ``is_billing_root``, for queryset filtering where
+    per-instance iteration is infeasible (bulk backfill migration, acceptance
+    query). Keep in sync with ``is_billing_root``.
+    """
+    return Q(parent__isnull=True) | Q(can_invite_organizations=True)
+
+
+def resolve_billing_root(organization: Organization) -> Organization:
+    """Resolve the organization whose ``Subscription`` pays for ``organization``.
```
hugobessa, line 53: Raises `BillingRootCycleError` rather than returning whatever node the walk
stopped on.

`parent` is user-mutable and nothing in the codebase enforces acyclicity —
`assert_target_in_subtree` guards subtree membership, not cycles. In a cycle
no organization is parent-less, so the previous code returned an arbitrary
node that had a parent, was never backfilled, and had no subscription.

The old test asserted `result.pk in (a, b)` and passed the entire time the
invariant was broken. It documented the bug instead of catching it. The
admin form now also rejects a cycle at the point one would be created.

payments/migrations/0009_backfill_unlimited_subscriptions.py
```diff
@@ -0,0 +1,164 @@
...
+from payments.services.subscription_service import billing_root_filter
+
+
+BATCH_SIZE = 500
+UNLIMITED_PLAN_SLUG = "unlimited"
+
+# Stamped onto every `Subscription` this migration creates so its reverse can
+# delete exactly those rows and no others — see `delete_backfilled_subscriptions`.
+BACKFILL_META_KEY = "backfilled_by"
+BACKFILL_META_VALUE = "payments.0009"
```
hugobessa, line 42: `BACKFILL_META_VALUE` must match this migration's own name, and it moved when
the file was renumbered `0008` → `0009` after the Stripe adapter merged.

It is what the reverse filters on to delete exactly the rows this migration
created. A stale value makes the reverse silently delete nothing — and then
the seed migration's reverse fails on `PROTECT`ed `Subscription.plan`, which
is the precise BLOCKER this stamping was introduced to fix.

So a routine renumber came within one constant of quietly reintroducing a
fixed bug. The full chain was re-verified from a fresh schema afterwards.

payments/migrations/0009_backfill_unlimited_subscriptions.py
```diff
@@ -0,0 +1,164 @@
...
+
+    plan_limits = list(PlanLimit.objects.filter(plan=unlimited_plan))
+    plan_entitlements = list(PlanEntitlement.objects.filter(plan=unlimited_plan))
+
+    now = timezone.now()
+    period_end = now + datetime.timedelta(days=30)
+
+    last_pk = 0
+    while True:
+        batch_ids = list(
```
hugobessa, line 71: Backfills onto **`unlimited`**, never `free`.

Existing organizations must not acquire real ceilings at rollout — objective 4
sets rollout-caused blocking at zero. `unlimited` has a NULL `limit_value` for
every resource, so enforcement code from phase 6 onward runs but cannot block
anyone until an organization is deliberately migrated onto a real plan.

Note this uses `billing_root_filter()` rather than `parent__isnull=True`;
the latter is what skipped nested resellers.

PR #194, opened 2026-07-20
feat(payments): resolve effective limits and pooled usage (billing plans and limits, phase 5)

hugobessa:
Phase 5 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main`, which already has phases [1](https://github.com/vintasoftware/vinta-schedule-api/pull/189), [2](https://github.com/vintasoftware/vinta-schedule-api/pull/190), [2b](https://github.com/vintasoftware/vinta-schedule-api/pull/191), [3](https://github.com/vintasoftware/vinta-schedule-api/pull/192) and [4](https://github.com/vintasoftware/vinta-schedule-api/pull/193).

No user-visible behavior. This is the engine every enforcement phase from 6a onward calls: given an organization and a resource, what is the ceiling, and what is current usage — pooled across the whole reseller subtree.

The plan flagged this phase as the one where **a lock taken on the wrong row silently permits overshoot**. Overshoot has no exception and no failing test; it is just an organization quietly above its ceiling. That is why this took three review rounds.

## What changed

- **`payments/services/entitlement_service.py`** — `get_effective_limit`, `get_current_usage`, `check_limit`, `has_entitlement`, plus `USAGE_COUNTERS`, one counter per `LimitedResource` member with totality pinned by a test.
- **`SubscriptionAddOn`** + migration `0010`. Add-on quantity adds to the plan's ceiling.
- **`OverLimitError`** + a DRF exception-handler entry rendering the plan's over-limit contract.
- **Seven counters moved onto managers/querysets** — `OrganizationMembership.occupying_a_seat`, `OrganizationInvitation.pending`, `Calendar.live_of_type`, `AvailableTime.only_user_authored`, `WebhookConfiguration.live`, and new `public_api/{managers,querysets}.py` for `SystemUser.live`.
- **`assert_plan_is_complete`** in `subscription_service.py`, enforced on both paths that place a subscription on a plan.

`resolve_billing_root` / `is_billing_root` are **reused from Phase 4**, not redefined. A duplicated billing-root predicate is precisely the BLOCKER that cost Phase 4 two review rounds.

## Three BLOCKERs, each proven by reverting the fix

**1. Over-limit rejections committed their own transaction.** The handler returned a `Response` without `set_rollback()`. Under production's `ATOMIC_REQUESTS`, everything a guarded service wrote before the check — the `OrganizationInvitation` row, the `is_active` flip in `reactivate`, the audit entries Phase 6a writes ahead of its guard — would persist while the client got a 402. Proven through a real request, not a direct handler call.

**2. The availability counter locked organizations out below their real usage.** Editing one occurrence of a recurring window *inserts* an `AvailableTime` row. An org that made 3 recurring windows and edited 3 occurrences read as 6 against a limit of 5 — blocked while genuinely under, which the spec's goal 6 forbids outright.

**3. A plan downgrade granted an infinite ceiling.** `_sync_limits` deleted rows the new plan omits; `get_effective_limit` reads an absent row as *unlimited*. Each half is individually correct and separately documented — the delete revokes a grant the org stopped paying for, the fail-open stops a missing seed row locking anyone out. Composed, they invert the meaning of a downgrade.

This is the third defect in this plan that lived in a **composition** rather than in any single piece, after Phase 2's signature-plus-ledger and Phase 4's two disagreeing billing-root predicates.

## The fix for BLOCKER 3 took two attempts — worth reading

**Round 1** changed the delete to a *retain*, reasoning that a retained row is a bounded stale ceiling: bad, but not uncapped.

**That was wrong for the state that actually occurs.** Every organization is on `unlimited` for the entire rollout, and `unlimited`'s rows are all `limit_value=None`. Retaining a NULL row *is* an unlimited ceiling. The original outcome survived the fix intact — and the new test used `limit_value=3`, so it sailed straight past the case that mattered.

**Round 2** put the invariant in code instead of inferring a ceiling. `assert_plan_is_complete` refuses any plan missing a `LimitedResource` row, before any write, on both `change_plan` and `create_subscription_for_organization`; `BillingPlan.clean` and the admin inline formset raise the same condition at authoring time, so a support admin sees it in the admin rather than an end user seeing it mid-request.

Two tempting alternatives were rejected on the record:

- **Materializing an omission as `limit_value=0`** blocks an organization on a resource nobody agreed to restrict — the spec's goal 6 again.
- **Defaulting to unlimited** is the bug.

An incomplete plan is a catalog authoring error, so it fails loudly at the point it is introduced or used.

## Reviewer-facing notes

**Tests changed in rounds 2 and 3.** Round 2 replaced the round-1 downgrade test with four (NULL-row — the case round 1 missed, finite-row, creation-path, overridden-row) and made every `TestChangePlan` fixture build complete plans, since incomplete ones are now refused. Round 3 then caught that this fixture change had made the overridden-row assertion vacuous: `make_complete_plan` carries `organization_members` at `limit_value=0`, so `.exists()` passed whether or not the override survived. It now asserts the value is unchanged.

**The admin opts out of `BillingPlan.clean`, deliberately.** `ModelForm._post_clean` runs `full_clean()` before the inline formset saves, so validating there rejects the very edit that adds the missing row — leaving a gappy plan permanently unfixable through the UI. The formset validates the resulting row set instead. Even a saved incomplete plan is still refused at use time.

**`has_entitlement` fails closed while limits fail open.** `_sync_entitlements` *deletes* rows for entitlements a plan lacks, so absence encodes revocation; failing open there would hand every paid feature to every organization whose plan omits it.

## Carried forward

- ⚠️ **Phase 6a must call `check_seat_limit_for_invitation_accept(invitation)`**, not `check_limit`. Accepting is net zero on seats; the generic entry point makes an organization unable to fill its last seat.
- ⚠️ **Residual gap in the availability counter.** Editing *or cancelling* the **first** occurrence truncates the master and creates a replacement series row with no link back — the branch never reads `is_cancelled`, so both behave identically — and it compounds: once per operation, **unbounded**. Closing it needs a column, not a filter. Documented on `AvailableTimeQuerySet.only_user_authored`.
- `SystemUser.organization` is nullable, so an org-less system user is entirely unmetered. Pinned by a test so whoever makes the column non-nullable revisits it.
- The add-on aggregate has no period/expiry filter, so a one-time add-on raises the ceiling forever. Owned by Phase 9, which introduces one-time purchases.
- `UsageSnapshot` from the plan's data-model section remains deferred; rationale recorded in `billing_dataclasses.py`.

## Verification

Re-run independently by the orchestrator, not relayed from the agents:

- Full suite **4054 passed** (`main` baseline 3969; +85).
- `mypy` **305 errors / 57 files** — *below* the 308/58 baseline, with **zero `type: ignore` / `noqa` added** anywhere in the diff. One mypy error from overriding `_post_clean` was resolved by moving to the public `clean()` hook rather than suppressing it.
- `ruff check` + `format --check` clean; `makemigrations --check` clean; `check --deploy` unchanged.
- Migration `0010` verified forward and reverse. No further migration in the phase.
- Each BLOCKER fix confirmed by reverting it and watching the specific test fail.


common/exception_handlers.py
```diff
@@ -0,0 +1,41 @@
...
+    renders the same dict through its own error extension, so the two surfaces
+    stay byte-identical without either restating the shape.
+    """
+    if isinstance(exc, OverLimitError):
+        # Mandatory before returning a Response: swallowing the exception here
+        # would otherwise commit the ATOMIC_REQUESTS transaction, persisting
+        # whatever a guarded service wrote before it hit the limit check (an
+        # invitation row, a membership reactivation, audit entries) while the
+        # client is told 402.
+        set_rollback()
```
hugobessa, line 39: **BLOCKER 1.** Without this call the 402 response *commits* the request transaction under production's `ATOMIC_REQUESTS`.

Phase 6a guards `invite_user_to_organization`, `accept_invitation` and `reactivate`. Anything those write before reaching the guard — the invitation row, the `is_active` flip, audit entries — would persist while the client is told the request was rejected.

DRF's own handler calls `set_rollback()` for every `APIException`; returning a `Response` from a custom handler silently opts out of that. The regression test drives a real request rather than calling the handler directly, because a direct call cannot observe the transaction at all.

payments/services/entitlement_service.py
```diff
@@ -0,0 +1,554 @@
...
+            )
+            return EffectiveLimit(
+                resource_key=resource_key, limit_value=None, kind=None, overage_unit_price=None
+            )
+
+        limit = subscription.limits.filter(resource_key=resource_key).first()
+        if limit is None:
+            logger.debug(
+                "Subscription %s has no SubscriptionPlanLimit row for %s; treating it as "
+                "unlimited (fail-open).",
```
hugobessa, line 266: **Fail-open, and it is deliberate.** A resource with no `SubscriptionPlanLimit` row is unlimited, never zero — a missing seed row must not be able to lock an organization out.

Note the asymmetry with `has_entitlement`, which fails *closed*. That is not an inconsistency: `_sync_entitlements` deletes rows for entitlements a plan lacks, so for entitlements absence encodes revocation, and failing open would hand every paid feature to every organization whose plan omits it.

This fail-open is also half of what became BLOCKER 3 — see the comment on `assert_plan_is_complete`.

payments/services/entitlement_service.py
```diff
@@ -0,0 +1,554 @@
...
+                subscription.pk,
+                resource_key,
+            )
+            return EffectiveLimit(
+                resource_key=resource_key, limit_value=None, kind=None, overage_unit_price=None
+            )
+
+        if limit.limit_value is None:
+            # Unlimited plus any amount of purchased capacity is still unlimited;
+            # skip the add-on aggregate entirely rather than adding to NULL.
```
hugobessa, line 276: The NULL check short-circuits **before** the add-on aggregate deliberately. Unlimited plus purchased capacity is still unlimited; summing into a NULL ceiling is how "unlimited" silently becomes a finite number.

payments/services/entitlement_service.py
```diff
@@ -0,0 +1,554 @@
...
+            reviewer can see rather than a kwarg nobody can grep for. Only
+            meaningful for ``organization_members``; passing it with any other
+            ``resource_key`` raises, since it would otherwise be silently ignored.
+        """
+        _reject_inapplicable_invitation_exclusion(resource_key, exclude_invitation_id)
+        root = resolve_billing_root(organization)
+        if lock:
+            # Discard the returned row: the point is the row lock, and the
+            # subscription read below goes through the same transaction.
+            Subscription.objects.select_for_update().filter(organization=root).first()
```
hugobessa, line 411: The lock is taken on the **root's** subscription, not the calling organization's. That is the whole point: usage pools across the reseller subtree, so two children racing for the last unit must serialize on the same row.

The plan called this out as the failure mode that silently permits overshoot — a lock on the caller's own row would look correct and serialize nothing.

`check_limit` is only meaningful if the caller performs its create inside the same transaction. That contract is documented but unenforced; Phase 6a's call sites are where it has to be honored.

payments/services/entitlement_service.py
```diff
@@ -0,0 +1,554 @@
...
+        allowed = current_usage + delta <= ceiling
+        return LimitCheckResult(
+            allowed=allowed,
+            resource_key=resource_key,
+            current_usage=current_usage,
+            ceiling=ceiling,
+            remedy=(None if allowed else self._resolve_remedy_for(subscription, effective_limit)),
+        )
+
+    def check_seat_limit_for_invitation_accept(
```
hugobessa, line 439: ⚠️ **Phase 6a must call this on the accept path, not `check_limit`.**

Accepting an invitation is net zero on seats — the pending invitation disappears as the membership appears — but both count toward `organization_members`. Calling the generic entry point with `delta=1` means an org at 4 members + 1 pending reads as 5 against a ceiling of 5 and can never accept its own last invitation.

This exists as a named method rather than an `exclude_invitation_id` kwarg precisely so that omitting it is a missing *call* a reviewer can see, rather than a missing kwarg nobody can grep for. `lock` also defaults to `True` here, unlike `check_limit`.

payments/services/subscription_service.py
```diff
@@ -82,6 +86,35 @@ def resolve_billing_root(organization: Organization) -> Organization:
     return organization
 
 
+def assert_plan_is_complete(plan: BillingPlan) -> None:
```
hugobessa, line 89: **This is the fix for BLOCKER 3, and it took two attempts.**

The bug: `_sync_limits` deleted rows the new plan omits, and `get_effective_limit` reads an absent row as unlimited. Each half is correct alone; composed, a downgrade *uncapped* the resource.

Round 1 changed the delete to a retain, reasoning a retained row is a bounded stale ceiling. That was wrong for the dominant state — every org is on `unlimited`, whose rows are `limit_value=None`, and retaining a NULL row is an unlimited ceiling. The original outcome survived, and the round-1 test used `limit_value=3` so it never exercised it.

So the invariant moved into code. Two alternatives were rejected: materializing an omission as `limit_value=0` blocks an org on a resource nobody agreed to restrict (spec goal 6), and defaulting to unlimited is the bug itself. An incomplete plan is a catalog authoring error and now fails loudly, before any write, on both paths that put a subscription on a plan.

payments/services/subscription_service.py
```diff
@@ -219,11 +259,34 @@ def _sync_limits(self, subscription: Subscription, plan: BillingPlan) -> None:
...
                 unique_fields=["subscription", "resource_key"],
             )
-        # A resource_key with no PlanLimit row on the new plan, and never
-        # hand-overridden, is a grant the organization no longer pays for — leaving
-        # it behind on a downgrade would silently keep an entitlement (or a NULL /
-        # unlimited row) alive. Overridden rows are exempt: the support lever must
-        # survive a plan change untouched.
+        self._prune_stale_limits(subscription, plan_resource_keys)
+
+    def _prune_stale_limits(
```
hugobessa, line 264: This delete is now safe **only because `assert_plan_is_complete` guarantees the plan covers every `LimitedResource`** — so anything surviving the `exclude` is by construction a retired key that nothing consults.

If that guard is ever weakened, this line silently becomes BLOCKER 3 again: a deleted row for a live resource reads as unlimited. Rows predating the invariant are safe, since a stale row for a live key is in `plan_resource_keys` and gets overwritten by the upsert rather than deleted.

payments/admin.py
```diff
@@ -16,29 +19,112 @@
...
     RefundStatusUpdate,
     Subscription,
+    SubscriptionAddOn,
     SubscriptionEntitlement,
     SubscriptionPlanLimit,
     SubscriptionStatusUpdate,
 )
 
 
+class PlanLimitInlineFormSet(BaseInlineFormSet):
```
hugobessa, line 29: Authoring-time half of the completeness invariant, so a support admin sees the gap here rather than an end user hitting `IncompleteBillingPlanError` mid-request.

The check is skipped for a saved plan being **deactivated**. Without that exemption, an already-incomplete plan could not be retired at all until every missing row was added — the one edit you most want during an incident. The exemption is one-directional and tested: *activating* an incomplete plan is still rejected.

payments/admin.py
```diff
@@ -16,29 +19,112 @@
...
+            )
+
+
 class PlanLimitInline(admin.TabularInline):
     model = PlanLimit
+    formset = PlanLimitInlineFormSet
     extra = 0
     fields = ("resource_key", "limit_value", "kind", "overage_unit_price")
 
+    def get_extra(self, request: HttpRequest, obj: BillingPlan | None = None, **kwargs: Any) -> int:
```
hugobessa, line 73: Pre-renders one blank inline row per missing resource, so repairing a gappy plan is a single pass rather than N "Add another" clicks.

payments/admin.py
```diff
@@ -16,29 +19,112 @@
...
+            "is_active",
+            "is_default_for_new_organizations",
+            "monthly_price",
+            "annual_price",
+            "currency",
+            "grace_period_days",
+        )
+
+    def clean(self) -> dict[str, Any] | None:
+        self.instance.skip_limit_coverage_validation = True
```
hugobessa, line 117: **The one deliberate hole in the completeness validation, and why it is safe.**

`ModelForm._post_clean` runs `instance.full_clean()` *before* the inline formset is saved. So `BillingPlan.clean` firing here would reject the very edit that adds the missing row — leaving a gappy plan permanently unfixable through the admin.

The flag is set in exactly this one place; `BaseModel` never calls `full_clean` on save and nothing else builds a ModelForm or serializer over `BillingPlan`, so it cannot be reached elsewhere. The formset validates the resulting row set (existing − deleted + added) instead, which is strictly more accurate than validating the pre-save instance. And the runtime guard still refuses an incomplete plan at use time regardless.

payments/services/billing_dataclasses.py
```diff
@@ -0,0 +1,71 @@
...
+    """The ceiling that actually applies to one resource for one billing root.
+
+    ``limit_value`` is the subscription's ``SubscriptionPlanLimit.limit_value``
+    plus the quantity of every active ``SubscriptionAddOn`` for the same resource.
+    **``None`` means unlimited, never zero** — a resource with no
+    ``SubscriptionPlanLimit`` row at all also resolves to ``None``, so a missing
+    seed row cannot lock an organization out of a resource it used to be able to
+    create (the fail-open rule from the plan's Phase 5 body).
+
+    Deliberately carries no ``current_usage`` field, unlike the sketch in the
```
hugobessa, line 32: Deliberate deviation from the plan's data-model sketch, which put `current_usage` on this type. Folding a subtree-wide multi-table count into every ceiling lookup makes cheap calls expensive, and `check_limit` needs the two separately anyway.

Related: `LimitCheckResult.current_usage` is `int | None`, where `None` means *not measured* rather than zero. The unlimited path deliberately does not count — Phase 6a's "an unlimited org sees no change in query count" test depends on that.

calendar_integration/querysets.py
```diff
@@ -813,6 +824,43 @@ class AvailableTimeQuerySet(BaseOrganizationModelQuerySet, RecurringQuerySetMixi
     Custom QuerySet for AvailableTime model to handle specific queries.
     """
 
+    def only_user_authored(self) -> "AvailableTimeQuerySet":
```
hugobessa, line 827: **BLOCKER 2.** Editing an occurrence of a recurring availability window *inserts* an `AvailableTime` row, so counting every row reported usage above reality and blocked organizations that were genuinely under their limit — the spec's goal 6 forbids the rollout itself blocking anyone.

⚠️ **A residual gap survives, and it is wider than a filter can close.** When the edited occurrence is the **first** one, the master is truncated and a replacement series row is created with no link back. The branch never reads `is_cancelled`, so *cancelling* the first occurrence does the same thing. And it compounds — each subsequent first-occurrence edit or cancel adds another unlinked row, so the over-count is once per operation and unbounded, not "by one".

Closing it needs a column, not a filter. Size the schema change accordingly.

PR #195, opened 2026-07-20
feat(organizations): enforce seat limits on every path that adds a member (billing plans and limits, phase 6a)

hugobessa:
Phase 6a of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main`, which now has phases [1](https://github.com/vintasoftware/vinta-schedule-api/pull/189), [2](https://github.com/vintasoftware/vinta-schedule-api/pull/190), [2b](https://github.com/vintasoftware/vinta-schedule-api/pull/191), [3](https://github.com/vintasoftware/vinta-schedule-api/pull/192), [4](https://github.com/vintasoftware/vinta-schedule-api/pull/193) and [5](https://github.com/vintasoftware/vinta-schedule-api/pull/194).

This was developed stacked on phase 5; #194 merged while it was in review, so it targets `main` directly. `main` contains only the merge commit beyond this branch — the content diff between them is empty, so no rebase was needed.

The first phase that can actually block a real user. Every path that raises an organization's seat count now checks the ceiling, and every surface renders the same error.

## Two failure modes, pulling in opposite directions

This phase is unusual in that being *too* permissive and being *too* strict are both defects, and the second is the one that bit hardest here:

- **An unmetered path** — any way to add a member that skips the guard defeats the plan's objective 1.
- **A false block** — an organization prevented from doing something it is entitled to. The spec's goal 6 says no organization may be blocked as a consequence of the rollout itself.

Review found one of each, and both were invisible from inside the code the phase set out to change.

## What changed

- `invite_user_to_organization`, `accept_invitation`, and the invite branch of `provision_tenant_for_user` check seats inside their existing `transaction.atomic` blocks.
- `reactivate` is guarded via a new `OrganizationService.reactivate_membership` — it raises the seat count without being a create, and would otherwise be unmetered.
- All guarded methods take `bypass_limits: bool = False`, per the plan's **Enforcement bypass** decision. It is unreachable from every request surface.
- `OverLimitError` renders **byte-identically** through REST and GraphQL, asserted by comparing the two responses against each other rather than against a shared literal.

The accept paths call `check_seat_limit_for_invitation_accept`, not the generic `check_limit`. Accepting is net zero on seats — the pending invitation stops counting as the membership starts — so the generic entry point would leave an organization unable to fill its last seat.

## BLOCKER 1 — seat limits wedged signup with a 500

The phase guarded `provision_tenant_for_user` correctly. But two allauth adapters call it, and **allauth headless mounts as plain Django views, not DRF** — so `common/exception_handlers` never ran and `OverLimitError` escaped as a 500.

The failure compounds nastily. Under `ATOMIC_REQUESTS` the 500 rolls the request back, which also undoes the email-verification write — so the user's address stayed unverified and **every retry 500'd identically**. The account was wedged with no message. Social signup was worse: the whole user row rolled back, so Google signup simply failed.

Both adapters now fall through to the membership-less gated branch they already use for uninvited users. The invitation stays pending, so the user is recoverable once a seat frees — asserted, not assumed.

## BLOCKER 2 — resending an invitation was a false block

`POST /invitations/{id}/resend/` ran the guard while the invitation being resent still counted as pending. At the exact ceiling it returned 402 for an operation that **creates nothing**.

The triggering state is the common one: an org that just filled its seats, with one invite that never arrived. The only workaround was revoke-then-reinvite. The implementer had recorded this as an accepted corner case; review disagreed, and it was right — the machinery to fix it already existed one method over. A resend now excludes its own invitation, exactly like an accept.

## Review notes

**Three rounds.** Round 1 found the two BLOCKERs above plus the `reactivate` guard sitting in the viewset (contradicting the plan's "Service layer, not viewsets" decision, and leaving it with no `bypass_limits`), and `accept_invitation` marking the invitation accepted *outside* its transaction — a window where a seat double-counts, permanently if that write fails.

**Round 2 flagged a finding that turned out to be wrong, and the fixer caught it.** It reported the same atomicity bug in `provision_tenant_for_user`. That method carries a method-level `@transaction.atomic()` that `accept_invitation` lacks, so the bug did not exist there. The fixer stash-tested it, found the new test green against pre-fix code, applied the (harmless) clarifying move anyway, and said plainly that the test is not a regression guard rather than claiming a red→green. Recorded here because reviewer output needs verifying too.

**The GraphQL rollback is deliberate and worth knowing about.** graphql-core catches resolver exceptions and always returns 200, so nothing propagates out to trigger `ATOMIC_REQUESTS`. `raise_over_limit_graphql_error` calls `set_rollback()` itself. That rolls back the **whole request**, so a multi-root-field mutation where a later field hits the limit returns 200 with the earlier field's data populated *and* its write rolled back. Documented at the call site, deliberately not guarded.

## Verification

- Full suite **4094 passed** (phase-5 base 4054; +40).
- `mypy` **305 errors / 57 files** — at baseline, **zero `type: ignore` / `noqa` added**.
- `ruff check` + `format --check` clean; `makemigrations --check` clean (no model changes); `schema.yml` verified in sync by regenerating and confirming an empty diff.
- Each guard confirmed by deleting it and watching specific tests fail, then restoring.
- The concurrency test races genuinely — real threads, separate connections, a barrier — with a `lock=False` negative control that demonstrably overshoots.

## Carried forward

- The unlimited-plan query count is now pinned to an exact number rather than a loose budget, so any future regression on the hot invite path fails loudly.
- Everything carried forward from phase 5 (availability-counter residual, add-on expiry, nullable `SystemUser.organization`) is unaffected by this phase.

### DIFF
--- a/accounts/account_adapters.py
+++ b/accounts/account_adapters.py
@@ -23,6 +23,7 @@ from legal.services import ConsentService
 from organizations.exceptions import UserAlreadyHasMembershipError
 from organizations.models import get_active_organization_membership
 from organizations.services import OrganizationService
+from payments.exceptions import OverLimitError
 from users.models import Profile, User
 
 
@@ -149,6 +150,14 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
         mirroring AccountAdapter) and calls provision_tenant_for_user with no
         organisation_name — social signups never carry one.  No-ops silently on
         UserAlreadyHasMembershipError (re-login or race).
+
+        ``OverLimitError`` (the inviting org is at its seat limit) is caught the
+        same way: this runs under allauth.headless views, not DRF, so
+        ``vinta_exception_handler`` never sees it and an uncaught raise here would
+        surface as a bare 500 mid-signup with the social account otherwise fully
+        created and committed. Falling through leaves the user membership-less
+        (gated) — the same outcome as "no pending invitation" — rather than
+        wedging the signup.
         """
         if not user.email:
             logger.warning(
@@ -163,6 +172,15 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
             # Re-login or race: user already has a membership — no-op.
             logger.debug(
                 "Social user %s already has a membership; skipping re-provisioning.",
+                user.pk,
+            )
+            return
+        except OverLimitError:
+            # Inviting org is at its seat limit: leave the user membership-less
+            # (gated) instead of raising a 500 out of a headless view.
+            logger.info(
+                "Social user %s could not auto-join their inviting org: seat limit "
+                "reached. User remains membership-less (gated).",
                 user.pk,
             )
             return
@@ -341,6 +359,15 @@ class AccountAdapter(DefaultAccountAdapter):
         DI container global at the call site.
 
         Idempotent: swallows UserAlreadyHasMembershipError so re-confirmation is a no-op.
+
+        Also swallows ``OverLimitError`` (the inviting org is at its seat limit): this
+        adapter runs under allauth.headless views, not DRF, so
+        ``vinta_exception_handler`` never sees it. Left uncaught, it would surface as a
+        bare 500 *after* the email address is already marked verified — and under
+        ``ATOMIC_REQUESTS`` the whole request (including that verification) rolls back,
+        wedging the user with an unverified email and no way to make a retry succeed.
+        Falling through leaves the user membership-less (gated), matching the existing
+        no-pending-invitation and no-org-name branch below.
         """
         confirmed = super().confirm_email(request, email_address)
         if not confirmed:
@@ -368,6 +395,15 @@ class AccountAdapter(DefaultAccountAdapter):
             # Idempotent: re-confirmation or race — user already has a membership.
             logger.debug(
                 "User %s already has a membership; skipping re-provisioning.",
+                user.pk,
+            )
+            return confirmed
+        except OverLimitError:
+            # Inviting org is at its seat limit: leave the user membership-less
+            # (gated) instead of raising a 500 out of a headless view.
+            logger.info(
+                "User %s could not auto-join their inviting org: seat limit "
+                "reached. User remains membership-less (gated).",
                 user.pk,
             )
             return confirmed
--- a/organizations/services.py
+++ b/organizations/services.py
@@ -309,22 +309,27 @@ class OrganizationService:
         # handler to roll the request transaction back (REST does; a caller that
         # does not would silently commit a rejected invitation).
         if not bypass_limits:
-            # A resend reuses the still-pending row `get_or_create` below would find
-            # (same email/organization, not yet accepted); excluding it from the count
-            # makes the resend net-zero rather than a false block at the exact ceiling.
-            existing_invitation = OrganizationInvitation.objects.filter(
-                organization=organization,
-                email=email,
-                accepted_at__isnull=True,
-                membership_user_id__isnull=True,
-            ).first()
+
+            def _resolve_existing_invitation_id() -> int | None:
+                # A resend reuses the still-pending row `get_or_create` below would
+                # find (same email/organization, not yet accepted); excluding it from
+                # the count makes the resend net-zero rather than a false block at the
+                # exact ceiling. Resolved lazily -- only called once the ceiling is
+                # known to be finite -- so an `unlimited` organization never pays for
+                # this extra query.
+                existing_invitation = OrganizationInvitation.objects.filter(
+                    organization=organization,
+                    email=email,
+                    accepted_at__isnull=True,
+                    membership_user_id__isnull=True,
+                ).first()
+                return existing_invitation.pk if existing_invitation is not None else None
+
             result = self.entitlement_service.check_limit(
                 organization,
                 LimitedResource.ORGANIZATION_MEMBERS,
                 lock=True,
-                exclude_invitation_id=(
-                    existing_invitation.pk if existing_invitation is not None else None
-                ),
+                exclude_invitation_id_resolver=_resolve_existing_invitation_id,
             )
             if not result.allowed:
                 raise OverLimitError.from_check_result(result)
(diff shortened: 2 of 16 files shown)


accounts/account_adapters.py
```diff
@@ -166,6 +175,15 @@ def _provision_org_membership(self, user: User) -> None:
                 user.pk,
             )
             return
+        except OverLimitError:
```
hugobessa, line 178: **BLOCKER 1.** This phase guarded `provision_tenant_for_user`, but this adapter runs under **allauth headless views, which are not DRF** — so `common/exception_handlers` never runs and an uncaught `OverLimitError` becomes a Django 500.

Note `OverLimitError` is deliberately *not* a `ValueError` (phase 5 removed that lineage so the structured error contract could not be flattened), which is exactly why no existing handler caught it.

Social signup was the worse of the two paths: the 500 rolls back the whole request, so the user row itself disappeared and Google signup just failed.

Falling through to `membership = None` is not a fallback invented here — it is the state these adapters already produce for a user who signs up with no invitation.

accounts/account_adapters.py
```diff
@@ -371,6 +398,15 @@ def confirm_email(self, request, email_address: EmailAddress) -> bool:
                 user.pk,
             )
             return confirmed
+        except OverLimitError:
```
hugobessa, line 401: Same fix on the email/password path, where the wedge was more insidious.

`confirm_email` marks the address verified and *then* provisions. A 500 here rolled back the verification along with everything else, so the user's email stayed unverified and **every retry failed identically** — an account permanently stuck with no explanation.

The test asserts the invitation remains pending, which is what makes the user recoverable once a seat frees. Without that assertion the fix would look correct while leaving people stranded.

organizations/services.py
```diff
@@ -282,7 +288,52 @@ def invite_user_to_organization(
...
+                    accepted_at__isnull=True,
+                    membership_user_id__isnull=True,
+                ).first()
+                return existing_invitation.pk if existing_invitation is not None else None
+
+            result = self.entitlement_service.check_limit(
+                organization,
+                LimitedResource.ORGANIZATION_MEMBERS,
+                lock=True,
+                exclude_invitation_id_resolver=_resolve_existing_invitation_id,
```
hugobessa, line 332: **BLOCKER 2, and the reason this is a resolver rather than a plain id.**

Resending an invitation creates nothing and consumes no seat, but the guard counted the very invitation being resent — so an org at its exact ceiling could never resend, in precisely the state where resending matters most (seats just filled, one invite never arrived). That is a false block, which the spec's goal 6 forbids.

The exclusion has to match *exactly* the invitation `get_or_create` will reuse — same organization, same email, same pending predicate. Excluding a row that is not the reused one would make the ceiling one seat too generous, turning a false-block fix into an unmetered path. Review verified the two predicates are identical.

It is lazy because it would otherwise run on every invite including on `unlimited` plans, where phase 5 deliberately skips usage counting entirely. The resolver fires only once a finite ceiling is known to apply.

organizations/services.py
```diff
@@ -574,3 +691,44 @@ def revoke_invitation(self, invitation_id: str) -> None:
                 }
             },
         )
+
+    @transaction.atomic()
+    def reactivate_membership(
```
hugobessa, line 696: `reactivate` started life as a guard inside the viewset. That contradicted the plan's **Enforcement layer** decision ("Service layer, not viewsets") and left it as the one guarded operation with no `bypass_limits` — so a management command reactivating a membership bypassed the ceiling with no opt-in and no way to ask for one.

Reactivation matters because it raises the seat count *without being a create*, which is exactly the shape an enforcement sweep tends to miss.

The guard fires only when the member is currently inactive. That is not a hole: an already-active member is already consuming a seat, so the skip is a genuine no-op and keeps the endpoint idempotent at the ceiling.

public_api/extensions.py
```diff
@@ -8,10 +9,58 @@
...
+    just the current field. GraphQL executes a mutation document's root-level
+    fields serially in one transaction, so a document with more than one root
+    mutation field — e.g. ``mutation { a: createCalendar(...) b:
+    createInvitation(...) }`` where ``b`` hits this — rolls back ``a``'s write too,
+    even though the response still reports 200 with ``data.a`` populated: the
+    client is told ``a`` succeeded when its write did not survive. This is
+    documented rather than guarded against — rejecting multi-root-field
+    documents outright is out of scope here.
+    """
+    set_rollback()
```
hugobessa, line 62: **Why a GraphQL error renderer touches transactions at all.**

graphql-core catches every resolver exception internally and always returns HTTP 200. So unlike a DRF view, nothing propagates out to trigger the `ATOMIC_REQUESTS` rollback — a resolver that wrote before hitting its limit would silently commit. This is the GraphQL analogue of phase 5's own BLOCKER 1 on the REST handler, and it is proven load-bearing by a test that drives a real GraphQL request.

⚠️ `set_rollback()` marks the **whole request's** transaction. GraphQL executes root mutation fields serially, so `mutation { a: createCalendar(...) b: createInvitation(...) }` where `b` hits the limit returns 200 with `data.a` populated *and* `a`'s write rolled back — the client is told `a` succeeded when it did not. Documented rather than guarded; rejecting multi-root-field documents is a bigger decision than this phase should make.

The function raises rather than returns, and is named for it — an earlier version returned the error, so a caller who forgot `raise` would have rolled back the transaction and returned success.

payments/services/entitlement_service.py
```diff
@@ -360,6 +361,7 @@ def check_limit(
         delta: int = 1,
         lock: bool = False,
         exclude_invitation_id: int | None = None,
+        exclude_invitation_id_resolver: Callable[[], int | None] | None = None,
```
hugobessa, line 364: The lazy counterpart to `exclude_invitation_id`, added so the resend fix does not cost a query on the `unlimited` path.

Phase 5 made the unlimited path skip usage counting entirely, and phase 6a is required to prove an unlimited org sees no change in behavior *or query count*. An eager lookup in the caller would have quietly spent that budget on every invite. The resolver is invoked only after the ceiling is known to be finite.

Both forms remain rejected for any `resource_key` other than `organization_members`, where no counter reads them.

PR #196, opened 2026-07-20
feat(calendar_integration): enforce pre-paid limits on calendars, groups, bundles, and availability (billing plans and limits, phase 6b)

hugobessa:
Phase 6b of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Targets `main`, which now has phases [1](https://github.com/vintasoftware/vinta-schedule-api/pull/189)–[6a](https://github.com/vintasoftware/vinta-schedule-api/pull/195).

Developed stacked on 6a; #195 merged mid-review, so this targets `main`. The content diff between them is empty, so no rebase was needed.

## The plan predicted where this would go wrong, and it was right

The plan assigned 6b a Tier-4 reviewer for one stated reason: the bulk sync writers in `calendar_sync_service.py` are *"the single most likely place for an unmetered path to survive, and objective 1 depends on them."*

That is exactly where the worst defect was. Three review rounds; two BLOCKERs, one in each direction.

## BLOCKER 1 — the sync writer let unlimited resource calendars through

The import decided "have we already imported this?" by querying `Calendar` on `(organization, external_id)`. The counter it guards, `_count_resource_calendars`, counts only `calendar_type=RESOURCE`. **Those two predicates disagreed in the permissive direction**, and the write loop's `update_or_create` puts `calendar_type: RESOURCE` in `defaults` — so it *promotes* a matched non-RESOURCE row into the counted set.

An existing PERSONAL calendar on the same `external_id` was therefore judged "already imported", consumed **zero** headroom, and was silently retyped to RESOURCE. Usage rises, headroom does not fall. Repeat per row: unbounded.

**For Microsoft this is a total bypass, not an edge case.** `get_account_calendars` and `get_calendar_resources` both call `client.list_calendars()` and both emit `external_id=calendar.id` — one ID space. Import the account's calendars (correctly unguarded, as PERSONAL), then run the rooms import: every calendar takes the bypass. The only reason it is not live today is that the sole current caller happens to resolve a Google service account. One caller away from the limit meaning nothing.

This is the fourth defect in this plan caused by **two predicates that must agree and were written separately** — after Phase 4's two billing-root definitions, Phase 5's delete-plus-fail-open, and Phase 6a's seat counting on the accept path.

So the fix does not just correct the predicate. It **derives** it: the split now uses `CalendarQuerySet.not_newly_counted_as_type`, defined as the complement of *newly entering* `live_of_type`, sitting directly beside the counter's own predicate — and a property test asserts the two agree across every `(calendar_type, visibility)` combination. Drift is now a test failure rather than a silent bypass.

## BLOCKER 2 — a net-zero availability batch was refused

`batch_modify_available_times` computed `delta` from create operations and ignored deletes, so `[delete one window, create one window]` was refused at the ceiling despite changing usage by zero. An org at its limit could never edit availability by replace-semantics. Same defect class as Phase 6a's resend-at-the-ceiling; the spec's goal 6 forbids it.

The credit is computed with **the counter's own predicate**, not a raw delete count — crediting raw deletes would let `[delete a recurrence-exception row, create]` net to zero and push real usage past the ceiling, which is BLOCKER 1's shape again in miniature.

## What changed

- Guards on `create_resource_calendar`, `create_group`, `create_bundle_calendar`, `create_available_time`, `bulk_create_availability_windows`, and `batch_modify_available_times`, all inside their existing transactions, all with `bypass_limits`.
- Bulk sync: headroom checked **before** the write loop; imports up to the ceiling; records a `PARTIAL` status rather than failing the sync, per the spec's preference for partial import over unmetered creation. Re-runs do not double-count, because headroom is computed from newly-counted rows only.
- `EntitlementService.lock_billing_root()`, extracted so reads that inform a limit decision happen inside the lock.
- GraphQL paths render `OverLimitError` through the shared contract. REST needed no changes — `OverLimitError` is deliberately not a `ValueError`, so it already passes through the existing `except ValueError` wrappers untouched.

**Three methods the plan named are deliberately unguarded.** `create_virtual_calendar`, `create_application_calendar`, and `create_calendar` produce VIRTUAL / INTERNAL / PERSONAL calendars, while `resource_calendars` counts only RESOURCE. Guarding them would consume headroom they never contribute to — a false block. Verified against the counter and the hardcoded types; documented in-code on each method.

**One guard the plan did not name was added.** `batch_modify_available_times` is reachable from the `batchUpdateAvailabilityWindows` mutation, and a batch of N creates would otherwise have bypassed the single-window guard entirely.

## Known limitations, stated rather than buried

- **The root `Subscription` lock is held across the whole import loop.** For a reseller root that serializes every guarded creation across the subtree for the duration. Chunking (re-lock and re-check per chunk) would preserve the invariant and bound the hold; recorded as a follow-up rather than done here.
- **The availability-window counter over-counts** when the first occurrence of a recurring series is edited *or cancelled* — once per operation, unbounded. Carried forward from Phase 5; needs a schema change, not a filter.
- **Pre-existing bug found in passing:** `CalendarQuerySet.update()` reads `self._meta` where it means `self.model._meta`, so any runtime `.update()` on a Calendar queryset raises `AttributeError`. Latent — no runtime caller today. Recorded, deliberately not fixed in a billing phase.

## Verification

- Full suite **4148 passed** (6a base 4094; +54).
- `mypy` **305 errors / 57 files** at baseline; **zero `type: ignore` added**. The only `noqa` are `S106` on test-only dummy credentials, matching five existing precedents.
- `ruff` clean; `makemigrations --check` clean; migration `0041` (choices-only, for `PARTIAL`) applies, reverses, re-applies.
- Both BLOCKER fixes confirmed by reverting them and watching specific tests fail — the promotion tests, and a real threaded overshoot of 4 rows against a ceiling of 3.
- Concurrency tests use real threads on separate connections with a barrier, each with a negative control that demonstrably overshoots without the lock.


calendar_integration/services/calendar_sync_service.py
```diff
@@ -231,26 +256,170 @@ def import_organization_calendar_resources(
...
+        # ``update_or_create`` semantics (later duplicates would just re-write the same
+        # row anyway).
+        resources = list({resource.external_id: resource for resource in resources}.values())
+
+        # Take the guard lock before the split read below -- see the docstring.
+        entitlement_service.lock_billing_root(organization)
+
+        free_ids = Calendar.objects.filter_by_organization(
+            organization.id
+        ).external_ids_not_newly_counted_as_type(
```
hugobessa, line 348: **BLOCKER 1 — the unmetered path the plan predicted would hide in this file.**

The old split asked "does a row already exist for this org + external_id?". The counter asks "how many RESOURCE calendars are live?". Those disagree, and the `update_or_create` below forces `calendar_type=RESOURCE` — so a matched PERSONAL row was judged already-imported, consumed no headroom, and was **promoted** into the counted set. Usage rises, headroom does not fall, unbounded.

For Microsoft it was a total bypass: `get_account_calendars` and `get_calendar_resources` both call `client.list_calendars()` and emit the same `external_id`, so every calendar took it. Not live only because the current sole caller resolves a Google service account.

The correct question is **"will this write increase `_count_resource_calendars`?"** — which is why the predicate is now derived from the counter rather than written next to it.

calendar_integration/querysets.py
```diff
@@ -184,6 +184,53 @@ def live_of_type(self, calendar_type: str) -> "CalendarQuerySet":
         """
         return self.filter(calendar_type=calendar_type).exclude_inactive()
 
+    def not_newly_counted_as_type(self, calendar_type: str) -> "CalendarQuerySet":
```
hugobessa, line 187: This predicate exists so the enforcement side and the counting side **cannot drift apart** — that drift was BLOCKER 1, and it is the fourth time this plan has produced a defect from two predicates that had to agree and were maintained separately (Phase 4's billing roots, Phase 5's delete-plus-fail-open, Phase 6a's seat counting).

It lives beside `live_of_type` deliberately: it is the complement of *newly entering* `live_of_type`, so the two are read and changed together.

Note the subtlety worth preserving — a **soft-deleted** row retyped by the upsert stays `INACTIVE`, never enters `live_of_type`, and so must stay free. Charging it would be a false block on a write that moves usage by zero. That is why this is not simply `calendar_type=RESOURCE`.

`TestSplitPredicateMatchesTheUsageCounter` asserts the equivalence across every `(calendar_type, visibility)` combination, so future drift fails a test instead of silently uncapping a limit.

calendar_integration/services/availability_service.py
```diff
@@ -522,6 +569,47 @@ def batch_modify_available_times(
...
+        entitlement_service = self._context.entitlement_service
+        # A credit is about to be granted (create_count and delete_ids), which means the
+        # delete-credit read below can race a concurrent batch computing the same credit
+        # from the same pre-write snapshot. Lock *before* that read -- not after, guarded
+        # by ``delta`` -- because ``delta == 0`` (the net-zero replace case) is exactly the
+        # case that must not skip the lock: two concurrent net-zero batches on the same
+        # delete id would otherwise both see it as live, both credit it, and both skip
+        # ``check_limit`` entirely. See ``EntitlementService.lock_billing_root``.
+        if not bypass_limits and entitlement_service is not None and create_count and delete_ids:
+            entitlement_service.lock_billing_root(context.organization)
```
hugobessa, line 587: **Two bugs met here, and the second is the subtle one.**

First, the delta counted creates and ignored deletes, so `[delete one, create one]` was refused at the ceiling despite net-zero usage — a false block, same class as Phase 6a's resend, forbidden by the spec's goal 6.

Second, crediting a *raw* delete count reopens an unmetered path: the counter only counts user-authored rows, so `[delete a recurrence-exception row, create]` would net to zero, skip the check, and push real usage past the ceiling. The credit is therefore computed with the counter's own predicate — BLOCKER 1's lesson applied in miniature.

The lock is taken **before** the credit is read and regardless of whether `delta` ends up zero. Without that, two concurrent `[delete X, create]` batches both compute `delta=0`, both skip `check_limit` entirely, and the loser's delete silently affects zero rows — both creates land, one over the ceiling. Proven with a real threaded test plus a negative control.

payments/services/entitlement_service.py
```diff
@@ -354,6 +354,32 @@ def _count_usage(
...
+    @staticmethod
+    def _lock_billing_root_row(root: Organization) -> None:
+        """Take ``SELECT ... FOR UPDATE`` on ``root``'s ``Subscription`` row.
+
+        Discards the returned row: the point is the row lock, and every subsequent
+        read in the caller's transaction goes through the same connection.
+        """
+        Subscription.objects.select_for_update().filter(organization=root).first()
+
+    def lock_billing_root(self, organization: Organization) -> None:
```
hugobessa, line 366: Extracted so a caller can hold the billing-root lock across a **read that informs a limit decision**, not just across `check_limit` itself.

The sync importer needs it because it reads which calendars are already imported and then decides how much headroom to consume; reading that outside the lock lets a concurrent import compute stale headroom. `check_limit` now shares the same `_lock_billing_root_row` implementation, so there is exactly one place that decides which row is locked — still the **root** subscription, and still fail-open when no subscription exists.

calendar_integration/constants.py
```diff
@@ -38,6 +38,12 @@ class CalendarSyncTriggerSource(TextChoices):
 
 class CalendarOrganizationResourceImportStatus(TextChoices):
     SUCCESS = "success", "Success"
+    # The import ran without error but did not import everything it discovered --
+    # currently only because the organization ran out of `resource_calendars`
+    # headroom mid-import. A distinct terminal status rather than SUCCESS plus an
+    # advisory string on `error_message`: a consumer that has to string-match an
+    # error column to tell a clean import from a truncated one has no contract.
+    PARTIAL = "partial", "Partial"
```
hugobessa, line 46: A partial import previously reported `SUCCESS` with an advisory string in `error_message`. No consumer could distinguish a clean import from one that silently dropped rooms without string-matching an error column, which is not a contract.

`PARTIAL` is choices-only (migration `0041`, no DDL, reverses cleanly) and no serializer or GraphQL type exposes this status, so nothing downstream breaks on the new member.

calendar_integration/services/calendar_sync_service.py
```diff
@@ -89,10 +90,26 @@
...
+    from organizations.models import Organization
 
 
 logger = logging.getLogger(__name__)
 
+#: How many skipped ``external_id``s a partial-import warning names before eliding.
+#: The warning goes into a log line *and* a ``TextField`` on the import workflow
+#: state; a 5000-room organization would otherwise write a multi-hundred-KB string
+#: to the database on every capped import.
+MAX_SKIPPED_EXTERNAL_IDS_IN_WARNING = 20
```
hugobessa, line 102: The skipped-id list is capped before it reaches a log line and a DB column. A 5000-room organization over its ceiling would otherwise write a multi-hundred-KB string into `error_message` on every sync.

PR #197, opened 2026-07-20
feat(payments): guard webhook subscriptions and API system users, add entitlement gates (billing plans and limits, phase 6c)

hugobessa:
Phase 6c of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). **Stacked on [phase 6b](https://github.com/vintasoftware/vinta-schedule-api/pull/196)** — review that one first.

**This phase closes spec objective 1 for pre-paid resources.** Every `kind=prepaid` member of `LimitedResource` now has a guarded creation path with a test proving it blocks. It also adds the three boolean entitlement gates.

## The acceptance condition, machine-checked

`payments/tests/test_prepaid_resource_coverage.py` derives the prepaid set from the seeded plan's own `PlanLimit.kind` rather than a hand-typed list, and drives each probe through the **real** guarded service method:

| resource | guarded path | phase |
|---|---|---|
| `organization_members` | `OrganizationService.invite_user_to_organization` | 6a |
| `resource_calendars` | `CalendarService.create_resource_calendar` | 6b |
| `calendar_groups` | `CalendarGroupService.create_group` | 6b |
| `bundle_calendars` | `CalendarService.create_bundle_calendar` | 6b |
| `availability_windows` | `CalendarService.create_available_time` | 6b |
| `webhook_subscriptions` | `WebhookService.create_configuration` | **6c** |
| `public_api_system_users` | `PublicAPIAuthService.create_system_user` | **6c** |

`event_occurrences` is asserted as the one deliberate exclusion (`kind=postpaid`, owned by Phase 8). The registry check is bidirectional, so both an unguarded new prepaid resource and a stale probe fail loudly.

### A deliberate design decision was reversed to make tests pass — and reverted

`has_entitlement` fails **closed** by Phase 5's reviewed design: absence encodes revocation, because `_sync_entitlements` *deletes* rows for entitlements a plan lacks.

The first implementation changed the missing-subscription branch to fail **open**, after 106 test failures caused by calendar-integration fixtures that build organizations with no `Subscription`. The justification offered was symmetry with `get_effective_limit`, which fails open on the same condition.

**That symmetry is invalid.** For a numeric ceiling, "we don't know" → NULL → *unlimited*, which coincides with what the rollout actually seeds. For a **boolean** gate, "we don't know" → `True` grants **paid** features — strictly more than the `free` plan grants. And the rollout-safety argument was already satisfied: the backfill migration puts every billing root on `unlimited`, whose entitlements are all enabled. The fail-open branch contributed nothing to rollout safety and fired only when billing state was *already* broken — handing the most access to exactly those organizations. Ops deleting a `Subscription` to re-provision an org would, in that window, grant a `free` org the entire partner API.

Reverted. `payments/services/entitlement_service.py` has **zero net diff** against phase 6b. The fixtures were fixed instead — an autouse fixture provisions a subscription for test organizations, with 20 modules opting out via `@pytest.mark.no_auto_subscription` because they own their own rows or deliberately exercise the plan-less path.

### The entitlement gate was bypassable

The plan chose `authenticate` as the chokepoint "both connection paths flow through". It is not. `_get_write_adapter_for_calendar` builds a provider adapter for whoever *owns* the calendar. An org with Google enabled and Microsoft disabled could authenticate via Google, pass the gate, then write to a Microsoft calendar owned by another user — ungated Microsoft traffic.

### A previously-working admin path started 500ing

`create_system_user(organization=None)` is explicitly supported (`SystemUser.organization` is nullable, and `save_model` branches on the org-less case). The guard passed `None` into `resolve_billing_root`, producing `AttributeError` and an admin 500.

## Two `@inject` landmines found, neither billing-related

Both were invisible to the test suite, and one is a live production bug:

1. **Silent no-op under `from __future__ import annotations`.** dependency_injector cannot read `Provide[...]` out of a stringified annotation, so it emits a warning and returns the function **unpatched**. The branding guard passed every test while doing nothing.
2. **Admin autodiscovery beats DI wiring.** `SystemUserAdmin` was constructed by `@admin.register` *before* `DICoreConfig.ready()` calls `container.wire()`, so its injected service was permanently `None` — **every** admin-created system user raised `ValueError`, entirely independent of billing.

The pattern exists elsewhere in the codebase. Worth a deliberate sweep rather than fixing it where it happens to surface.

## Reviewer-facing notes

- **`_get_write_adapter_for_calendar` now raises** rather than returning `None`. Callers were traced individually: GraphQL is inside `try/except OverLimitError`, DRF maps to 402 via the global handler, and the five `if write_adapter := ...` sites run under `ATOMIC_REQUESTS` so the local write rolls back rather than diverging from the remote. The **calendar webhook service was the one caller** where skipping the remote work was intended — it now catches `OverLimitError` explicitly, because otherwise the provider gets a 500 and retries until the channel expires while the event row is never recorded.
- **`resolve_branding` was split.** The gate reached further than branding: it also backs `validate_return_url`, so a reseller downgrading off `white_label_branding` would have broken the OAuth return flow for every downstream tenant. `resolve_branding` is now ungated; `resolve_branding_for_display` carries the gate.
- **`partner_api` fails closed when DI is broken**, deliberately — a wiring failure should deny, not grant every org unrestricted API access.

## Verification

Re-run independently by the orchestrator in the container, not relayed:

- Full suite **4200 passed** (6b base 4148; +52).
- `mypy` **305 errors / 57 files** — at baseline. The first implementation regressed this to 309/60 and reported "net new: zero"; its `git stash` verification had left the new untracked test files on disk, so it measured the phase against itself. Verified here against a clean checkout of the base branch.
- `ruff` clean; `makemigrations --check` clean; `schema.yml` unchanged.
- Each guard confirmed by deleting it and watching specific tests fail, then restoring.
- 5 × `noqa: S106`, all dummy credentials in tests.

## Carried forward

- `ADVANCED_SCHEDULING` is a declared `Entitlement` with **no gate anywhere**. Out of scope here (the phase body names four gates); recorded so it is not mistaken for covered.
- `SystemUserAdmin`'s changelist raises `ImproperlyConfigured: QuerySet must be filtered by 'organization'` — a real pre-existing admin defect, unrelated to this phase.
- The `@inject` sweep described above.
- Everything from phases 5/6b (availability-counter residual, add-on expiry, nullable `SystemUser.organization`, `CalendarQuerySet.update()`'s `self._meta` bug, sync-lock chunking).

### DIFF
--- a/calendar_integration/services/calendar_service.py
+++ b/calendar_integration/services/calendar_service.py
@@ -168,6 +168,21 @@ _PROVIDER_ENTITLEMENTS: dict[str, str] = {
 }
 
 
+def _provider_for_account(account: object) -> str | None:
+    """The ``CalendarProvider`` an account authenticates against, or ``None`` when it
+    cannot be determined without resolving an adapter.
+
+    ``None`` is returned for a bare ``User``: which of that user's social accounts is
+    used is ``get_calendar_adapter_for_account``'s decision, so the caller has to gate
+    after resolution instead. Every other case is readable off the object.
+    """
+    if isinstance(account, GoogleCalendarServiceAccount):
+        return CalendarProvider.GOOGLE
+    if isinstance(account, User):
+        return None
+    return getattr(account, "provider", None)
+
+
 def _resolve_owner_membership_user_id(user: User, organization: Organization) -> int | None:
     """Resolve the membership-scoped owner id for a CalendarOwnership write.
 
@@ -224,6 +239,9 @@ class CalendarService(BaseCalendarService):
         self.external_event_change_request_service = external_event_change_request_service
         self.booking_policy_service = booking_policy_service
         self.entitlement_service = entitlement_service
+        # Set by authenticate(bypass_limits=True); disables every provider entitlement
+        # guard on this instance, not just the authenticate-time one.
+        self._bypass_entitlement_limits = False
         # Per-instance calendar lookup cache: keyed on (organization_id, id_or_external_id).
         # This replaces the @lru_cache approach which was keyed only on id/external_id and
         # could return a cached Calendar from a different organization when the service
@@ -406,6 +424,23 @@ class CalendarService(BaseCalendarService):
             }
         ), token.account
 
+    def _assert_provider_entitlement(self, provider: str | None) -> None:
+        """Raise ``OverLimitError`` if ``provider`` is gated and the organization is not
+        entitled to it.
+
+        Providers with no ``_PROVIDER_ENTITLEMENTS`` entry (INTERNAL, APPLE, ICS) are
+        ungated: the spec's closed ``Entitlement`` set only names Google and Microsoft.
+        A no-op when the service has no ``entitlement_service`` injected, or when
+        ``authenticate(bypass_limits=True)`` put this instance in bypass mode.
+        """
+        if self._bypass_entitlement_limits or self.entitlement_service is None:
+            return
+        entitlement_key = _PROVIDER_ENTITLEMENTS.get(provider) if provider else None
+        if entitlement_key is None or self.organization is None:
+            return
+        if not self.entitlement_service.has_entitlement(self.organization, entitlement_key):
+            raise OverLimitError.from_missing_entitlement(entitlement_key)
+
     def authenticate(
         self,
         account: "User | SocialAccount | GoogleCalendarServiceAccount",
@@ -420,14 +455,16 @@ class CalendarService(BaseCalendarService):
             ``CalendarOwnership``).
         :param organization: Calendar organization instance.
         :param bypass_limits: When True, skips the ``external_calendar_google`` /
-            ``external_calendar_microsoft`` entitlement guard below. Only management
-            commands and one-off repair scripts should pass this -- never a
-            request-handling path.
+            ``external_calendar_microsoft`` entitlement guard below, and every guard on
+            this service instance for the rest of its life (writes resolved through
+            ``_get_write_adapter_for_calendar`` included). Only management commands and
+            one-off repair scripts should pass this -- never a request-handling path.
         :raises OverLimitError: When the resolved account's provider is Google or
-            Microsoft and the organization lacks the matching entitlement. This is
-            the common chokepoint both connection paths flow through -- every caller
-            that authenticates a Google or Microsoft account, including calendar
-            sync, routes through here.
+            Microsoft and the organization lacks the matching entitlement. Every caller
+            that authenticates a Google or Microsoft account, including calendar sync,
+            routes through here -- but this is **not** the only enforcement point:
+            ``_get_write_adapter_for_calendar`` gates on the *calendar's* provider,
+            which can differ from the authenticated account's. See its docstring.
         """
         if isinstance(account, User):
             self.user_or_token = account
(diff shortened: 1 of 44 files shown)


calendar_integration/services/calendar_service.py
```diff
@@ -397,10 +424,28 @@ def get_calendar_adapter_for_account(
             }
         ), token.account
 
+    def _assert_provider_entitlement(self, provider: str | None) -> None:
```
hugobessa, line 427: **The plan's chokepoint assumption was wrong, and this is the fix.**

Phase 6c's body names `authenticate` as "the common chokepoint both connection paths flow through". It is not. `_get_write_adapter_for_calendar` builds an adapter for whoever *owns* the calendar, not the actor — so an org with `external_calendar_google=True` and `external_calendar_microsoft=False` could authenticate via Google, pass the gate here, then write to a Microsoft calendar owned by another user and generate ungated Microsoft traffic.

The gate is therefore applied at both places. Note this resolves the provider *before* `get_calendar_adapter_for_account` where it can, because that call refreshes an expired access token on construction — an outbound provider call and a token mutation for a request that is about to be rejected.

calendar_integration/services/calendar_webhook_service.py
```diff
@@ -413,6 +414,20 @@ def process_webhook_notification(
         except (ServiceNotAuthenticatedError, Calendar.DoesNotExist):
             # Calendar not found or not authenticated - we'll still record the webhook event
             pass
+        except OverLimitError as exc:
```
hugobessa, line 417: **The one caller where "skip the remote work and carry on" is the intended behavior.**

`_get_write_adapter_for_calendar` now *raises* rather than returning `None` — deliberately, because callers using `if write_adapter := ...` would otherwise complete the local write while silently skipping the remote one, diverging state on a billing decision.

Every other caller is correctly mapped: GraphQL wraps it, DRF renders 402, and the `calendar_event_service` sites run under `ATOMIC_REQUESTS` so the local write rolls back. This one is different. Without this catch the provider receives an **HTTP 500** and the `CalendarWebhookEvent` row is never recorded, so Google/Microsoft retry against the error until the channel expires — a phase that should degrade one org's sync would instead break its webhook channel.

organizations/models.py
```diff
@@ -528,3 +538,54 @@ def resolve_branding(org: Organization) -> OrganizationBranding | None:
     if branding_root is None:
         return None
     return getattr(branding_root, "branding", None)
+
+
+def resolve_branding_for_display(org: Organization) -> OrganizationBranding | None:
```
hugobessa, line 543: **Why the branding gate needed splitting, not just adding.**

`resolve_branding` also backs `validate_return_url` (`public_api/queries.py`). Gating it wholesale meant a reseller downgrading off `white_label_branding` would return `{allowed: False, sanitized_url: None}` for **every downstream tenant's OAuth interstitial** — an authentication-flow lockout caused by a cosmetic billing gate, and exactly the "no organization is blocked as a consequence of the rollout" rule this plan is bound by.

So `resolve_branding` stays ungated for allowlist checks, and this display-only variant carries the gate. Both presentation callers (`branding_for_tenant`, notification contexts) use this one; `validate_return_url` uses the plain version.

It fails **closed** when the container is unavailable, matching the middleware rather than contradicting it.

public_api/middlewares.py
```diff
@@ -77,22 +85,72 @@ def _get_organization_from_request(self, request: PublicApiHttpRequest) -> Organ
             return None
         return Organization.objects.filter(id=organization_id).first()
 
+    @inject
+    def _has_partner_api_entitlement(
```
hugobessa, line 89: This gate sits in front of **every** GraphQL operation, so a mistake here takes an org's whole partner API down rather than blocking one action. Three things keep that safe:

- It evaluates only after a successful system-user resolution, so anonymous/public operations are untouched and a denial can never be mistaken for an auth failure.
- Every billing root gets `partner_api` from the seeded plans; `free` sets it `False` as a deliberate restriction, never by omission.
- It returns a real HTTP 402 carrying the shared error contract, not a per-resolver `errors` entry — "cannot use the GraphQL API at all" is the intended semantics.

`entitlement_service is None` (DI wiring broken) fails **closed** on purpose: a wiring failure should deny, not silently grant every organization unrestricted API access.

public_api/admin/system_user.py
```diff
@@ -10,10 +10,77 @@
...
+from payments.exceptions import OverLimitError
 from public_api.models import ResourceAccess, SystemUser
 from public_api.services import PublicAPIAuthService
 
 
+if TYPE_CHECKING:
+    from payments.services.entitlement_service import EntitlementService
+
+
+class SystemUserAdminForm(forms.ModelForm):
```
hugobessa, line 23: The limit check lives in the form's `clean()`, not in `save_model`, because catching the error in `save_model` does not actually work: Django proceeds to `save_related`, `log_addition`, and `response_add` regardless, producing a confusing redirect that reports both an error and "added successfully" while creating nothing, plus a junk `LogEntry`.

Moving `@inject` off `__init__` onto `save_model` fixes a **separate, live production bug** this surfaced: `@admin.register` constructs the admin during autodiscovery, *before* `DICoreConfig.ready()` calls `container.wire()`, so the injected service was permanently `None` and **every** admin-created system user raised `ValueError` — nothing to do with billing.

That is the second `@inject` failure mode found in this phase; the other is a silent no-op under `from __future__ import annotations`, where the decorator returns the function unpatched. Both were invisible to the test suite, and the pattern exists elsewhere in the codebase.

conftest.py
```diff
@@ -106,6 +106,117 @@ def mock_external_calendar_clients(monkeypatch):
...
+        )
+    for entitlement_key in Entitlement.values:
+        PlanEntitlement.objects.update_or_create(
+            plan=plan, entitlement_key=entitlement_key, defaults={"is_enabled": True}
+        )
+    return plan
+
+
+@pytest.fixture(autouse=True)
+def provision_default_subscription(request):
```
hugobessa, line 156: **This fixture is the fix for a finding, and it is worth understanding why it exists.**

The first implementation of this phase changed `has_entitlement` to fail **open** on a missing subscription, because 106 tests failed — their fixtures build organizations with no `Subscription`. That reversed a deliberate, previously-reviewed Phase 5 decision, and it granted paid features to exactly the organizations whose billing state was broken.

The fixtures were the defect, not the semantics. This provisions a subscription for test organizations so they reflect Phase 4's production invariant ("every billing root holds exactly one subscription"). Twenty modules opt out via `@pytest.mark.no_auto_subscription` — either they own their own `Subscription` rows (it is a OneToOne) or they deliberately exercise the plan-less/entitlement-less path, which must stay exercisable.

`payments/services/entitlement_service.py` has zero net diff against phase 6b as a result.

PR #198, opened 2026-07-20
feat(payments): meter event occurrences idempotently (billing plans and limits, phase 7)

hugobessa:
Phase 7 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). **Based on `main`** — phases 6b/6c merged mid-run, so this no longer stacks on either.

**This is the spec's highest-severity risk.** Occurrences of a recurring series are computed in Postgres and never stored, so nothing exists to bill until this phase writes it. A double-count here is silent revenue drift or an overcharge, invisible until a customer disputes an invoice — there is no exception, no failing test, just a wrong number on the page they see.

## What this ships

`payments/services/metering_service.py` (`MeteringService`) records every event occurrence, starting in a swept `[window_start, window_end)`, as a `MeteredOccurrence` row — exactly once, ever. `payments/tasks.py` adds a Celery beat entry (`meter_event_occurrences`, every 15 minutes) that fans out one sweep per subscription over a **6-hour** window, deliberately wider than the beat interval so up to 23 consecutive missed runs self-heal on the next successful one.

**The unique constraint is the correctness mechanism, not the code path.** `MeteredOccurrence(organization, event_id, occurrence_start)` plus `bulk_create(..., ignore_conflicts=True)` is what makes re-running a window, or running two overlapping windows, a database-level no-op. That is deliberate: an application-level "have I already seen this?" lookup would race itself, and the failure mode is a silently wrong invoice, not an exception.

`is_within_allowance` and `unit_price` are stamped **at meter time** against the effective limit in force then, so a later plan change or limit override cannot retroactively reprice usage that already happened.

`reconcile_period(subscription, period)` recomputes a closed cycle from the same expansion (`expand_occurrence_identities`) and reports drift both ways — the spec's named mitigation for the residual risk this phase can't eliminate outright.

## Three commits, and the middle one matters most

1. `64ea614` — the model, service, migration, task, and beat entry.
2. `ec758f1` — three review blockers: the meter and the pre-paid usage counter were reading **different definitions of "billing period"** (the counter read `Subscription.current_period_start` directly, which nothing advances until Phase 13's cycle close, while the meter resolved each occurrence's own period — they agreed only by fixture coincidence); a "recurrence slot" abstraction that claimed to survive re-timing was reverted because `calculate_recurring_events` never emits the thing it claimed to key on; and a first pass at diagnosing a bulk-modification over-count, which turned out to be wrong.
3. `a05ddb9` — corrects that diagnosis. See below.

## The upstream recurrence defect found in passing

Not introduced by this phase, not fixed by it, and the most important thing in this PR.

`recurrence_utils`' module docstring promises that splitting a rule returns "new, unsaved instances." It doesn't. `copy.deepcopy` of a **saved** Django model preserves its `pk`, so both rules `RecurrenceRuleSplitter.split_at_date` returns are aliases for the original row. The splitter's *arithmetic* is correct — a truncated parent (`count=None`, `until=<last occurrence before the split>`) and a continuation carrying the remaining count, `1 + 4 = 5` for a `COUNT=5` series split at its second occurrence.

The bug is in persistence. `RecurrenceManager.create_recurring_bulk_modification_generic` truncates the parent first, then calls `continuation_rule.save()` — which still carries the original pk, so it `UPDATE`s the **parent's** rule row and overwrites the `UNTIL` just written. For a `COUNT`-bounded series this leaves the parent at `COUNT=4, until=NULL`, so parent and continuation together yield eight occurrences where five exist. For an **open-ended** series the continuation rule is byte-identical to the original unbounded rule, so the parent never stops — the series duplicates indefinitely, not just across the split window. This is the shape most standing meetings have.

Measured directly in `payments/tests/test_metering_reconciliation.py` (offset modification, 5 real occurrences):

| Scenario | Billed | `reconcile_period` |
|---|---|---|
| Modify, then sweep once (no prior metering) | **8** | `drift == 0`, reports **clean** |
| Sweep, then modify, then sweep | **9** | `drift == 1` (1 `orphaned`) |

**Reconciliation is blind to the first case.** It recomputes from the same calendar, which genuinely contains eight rows, so it has no independent signal to catch a defect upstream of the expansion it trusts. The second case underestimates by 3.

This is a `CalendarEventService.create_recurring_event_bulk_modification` defect, not a metering one — `get_calendar_events_expanded` returns exactly what got billed, so the meter is faithful to what it's shown.

**Where it gates.** Phase 8's post-paid guard is inert on every organization that exists today: `unlimited` carries a NULL `event_occurrences` limit, and a NULL ceiling always returns headroom, so no code path can block on an over-counted number yet — Phase 8 ships as planned. The real gate is the first non-NULL `event_occurrences` limit reaching a live organization: **Phase 14**'s rollout, and **Phase 13**'s overage charge, both must wait for the upstream fix, because that is where the over-count turns into money.

## The bundle-booking decision (binding on Phase 8, not changed here)

One bundle booking bills as **`1 + n_internal_children`** occurrences, not 1. `CalendarBundleService` creates a primary `CalendarEvent` plus, per child calendar, either a full `CalendarEvent` (provider `INTERNAL`) or a `BlockedTime` (every other provider) — and the meter counts `CalendarEvent` rows without a bundle-specific exception, so the same user action costs 1 across five Google calendars but 5 across five internal ones.

Left as-is: an internal child is a real, independently-editable `CalendarEvent` that costs the same to store and expand as any other; a `BlockedTime` is not billable anywhere in this plan, so there is nothing arbitrary about billing what exists. Teaching the meter's one expansion a bundle-specific exception is exactly the "two predicates that must mean the same thing, written separately" shape that has produced a defect in every phase of this plan so far. Nobody is charged for it yet — every organization is `unlimited` through the rollout.

**Revisit trigger.** If product wants one bundle booking to always cost 1 unit, the change is `expand_occurrence_identities` skipping `bundle_primary_event__isnull=False`, made **together with** Phase 8's guard and Phase 12's usage readout — three places that must agree, or the meter and the guard disagree on what a booking costs.

## Verification

- Full suite **4262 passed** (base 4237; +25).
- `mypy` **305 errors / 57 files** — zero new errors by set diff against the 6c baseline.
- `ruff` clean (500 files); `makemigrations --check` clean; migration `0011` reverses.

**A fabricated mechanism was caught before it shipped.** The fixer's first pass claimed `truncate_parent` decrements a `COUNT`-bounded parent's count by one. Reading `RecurrenceRuleSplitter` showed the arithmetic was correct; challenged for persisted state rather than reasoning, the agent replied that the claim was fabricated — it saw `COUNT=4` in the persisted row and wrote a plausible-sounding cause instead of finding the real one. `a05ddb9` is the retraction and the real diagnosis above.

**A `mypy` regression report (315/58) did not reproduce.** Three re-runs, one with `.mypy_cache` removed, plus a direct error-set diff, all returned 305/57 with zero new errors. The 315 run is attributed to container contention under parallel load, which also produced three non-reproducible phantom test failures elsewhere in the same run. Corroborate a gate delta with a second run before acting on it.

## Carried forward

- `test_without_the_lock_the_net_zero_race_overshoots` (`calendar_integration/tests/services/test_availability_limit_concurrency.py`), a Phase 6b test already merged in #196, is **flaky**: a negative control asserting a race reproduces when a lock is removed. Passed 3/3 in isolation, failed once under full parallel load in this run. Inherently timing-dependent by what it's testing; worth making deterministic or marking `flaky` so it doesn't intermittently redden CI on unrelated PRs.
- The re-timing identity-churn defect (`expand_occurrence_identities`'s second bullet): moving an occurrence mints a new `(series root, start time)` identity and bills it again if the original stretch was already metered. Bounded, characterised, surfaced as `orphaned` drift by `reconcile_period`. A Phase 13 precondition alongside the bulk-modification defect above.
- Everything carried from phases 5/6b/6c (availability-counter residual, add-on expiry, nullable `SystemUser.organization`, `CalendarQuerySet.update()`'s `self._meta` bug, sync-lock chunking, the `@inject` landmines, `ADVANCED_SCHEDULING`'s ungated entitlement).


payments/models.py
```diff
@@ -441,3 +441,84 @@ class Meta(BaseModel.Meta):
...
+    unit_price = models.DecimalField(max_digits=10, decimal_places=4)
+
+    objects: ClassVar[MeteredOccurrenceManager] = MeteredOccurrenceManager()
+
+    class Meta(BaseModel.Meta):
+        constraints: ClassVar = [
+            UniqueConstraint(
+                fields=["organization", "event_id", "occurrence_start"],
+                name="uniq_metered_occurrence",
+            )
```
hugobessa, line 514: **The unique constraint is the correctness mechanism, not the code path.**

`(organization, event_id, occurrence_start)` plus `bulk_create(..., ignore_conflicts=True)` is what makes re-running a window, or running two overlapping windows, a no-op at the database level. That is load-bearing: the sweep window (`payments/tasks.py`, `METERING_SWEEP_WINDOW`) deliberately overlaps the previous run so a missed sweep self-heals, and that is only safe because idempotence is enforced below the application rather than remembered by it. An application-level "have I already seen this?" lookup would race itself, and the failure mode is a silently wrong number on an invoice, not an exception.

`event_id` is a soft reference (`BigIntegerField`, not a `ForeignKey`) on purpose — an occurrence is billed at most once *ever*, and that fact has to outlive the event it billed if the event is later deleted.

payments/models.py
```diff
@@ -441,3 +441,84 @@ class Meta(BaseModel.Meta):
...
+        related_name="metered_occurrences",
+    )
+    subscription = models.ForeignKey(
+        Subscription, on_delete=models.CASCADE, related_name="metered_occurrences"
+    )
+    event_id = models.BigIntegerField()
+    occurrence_start = models.DateTimeField()
+    billing_period_start = models.DateTimeField(db_index=True)
+    is_within_allowance = models.BooleanField()
+    unit_price = models.DecimalField(max_digits=10, decimal_places=4)
```
hugobessa, line 505: **Stamped at meter time, not read live.**

`is_within_allowance` and `unit_price` are resolved against the allowance and overage price in force *when the occurrence is recorded*, so a later plan change or limit override cannot retroactively reprice usage that already happened. This is what makes the row a billing record rather than a cache of a live computation — read it back for what the customer was actually charged, not for what the current plan would charge today.

payments/services/subscription_service.py
```diff
@@ -86,6 +87,104 @@ def resolve_billing_root(organization: Organization) -> Organization:
...
+    ``relativedelta`` rather than ``timedelta`` so a monthly cycle anchored on the
+    31st, or one spanning a DST transition, lands on the same wall-clock anchor
+    instead of drifting by a day.
+    """
+    if billing_interval == BillingInterval.ANNUAL:
+        return relativedelta(years=1)
+    return relativedelta(months=1)
+
+
+def resolve_billing_period(
```
hugobessa, line 109: **The single definition of "which cycle does this occurrence belong to" — and the bug this phase found in its own first draft.**

Before `ec758f1`, `MeteringService` stamped `billing_period_start` by resolving each occurrence's own start time, while the pre-paid `event_occurrences` usage counter read `Subscription.current_period_start` directly. Those two only agreed while the stored column happened to still contain "now" — and nothing advances it until Phase 13's cycle close, so once a subscription's stored period elapsed, the meter wrote to one period and the counter read from another, permanently returning zero. The old tests passed only because a fixture pinned the stored period to the test window.

Both now derive from this one function — `MeteringService` via `resolve_billing_period_start(subscription, occurrence_start)`, the counter via `current_billing_period_start(subscription)` (same function, `moment=now()`). Three hand-written date comparisons that are supposed to agree is exactly how a charge lands on the wrong invoice.

calendar_integration/querysets.py
```diff
@@ -605,6 +605,115 @@ def annotate_recurring_occurrences_with_bulk_modifications_on_date_range(
             )
         )
 
+    def occurrence_bearing_masters_in_range(
```
hugobessa, line 608: **This is the one definition of "what occurrences exist in this range," and the docstring below is where the upstream recurrence defect is recorded for a reader who would otherwise assume the parent and continuation of a bulk-modification split can't overlap.**

They can, and after `create_recurring_event_bulk_modification` they usually do: `copy.deepcopy` of a saved Django model preserves its pk, so the continuation rule `RecurrenceRuleSplitter` returns is an alias of the original row, not a new one. Saving it after the parent was truncated silently overwrites the parent's truncation. For a `COUNT=5` series split at occurrence 2, that leaves parent + continuation yielding 8 occurrences where 5 exist; for an open-ended series the parent never stops at all.

Not fixed here — this phase found it while defining "what counts as an occurrence" and needed to record it precisely so `MeteringService` isn't blamed for faithfully billing what the calendar genuinely (if wrongly) contains. See the PR description for the full mechanism and the gating consequence for Phases 13/14.

payments/services/metering_service.py
```diff
@@ -0,0 +1,584 @@
...
+            for organization_id, event_id, occurrence_start in existing.values_list(
+                "organization_id", "event_id", "occurrence_start"
+            )
+        }
+
+    # ------------------------------------------------------------------
+    # Expansion
+    # ------------------------------------------------------------------
+
+    def expand_occurrence_identities(
```
hugobessa, line 319: **Identity is `(series root pk, occurrence start)` — and only half of that is durable.**

The series-root half holds: `_resolve_series_root_ids` walks a bulk-modification chain back to the original master, so a split doesn't re-bill the tail under the continuation's new pk. The start-time half does not — re-timing an occurrence mints a new identity and bills it again if the original stretch was already metered. An earlier revision of this phase (reverted in `ec758f1`) tried to thread a "recurrence slot" through the expansion that would have survived a move; it was based on a false premise, since `calculate_recurring_events` emits a modified exception at its own moved `start_time`, never at the original slot. No durable identity is available anywhere in the current expansion, so this is recorded as a deferred, bounded defect (`reconcile_period` reports it as `orphaned`) rather than worked around here.

This function has no special case for `bundle_primary_event` — a bundle's internal-provider children are real masters like any other, so one bundle booking currently costs `1 + n_internal_children`, not 1. See the PR description's "bundle-booking decision" for why, and its three-place revisit trigger if that changes.

payments/tasks.py
```diff
@@ -0,0 +1,102 @@
...
+#: **Operator action after an outage longer than this.** Self-healing stops at six
+#: hours; beyond that the un-swept stretch is never billed, and nothing raises,
+#: because the next sweep only ever looks six hours back. There is no backfill
+#: management command. Re-meter the gap by calling
+#: ``MeteringService.meter_occurrences_for_period(subscription, gap_start, gap_end)``
+#: for each subscription in ``MeteringService.subscriptions_to_sweep()`` — it is
+#: idempotent, so an over-wide window is safe — then confirm with
+#: ``reconcile_period``, which reports the recovered stretch as ``unmetered``
+#: before the backfill and clean after it.
+METERING_SWEEP_WINDOW = datetime.timedelta(hours=6)
```
hugobessa, line 44: **Why the sweep window (6h) is wider than the beat cadence (15m), and what breaks past it.**

The overlap is what makes a missed run self-heal — up to 23 consecutive missed runs are absorbed by the next successful one, with no operator action and no backfill command, because re-metering an already-swept stretch inserts nothing (the unique constraint on `MeteredOccurrence` absorbs it). Past six hours of outage, self-healing stops silently: nothing raises, because the next sweep only ever looks six hours back. The docstring below spells out the manual recovery path (`meter_occurrences_for_period` per subscription in `subscriptions_to_sweep()`, then confirm with `reconcile_period`) since there's no command for it.

PR #199, opened 2026-07-20
feat(payments): enforce the postpaid event_occurrences allowance (billing plans and limits, phase 8)

hugobessa:
Phase 8 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Built on Phase 7 (`plan/billing-plans-and-limits/phase-7`, [#198](https://github.com/vintasoftware/vinta-schedule-api/pull/198)), which built the meter this phase gates against and has since merged into `main` — this PR targets `main` directly since `plan/billing-plans-and-limits/phase-7` was deleted on merge and `origin/main` now sits at exactly that commit.

## What this ships

An organization with a payment method accrues past its included `event_occurrences` allowance and is never interrupted; one without a payment method is blocked at the allowance with `remedy="add_payment_method"`. Two pieces make that true:

- `EntitlementService.has_payment_method(organization)` — resolved at the billing root, like every other check in this service — and `check_postpaid_allowance(organization, delta, lock, delta_resolver)` in `payments/services/entitlement_service.py`.
- Six named event-creation surfaces now call it: `CalendarEventService.create_event`, `CalendarBundleService.create_bundle_event`, the booking-code paths in `calendar_integration/mutations.py`, `public_api/mutations.py`'s `schedule_event`, and the bulk sync writer in `calendar_integration/services/calendar_sync_service.py`.

**Inert today.** Every organization currently exists on the `unlimited` plan — a NULL `event_occurrences` ceiling — and a NULL ceiling always returns headroom before any usage is even counted. Every guarded path carries an unlimited-behaviour test proving that: 201/success on every surface, no query issued that a finite ceiling would need. Nothing in this PR can block a real organization yet; it becomes load-bearing the moment the first non-NULL `event_occurrences` limit reaches production (Phase 14).

## Two commits

1. `c9643ce` — the guard, `has_payment_method`, the six call sites, and the initial test suites.
2. `0c7b705` — three BLOCKERs from review, all substantive, plus SHOULD-FIXes. This is the commit that matters.

## The three BLOCKERs review found, and why each one is the actual risk

**1. The lock was taken before the unlimited early-return — the phase was not inert.** `create_event` passes `lock=True`. The first draft of `check_postpaid_allowance` took `SELECT ... FOR UPDATE` on the billing root's `Subscription` row *before* checking `is_unlimited`, inside the same `@transaction.atomic` block that also holds an external provider round-trip. Every event create, for every organization, serialized org-wide behind a Google API call — for a NULL ceiling that could never block anyone. The unlimited tests only asserted the response code and couldn't see it. Fixed by resolving the effective limit first and locking only once a finite ceiling exists; a `CaptureQueriesContext` test now asserts no `FOR UPDATE` is issued on the unlimited path, with a finite-ceiling negative control proving the lock still fires when it should.

**2. `delta` and `current_usage` were different units.** `current_usage` counts `MeteredOccurrence` rows — occurrences. The first draft's `delta` counted `CalendarEvent` masters. A free organization one unit below its ceiling could create an open-ended daily series (`49 + 1 <= 50`) and accrue roughly 30 unbillable occurrences a month from it, forever, without the guard ever seeing it again. Fixed with a two-stage guard for the single-event path: stage 1 keeps `delta=1` (exact for a one-off, a deliberate lower bound for a recurring master) and runs before the provider write, so an organization with zero headroom is rejected before any external side effect. Stage 2 runs after `event.save()`, inside the same transaction, and expands the just-created master through `MeteringService.occurrence_starts_of` — made `@staticmethod` and public specifically so the guard and the meter share one definition of "how many units does this master cost" instead of two that can drift apart. If the series doesn't fit, the create rolls back. It has to run after the insert because the expansion is a Postgres function keyed on the event's id — there is no row to expand before it exists, and re-deriving the count in Python from the rrule string would be exactly the "two predicates that must agree, computed separately" shape that has produced a defect in every phase of this plan so far.

**3. `billing_state != FREE` granted unbounded accrual to an organization whose payment had already failed.** Replaced with an explicit allow-list, `PAYMENT_METHOD_BILLING_STATES = {ACTIVE}`. Every `BillingState` is pinned to an assertion in `EXPECTED_HAS_PAYMENT_METHOD`, so a newly-added state fails that test until someone decides for it — deliberately biased toward "no payment method" as the safe default, since reading `True` for a state with no working instrument lets an organization accrue overage nobody can be billed for, while reading `False` only asks it to attach a card.

The fixer went narrower than the review asked on BLOCKER 3, excluding `CANCELLED` from the allow-list: "a card was attached at some point" is the wrong tense for deciding whether overage can be billed right now, and a cancelled organization may well have removed its instrument along with the subscription. Both of the fixer's independent judgment calls moved toward the safe side.

## SHOULD-FIX cleared alongside the BLOCKERs

- The registry test that proves all six surfaces are guarded now derives the guarded set from an AST walk, rather than a same-file list that could silently drift from the real call sites.
- `bypass_limits` added to `create_event` and threaded through the bundle fan-out, matching the existing spelling/default on the pre-paid guards (`create_calendar` / `create_bundle_calendar` / `create_availability_windows`).
- The guard now honours `CalendarService.authenticate(bypass_limits=True)` via a new `bypass_entitlement_limits` field on `CalendarServiceContext` — before this, a facade explicitly placed in bypass mode still had its post-paid guard enforcing, because that gate reads the context rather than the facade attribute the provider-entitlement gate checks.
- `transfer_event` passes `_check_postpaid_allowance=False` — a move is not a creation (see the inline comment below).
- The sync-writer's over-allowance raise now logs the organization id, the unit count, current usage/ceiling, and the remedy before raising, with a recovery test — otherwise an over-allowance refusal and a flaky provider outage look identical in the `CalendarSync.error_message` field on every retry.

## Verification

- Full suite **4316 passed** (base 4262; +54).
- `mypy` **305 errors / 57 files** — at the same ceiling as Phase 7, zero new.
- `ruff` clean (502 files).
- `makemigrations --check` clean — this phase ships no schema change.

## Carried forward

**`has_payment_method` is a proxy, and two later phases are bound by it.** There is no payment-method record in the schema yet, so `has_payment_method` reads `Subscription.billing_state` through the allow-list above as a stand-in.

- **Phase 9 must re-point it at the real payment-method record and delete the `billing_state` proxy** — not widen the allow-list. Once an instrument is actually persisted, `billing_state` stops being evidence of anything.
- **Phase 10 must keep `GRACE` resolving to `False`.** Phase 10 moves `ACTIVE → GRACE` on a failed charge while leaving the organization fully operational — only `RESTRICTED` is write-blocked, in Phase 11. If `GRACE` read as "has a payment method", an organization whose card just declined would get unbounded, unbillable accrual for the entire dunning window, making the dunning ladder the largest-bill path in the product.

**The bundle fan-out residual.** `check_postpaid_allowance`'s unit is occurrences, and the single-event path now matches that exactly via the two-stage guard above. The bundle path does not: `_bundle_event_billable_units` counts `1 + n_internal_children` **masters**, checked once before the fan-out writes anything, with no equivalent second-stage expansion. A *recurring* bundle event is therefore under-charged by the same factor the single-event path used to be, before this PR. Bounded and inert today — every organization is `unlimited` — but it's a **Phase 13 gating precondition** alongside the identity-churn and bulk-modification-split defects Phase 7 already recorded: an over-allowance number that becomes money must be in occurrence units on every path, not just one.

### DIFF
--- a/calendar_integration/services/calendar_bundle_service.py
+++ b/calendar_integration/services/calendar_bundle_service.py
@@ -116,6 +116,7 @@ class BundleServiceHost(Protocol):
         calendar_id: int,
         event_data: CalendarEventInputData,
         *,
+        bypass_limits: bool = False,
         _enforce_policy: bool = True,
         _check_postpaid_allowance: bool = True,
     ) -> CalendarEvent: ...
@@ -402,7 +403,11 @@ class CalendarBundleService:
     # ------------------------------------------------------------------
 
     def create_bundle_event(
-        self, bundle_calendar: Calendar, event_data: CalendarEventInputData
+        self,
+        bundle_calendar: Calendar,
+        event_data: CalendarEventInputData,
+        *,
+        bypass_limits: bool = False,
     ) -> CalendarEvent:
         """
         Create an event in a bundle calendar by:
@@ -447,7 +452,16 @@ class CalendarBundleService:
         # recheck would be redundant rather than merely wasted work: nothing is
         # metered until a Celery sweep runs, so a re-check would ask the same
         # question against an unchanged usage count).
-        entitlement_service = self._context.entitlement_service
+        #
+        # Skipped when this call, or the whole service
+        # (``CalendarService.authenticate(bypass_limits=True)``), is in bypass mode --
+        # the same two conditions ``CalendarEventService._postpaid_entitlement_service``
+        # resolves for the single-event path.
+        entitlement_service = (
+            None
+            if bypass_limits or self._context.bypass_entitlement_limits
+            else self._context.entitlement_service
+        )
         if entitlement_service is not None:
             billable_units = self._bundle_event_billable_units(primary_calendar.id, child_calendars)
             result = entitlement_service.check_postpaid_allowance(
--- a/calendar_integration/services/calendar_bundle_service.py
+++ b/calendar_integration/services/calendar_bundle_service.py
@@ -117,6 +117,7 @@ class BundleServiceHost(Protocol):
         event_data: CalendarEventInputData,
         *,
         _enforce_policy: bool = True,
+        _check_postpaid_allowance: bool = True,
     ) -> CalendarEvent: ...
 
     def update_event(
@@ -439,6 +440,22 @@ class CalendarBundleService:
         # Get the designated primary calendar
         primary_calendar = self._get_primary_calendar(bundle_calendar)
 
+        # Postpaid ``event_occurrences`` allowance guard -- checked ONCE, for the whole
+        # fan-out, before anything is written. Each per-child ``create_event`` call
+        # below passes ``_check_postpaid_allowance=False`` so it is not re-checked
+        # (see ``CalendarEventService.create_event``'s docstring for why a per-child
+        # recheck would be redundant rather than merely wasted work: nothing is
+        # metered until a Celery sweep runs, so a re-check would ask the same
+        # question against an unchanged usage count).
+        entitlement_service = self._context.entitlement_service
+        if entitlement_service is not None:
+            billable_units = self._bundle_event_billable_units(primary_calendar.id, child_calendars)
+            result = entitlement_service.check_postpaid_allowance(
+                context.organization, delta=billable_units, lock=True
+            )
+            if not result.allowed:
+                raise OverLimitError.from_check_result(result)
+
         # Collect all attendees from child calendar ownerships
         all_attendees = self._collect_bundle_attendees(child_calendars, event_data)
 
@@ -461,7 +478,10 @@ class CalendarBundleService:
         # rejections when a child has its own stricter individual policy that would block
         # a booking the bundle policy (and Phase-5 discovery) correctly permits.
         primary_event = self._host.create_event(
-            primary_calendar.id, primary_event_data, _enforce_policy=False
+            primary_calendar.id,
+            primary_event_data,
+            _enforce_policy=False,
+            _check_postpaid_allowance=False,
         )
 
         # Mark primary event as part of bundle
@@ -474,7 +494,7 @@ class CalendarBundleService:
             if child_calendar.id == primary_calendar.id:
                 continue
 
-            if child_calendar.provider == CalendarProvider.INTERNAL:
+            if self._child_gets_full_event(child_calendar):
                 # Create full CalendarEvent for internal calendars
                 child_event_data = CalendarEventInputData(
                     title=f"[Bundle] {event_data.title}",
@@ -488,7 +508,10 @@ class CalendarBundleService:
                 )
 
                 child_event = self._host.create_event(
-                    child_calendar.id, child_event_data, _enforce_policy=False
+                    child_calendar.id,
+                    child_event_data,
+                    _enforce_policy=False,
+                    _check_postpaid_allowance=False,
                 )
 
                 # Link to primary event and bundle
@@ -647,6 +670,49 @@ class CalendarBundleService:
     # ------------------------------------------------------------------
     # Internal helpers
     # ------------------------------------------------------------------
+
+    @staticmethod
+    def _child_gets_full_event(child_calendar: Calendar) -> bool:
+        """Whether a non-primary bundle child gets a real ``CalendarEvent`` (rather
+        than a non-billable ``BlockedTime``) when a bundle event is created.
+
+        The single definition of that predicate. ``create_bundle_event``'s write
+        loop calls this to decide what to create, and
+        ``_bundle_event_billable_units`` calls the identical function to decide
+        what to charge for -- so "which children become billable rows" cannot be
+        answered two different ways by two pieces of code that drift apart over
+        time. The primary calendar is handled separately by its caller (it always
+        gets a real event, unconditionally); this only decides for the rest.
+        """
+        return child_calendar.provider == CalendarProvider.INTERNAL
+
+    @classmethod
+    def _bundle_event_billable_units(
+        cls, primary_calendar_id: int, child_calendars: list[Calendar]
+    ) -> int:
+        """How many ``event_occurrences`` units one bundle-event booking costs.
+
+        Binding decision (Phase 7 tracking doc, "what a bundle booking costs"):
+        **1 + n_internal_children**. The primary calendar always gets a real
+        ``CalendarEvent`` (the actual booking); every other child gets one too only
+        when ``_child_gets_full_event`` says so -- the exact predicate
+        ``create_bundle_event``'s write loop uses to decide "full CalendarEvent vs.
+        BlockedTime" for that same calendar. A ``BlockedTime`` is never billable
+        anywhere in this plan, so a bundle over five Google calendars costs 1, not
+        5, and this must never be computed a second, independent way.
+
+        Not derived by calling ``MeteringService.expand_occurrence_identities``:
+        nothing exists to expand yet -- the events this counts are about to be
+        created, not already in the database. This exists so the guard counts
+        exactly the rows the write loop below is about to write, using the same
+        predicate the write loop uses, rather than a second guess at what the
+        meter will later see.
+        """
+        return 1 + sum(
+            1
+            for calendar in child_calendars
+            if calendar.id != primary_calendar_id and cls._child_gets_full_event(calendar)
+        )
 
     def _get_primary_calendar(self, bundle_calendar: Calendar) -> Calendar:
         """Get the designated primary calendar for a bundle."""
--- a/calendar_integration/services/calendar_event_service.py
+++ b/calendar_integration/services/calendar_event_service.py
@@ -37,6 +37,7 @@ from typing import TYPE_CHECKING, Any, Protocol, cast
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
 from django.db.models import Q
+from django.utils import timezone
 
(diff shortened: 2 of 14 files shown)


payments/services/entitlement_service.py
```diff
@@ -526,6 +526,193 @@ def check_seat_limit_for_invitation_accept(
...
+        if effective_limit.is_unlimited:
+            return LimitCheckResult(
+                allowed=True,
+                resource_key=LimitedResource.EVENT_OCCURRENCES,
+                current_usage=None,
+                ceiling=None,
+            )
+
+        if lock:
+            self._lock_billing_root_row(root)
```
hugobessa, line 687: **The lock-after-unlimited ordering is the fix for BLOCKER 1, and it's deliberate, not incidental.**

`check_postpaid_allowance` resolves `effective_limit` and returns on `is_unlimited` *before* ever touching `_lock_billing_root_row`. The first draft locked first — and since every event-creation path passes `lock=True`, `create_event` is `@transaction.atomic` with an external provider round-trip inside it, and every organization today is `unlimited`, that meant every event create for every organization was serializing org-wide behind a Google round-trip, in service of a NULL ceiling that could never block anybody. Two users booking different calendars of the same organization would queue behind each other for no reason.

Nothing is lost by locking later: the ceiling itself isn't the racing quantity. `_count_usage` — the read the lock exists to serialize — still runs after the lock is acquired, so under READ COMMITTED it still sees a racing transaction's committed inserts. A `CaptureQueriesContext` test now asserts zero `FOR UPDATE` queries on the unlimited path, with a finite-ceiling case as the negative control proving the lock still fires when there's something real to protect.

payments/services/entitlement_service.py
```diff
@@ -526,6 +526,193 @@ def check_seat_limit_for_invitation_accept(
...
+    #: An explicit **allow-list**, deliberately, rather than ``!= FREE``. This is a
+    #: proxy (there is no payment-method record yet), and the two directions of
+    #: being wrong are not symmetric: reading ``True`` for a state with no working
+    #: instrument lets an organization accrue overage nobody can be billed for,
+    #: while reading ``False`` only asks it to attach a card. An allow-list makes a
+    #: newly-added state default to the second, safe direction; a deny-list makes it
+    #: default to the first. The plan's deferred ``TRIALING`` state
+    #: (``…IMPLEMENTATION_PLAN.md``, "no payment method to start") is exactly that
+    #: case, and would have read ``True`` under a negation.
+    PAYMENT_METHOD_BILLING_STATES = frozenset({BillingState.ACTIVE})
```
hugobessa, line 541: **An allow-list, not `!= FREE` — the fix for BLOCKER 3, and the asymmetry is the point.**

The two ways this predicate can be wrong aren't equally bad. Reading `True` for a state with no working payment instrument lets an organization accrue overage nobody can ever collect on; reading `False` for a state that does have one only asks it to attach a card again. An allow-list makes a newly-added `BillingState` default to the second, safe direction; the `!= FREE` deny-list it replaced would have defaulted to the first — and did, silently, for `GRACE` and `RESTRICTED` before this fix. `EXPECTED_HAS_PAYMENT_METHOD` in `test_postpaid_enforcement.py` pins every current state to an assertion, so a future state fails that test until someone decides for it rather than inheriting `True` by omission.

calendar_integration/services/calendar_event_service.py
```diff
@@ -398,16 +410,122 @@ def convert_naive_utc_datetime_to_timezone(
...
+
+        def resolve_delta(subscription: Subscription) -> int:
+            now = timezone.now()
+            _period_start, period_end = resolve_billing_period(subscription, now)
+            if period_end <= now:
+                return 1
+            starts = MeteringService.occurrence_starts_of(event, now, period_end)
+            return max(1, sum(1 for start in starts if now <= start < period_end))
+
+        result = entitlement_service.check_postpaid_allowance(
```
hugobessa, line 488: **Stage 2 of the two-stage guard — why it exists and why it can only run here.**

`current_usage` counts `MeteredOccurrence` rows (occurrences); the stage-1 guard just above this, like every other call site, charges `delta=1` per master. Exact for a one-off event, a large undercount for a recurring one — this is BLOCKER 2. A free organization one unit under its ceiling could otherwise create an open-ended daily series and accrue ~30 unbillable occurrences a month from it, forever, without the guard tripping on that series again.

`resolve_delta` calls `MeteringService.occurrence_starts_of` — made public in this PR for exactly this — over `[now, current billing period end)`, the same window the guard's `current_usage` was just read against. It has to run **after** `event.save()`: the expansion is a Postgres function keyed on the event's id, so a rule with no row can't be expanded, and re-deriving the count in Python from the rrule string would be a second, independent expansion — the "two predicates that must agree" shape this plan keeps producing defects from. `create_event` is `@transaction.atomic`, so a rejection here rolls the insert back. The one residual cost: an adapter-backed calendar's provider write may already have happened by this point, with no rollback for it — narrow, because stage 1 already rejected every organization with zero headroom at all, so this can only fire in the band where there's room for one booking but not the whole series.

payments/services/metering_service.py
```diff
@@ -389,11 +389,33 @@ def expand_occurrence_identities(
         return list(identities)
 
     @staticmethod
-    def _occurrence_starts_of(
+    def occurrence_starts_of(
```
hugobessa, line 392: **Renamed from `_occurrence_starts_of` and made public in this PR — the shared definition stage 2 depends on.**

Two callers now share this: the meter's own `expand_occurrence_identities` (turns each start into a billed identity) and the postpaid guard's stage 2 in `CalendarEventService._check_new_master_postpaid_allowance` (counts them to decide whether a new recurring master fits the remaining allowance). Before this phase, the guard charged 1 unit per master regardless of recurrence — see BLOCKER 2 above. There is now one function that answers "how many billable units does this master cost," called from both places, rather than the meter's expansion and the guard's estimate silently drifting apart the way this plan's other identity checks have.

calendar_integration/services/calendar_event_service.py
```diff
@@ -1897,7 +2048,21 @@ def transfer_event(self, event: CalendarEvent, new_calendar: Calendar) -> Calend
...
+        # against a ``current_usage`` that already includes the occurrences metered
+        # for the event being moved, and refuse an organization sitting at its
+        # allowance the right to move an event between its own calendars. A 402 for
+        # no net new billable capacity. (It is not perfectly net-zero for the
+        # *meter*: the new master has a new pk, so occurrences already metered under
+        # the old one are re-billed under the new identity. That is the known
+        # identity-churn defect recorded as a Phase 13 precondition, not something
+        # this guard could fix by charging a unit.)
+        new_event = self._host.create_event(
+            new_calendar.id, new_event_data, _check_postpaid_allowance=False
```
hugobessa, line 2064: **`transfer_event` opts out with `_check_postpaid_allowance=False` — a move, not a creation.**

The transfer pairs a `create_event` on the new calendar with a `delete_event` on the old one immediately after, net-zero on billable masters. Left enforced, the guard would compare `delta=1` against a `current_usage` that already includes the occurrences metered for the event being moved — refusing an organization sitting at its allowance the right to move an event between its own calendars, a 402 for zero net new billable capacity.

Noted as imperfect rather than papered over: the new master gets a new pk, so occurrences already metered under the old identity are re-billed under the new one at the next sweep. That's the identity-churn defect Phase 7 already recorded as a Phase 13 precondition — this guard opting out doesn't fix it, and charging a unit here wouldn't either, since the guard has no way to know the old event's occurrences are about to be superseded rather than duplicated.

calendar_integration/services/calendar_bundle_service.py
```diff
@@ -648,6 +685,49 @@ def delete_bundle_event(self, bundle_event: CalendarEvent) -> None:
...
+    def _child_gets_full_event(child_calendar: Calendar) -> bool:
+        """Whether a non-primary bundle child gets a real ``CalendarEvent`` (rather
+        than a non-billable ``BlockedTime``) when a bundle event is created.
+
+        The single definition of that predicate. ``create_bundle_event``'s write
+        loop calls this to decide what to create, and
+        ``_bundle_event_billable_units`` calls the identical function to decide
+        what to charge for -- so "which children become billable rows" cannot be
+        answered two different ways by two pieces of code that drift apart over
+        time. The primary calendar is handled separately by its caller (it always
```
hugobessa, line 698: **`_child_gets_full_event` is the one predicate both the write loop and the billing count use — and it's also where the residual noted in the PR description lives.**

`create_bundle_event`'s write loop calls this to decide `CalendarEvent` vs. non-billable `BlockedTime` per child; `_bundle_event_billable_units` calls the identical function to decide what to charge for. That agreement is deliberate — the same "two predicates that must mean the same thing" defect shape this plan keeps producing, avoided here by having one function instead of two.

What it doesn't do yet: the billable count is `1 + n_internal_children` **masters**, checked once before any child is written, with no second-stage expansion the way the single-event path now has via `MeteringService.occurrence_starts_of`. A recurring bundle event is under-charged by the same factor the single-event path was before this PR's BLOCKER 2 fix. Left as a Phase 13 gating precondition rather than fixed here — bounded and inert while every organization is `unlimited`, and fixing it means teaching this predicate to expand recurrence too, which is exactly the kind of change that should land together with Phase 13's overage charge rather than ahead of it.

PR #200, opened 2026-07-20
feat(payments): self-serve plan upgrade, downgrade, and add-on purchase (billing plans and limits, phase 9)

hugobessa:
Phase 9 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Built on Phase 7/8 (`plan/billing-plans-and-limits/phase-7`/`phase-8`, [#198](https://github.com/vintasoftware/vinta-schedule-api/pull/198)/[#199](https://github.com/vintasoftware/vinta-schedule-api/pull/199)), both merged to `main` mid-run — this PR is rebased directly onto `origin/main` and targets it. `git log --oneline origin/main..plan/billing-plans-and-limits/phase-9` shows only this phase's own commits.

## What this ships

The self-serve billing surface: an organization on the free plan can choose a paid plan, pay, and see its limits lift — no support or engineering intervention (spec objective 2).

- `payments/billing_views.py` — `SubscriptionViewSet` (`GET /billing/subscription/`, `POST .../change-plan/`, `POST .../cancel/`), `AddOnViewSet` (`POST /billing/add-ons/`, `DELETE .../{id}/`), plus read-only `BillingPlanViewSet` and `BillingUsageViewSet`.
- `IsBillingOwnerOrAdmin` (`organizations/permissions.py`) — admin-or-billing-owner-of-this-org, plus the reseller-subtree exception so an acting reseller root can manage a child's billing.
- `SubscriptionService`: `request_plan_change` / `confirm_plan_change` / `purchase_add_on` / `activate_add_on` / `cancel_add_on` / `cancel_subscription` / `record_payment_method` (`payments/services/subscription_service.py`).
- New `PaymentMethod` model (`payments/models.py`) — a real payment-instrument record, written only from the confirmed-webhook path.
- Provider-side proration on upgrade; downgrade takes no cash refund — the plan change is scheduled for the next period boundary while the lower limits apply immediately with a grace window.
- Capacity for both a plan change and an add-on purchase is granted **only on confirmed payment**, arriving through the Phase 2 webhook path (`confirm_plan_change` / `activate_add_on`), never synchronously from the request that initiates the charge.

## The Phase 8 precondition, discharged

Phase 8 left `EntitlementService.has_payment_method` reading a `Subscription.billing_state` allow-list (`PAYMENT_METHOD_BILLING_STATES = {ACTIVE}`) as a stand-in, because no payment-method record existed yet, and flagged that Phase 9 must re-point it at the real record rather than widen the allow-list. This PR does that: `has_payment_method` now queries `PaymentMethod(is_active=True)` at the billing root, and `PAYMENT_METHOD_BILLING_STATES` is deleted (survives only as docstring history). `GRACE`-with-a-card now correctly reads `True` — the case the old state-based proxy could not express, since Phase 10 moves `ACTIVE → GRACE` on a failed *charge*, which says nothing about whether the card itself is still attached.

## The two money-safety BLOCKERs — the substance of this PR

Both are the same trap: **`ATOMIC_REQUESTS` in production**. The inner `transaction.atomic()` used for a dedup/lock row is a *savepoint* inside the request's outer transaction in production, not an independently-committing transaction — the dedup row only becomes durable when the whole request commits, which is *after* the provider charge runs. The orchestrator's own spot-check missed both, having assumed the inner block commits before the charge; test settings omit `ATOMIC_REQUESTS`, so every test passed regardless.

1. **`purchase_add_on` could double-charge on a crash between the charge and the request commit.** The local `SubscriptionAddOn.objects.get_or_create(purchase_idempotency_key=...)` dedup row is not durable until the request transaction commits; a crash after `payment_service.create_payment(...)` but before that commit rolls the row back, and a retry finds nothing and charges again.
2. **`request_plan_change` had no working idempotency at all.** `ChangePlanRequestSerializer` required an `idempotency_key`, but it was never threaded past the view into the service or the provider call — concurrent first-upgrade requests could both observe a blank `external_id` and both drive the provider.

**Fix: provider-side idempotency**, not just a local dedup row. The client-supplied `idempotency_key` now reaches the provider's own idempotency mechanism on the money-moving call — Stripe `Idempotency-Key` (`stripe.PaymentIntent.create(idempotency_key=...)`), MercadoPago `x-idempotency-key` header — so the provider itself refuses a second charge for the same key even when the local dedup row didn't survive a rollback or process restart. `request_plan_change` additionally takes a `SELECT ... FOR UPDATE` on the subscription row so two concurrent first-upgrade requests serialize: the second blocks until the first commits, then sees the settled result and no-ops rather than charging again.

Fixing this also surfaced a pre-existing bug in `MercadoPagoPaymentAdapter.process`: the idempotency header was being built from `{payment.id}` — a **set literal**, not a string — which serialized wrong regardless of this phase. Now `idempotency_key or str(payment.id)`.

**Lesson recorded in the tracking doc:** green gates under test settings are not evidence about `ATOMIC_REQUESTS` behavior. Any charge/commit ordering has to be reasoned about against the production transaction wrapper explicitly, not inferred from a passing suite.

## Three more of the plan's signature two-predicates fixes

- **Downgrade-revert on APPROVED redelivery.** `confirm_plan_change` used to unconditionally re-copy limits from `subscription.plan`. A redelivered webhook during a downgrade's grace window would silently restore the *higher*, already-superseded plan's ceiling. It now checks whether a scheduled downgrade is still in its grace window and, if so, syncs from the pending (lower) plan instead — `subscription.plan` itself stays on the paid plan until the Phase 13 boundary sweep flips it.
- **Plan-divergence 409.** A second upgrade requested before the first confirmed used to grant whatever plan `subscription.plan` last pointed at when the webhook landed — the *latest* requested tier, not the one that charge actually paid for. `Subscription.plan_change_pending_confirmation` (new field, migration `0013`) now rejects a *different* plan change with `409 plan_change_pending_confirmation` until the in-flight one confirms; re-requesting the *same* plan/interval stays a no-op.
- **Object-level permission DENY, now tested on both write paths**, and a `ScopedRateThrottle` (`billing-write`) on every billing write action.

## Gates

- Full suite **4384 passed** (base 4316; +68) + 1 known flake — `payments/tests/services/test_limit_concurrency.py::test_lock_does_not_block_when_there_is_room_for_both`, a threaded lock-timing test that passes 3/3 in isolation and reddens only under full parallel load (same class as Phase 8's 6b negative control).
- `mypy` **305 errors / 57 files** — same ceiling as Phase 7/8, zero new.
- `ruff` clean (506 files).
- `makemigrations --check` clean; `schema.yml` regenerated (+976/−9).
- Migrations `0012` (`pending_billing_interval`/`pending_plan_effective_at`/`PaymentMethod`) and `0013` (`plan_change_pending_confirmation`) both reverse cleanly.
- The money-safety subset (131 tests covering the two BLOCKERs above) re-run green independently.

## Carried forward

- ⚠️ **`MercadoPagoPaymentAdapter.refund` has the same idempotency-key bug** the fixer just corrected in `process` — it still builds `x-idempotency-key` from a `{refund.id}` set literal. Latent (not exercised by any test that would catch the serialization), not in scope for this phase, left untouched. Worth a follow-up.
- The adapter idempotency-key plumbing is **unit-tested only** — no live Stripe/MercadoPago credentials exist in this environment (same caveat Phase 2b and Phase 7 left).
- Phase 10 (dunning) inherits two hard constraints from this phase, recorded in the tracking doc: `GRACE` must keep `has_payment_method` reading `True` via the real `PaymentMethod` record (do not reintroduce a `billing_state`-based answer), and Phase 10 must clear `plan_change_pending_confirmation` when a failed initial charge moves the subscription to `GRACE`/`RESTRICTED`, or the org is stuck unable to switch plans.


payments/services/subscription_service.py
```diff
@@ -325,6 +378,436 @@ def change_plan(self, subscription: Subscription, plan: BillingPlan) -> Subscrip
...
+            if subscription.plan_change_pending_confirmation:
+                raise UnconfirmedPlanChangeError(subscription.organization_id)
+
+            current_price = self._plan_price(subscription.plan, subscription.billing_interval)
+            new_price = self._plan_price(plan, billing_interval)
+            if new_price > current_price:
+                return self._initiate_upgrade(
+                    subscription, plan, billing_interval, payment_token, idempotency_key
+                )
+            return self._schedule_downgrade(subscription, plan, billing_interval)
```
hugobessa, line 466: **The `SELECT ... FOR UPDATE` here is half of BLOCKER 2's fix — provider-side idempotency is the other half, and neither alone was enough.**

Before this, `request_plan_change` had no working idempotency at all: `ChangePlanRequestSerializer` required an `idempotency_key`, but nothing threaded it past the view. Two concurrent first-upgrade requests for the same subscription could both read a blank `external_id` and both call the provider.

The lock closes the *concurrent-request* race: the second request blocks on `select_for_update()` until the first commits, then re-evaluates `already_settled` under the lock and no-ops instead of charging again. It does **not** close the *crash-after-charge-before-commit* race by itself — under production `ATOMIC_REQUESTS` the lock (and every write in this method) lives inside a savepoint that only becomes durable at request end, same trap as BLOCKER 1. `idempotency_key` is forwarded to the provider's own idempotency mechanism (Stripe `Idempotency-Key` / MercadoPago `x-idempotency-key`) precisely because that's the part of the fix that survives a rollback the lock can't protect against.

payments/services/payment_adapters/base.py
```diff
@@ -34,11 +34,20 @@ class Meta:
...
         :param payment: Payment object
         :param payment_token: Payment token
+        :param idempotency_key: A caller-supplied, stable key forwarded to the
+            provider as its own idempotency key (Stripe ``Idempotency-Key`` /
+            MercadoPago ``x-idempotency-key``). This is what makes a charge safe
+            to *retry* — the provider itself refuses to charge the same key
+            twice, independently of whether any local dedup row survived a
+            rollback or process restart. Empty falls back to a per-attempt
+            provider default (e.g. the local payment id), which does **not**
+            survive a Payment row being re-created on retry, so callers that need
```
hugobessa, line 49: **Provider-side idempotency, threaded onto the abstract signature so every adapter (and every caller) carries it — the actual fix for both BLOCKERs, not just a parameter add.**

Both BLOCKER 1 (`purchase_add_on`) and BLOCKER 2 (`request_plan_change`) had a local dedup row (a `get_or_create`'d model row, or a `SELECT ... FOR UPDATE`) that looked sufficient under test settings but isn't durable in production: `ATOMIC_REQUESTS` makes the surrounding `transaction.atomic()` a savepoint, so the dedup row commits *after* the charge, not before it. A crash in that window loses the dedup row and a retry re-charges.

Forwarding `idempotency_key` here moves the guarantee to where it can't be lost: the provider itself refuses a second charge for the same key, independent of whether any local row survived. The docstring calls out the corollary explicitly — an empty key falls back to a per-attempt default that does *not* survive a `Payment` row being re-created on retry, so a caller that needs retry-safety must pass a stable key. Fixing this also caught a pre-existing, unrelated bug in `MercadoPagoPaymentAdapter.process`: the old header value was `{payment.id}`, a set literal rather than a string.

payments/services/subscription_service.py
```diff
@@ -325,6 +378,436 @@ def change_plan(self, subscription: Subscription, plan: BillingPlan) -> Subscrip
...
+        only the first one after an upgrade) — both sync paths are bulk upserts —
+        and clears ``plan_change_pending_confirmation`` so a further plan change
+        is allowed again.
+        """
+        pending_plan = subscription.pending_plan
+        if pending_plan is not None and self._pending_downgrade_is_future(subscription):
+            self._sync_limits(subscription, pending_plan)
+            self._sync_entitlements(subscription, pending_plan)
+        else:
+            self.change_plan(subscription, subscription.plan)
```
hugobessa, line 620: **The branch that fixes the downgrade-revert-on-redelivery defect.**

Before this phase's fix, `confirm_plan_change` unconditionally re-copied limits from `subscription.plan`. That's correct for an upgrade confirmation, but wrong for a redelivered `APPROVED` webhook arriving *during* a scheduled downgrade's grace window: `_schedule_downgrade` already applied the lower pending plan's limits immediately while leaving `subscription.plan` on the still-paid higher plan (no refund, remember), so re-copying from `subscription.plan` would silently lift the ceiling back up mid-downgrade on every redelivery.

`_pending_downgrade_is_future` is the one predicate this now keys off — pending plan set *and* its effective date still ahead — so both branches read the same fact rather than two independently-maintained checks drifting apart, the plan's recurring failure shape elsewhere avoided the same way.

payments/services/entitlement_service.py
```diff
@@ -526,68 +526,51 @@ def check_seat_limit_for_invitation_accept(
...
+        row either, so this still reads ``False`` for it — nothing to charge.
         Note that on the postpaid path this rarely decides anything — a
         subscription-less pool resolves to an unlimited ceiling and returns before
         this is ever consulted (see ``check_postpaid_allowance``).
         """
-        return self._has_payment_method_for_subscription(self._get_root_subscription(organization))
+        root = resolve_billing_root(organization)
+        return self._has_payment_method_for_organization_id(root.pk)
 
     @classmethod
```
hugobessa, line 563: **`has_payment_method` re-pointed at the real `PaymentMethod` record — the Phase 8 precondition discharged, not widened.**

Phase 8 answered this from `Subscription.billing_state` through an allow-list (`PAYMENT_METHOD_BILLING_STATES = {ACTIVE}`) because no payment-method record existed yet, and explicitly flagged that Phase 9 must re-point the method at a real record rather than add more states to the allow-list — `billing_state` stops being evidence of anything once an instrument is actually persisted. This PR deletes `PAYMENT_METHOD_BILLING_STATES` entirely and queries `PaymentMethod.objects.filter(organization_id=..., is_active=True).exists()` instead.

This is what makes `GRACE`-with-a-card read `True` for the first time: Phase 10 moves `ACTIVE -> GRACE` on a failed *charge*, which says nothing about whether the card itself is still attached, and the old state-based proxy had no way to distinguish "card is fine, next retry will succeed" from "card is gone." The tracking doc pins this as a hard constraint for Phase 10: do not reintroduce a `billing_state`-based answer here even under pressure to unblock dunning — if Phase 10 needs to block writes, it does so through `RESTRICTED` (Phase 11), not by lying about the payment method.

payments/services/subscription_service.py
```diff
@@ -325,6 +378,436 @@ def change_plan(self, subscription: Subscription, plan: BillingPlan) -> Subscrip
...
+                    "is_recurring": is_recurring,
+                    "is_active": False,
+                },
+            )
+        if not created:
+            return add_on
+
+        payment_service = self._require_payment_service()
+        payment = payment_service.create_payment(
+            organization=subscription.organization,
```
hugobessa, line 730: **Row created (`is_active=False`) before the charge, not after — the ordering BLOCKER 1's fix depends on.**

`get_or_create(purchase_idempotency_key=...)` commits a `SubscriptionAddOn` row *before* `payment_service.create_payment(...)` runs, and the method returns immediately if the row already existed (`if not created: return add_on`). That ordering is what makes "the same key posted twice always resolves to the same row" true — but as the docstring above this block says explicitly, the row's local commit is not by itself durable against a crash between the charge and the request committing under `ATOMIC_REQUESTS` (same trap as BLOCKER 2). The provider-side `idempotency_key` forward on `create_payment` is what actually closes that gap; the row-first ordering only guarantees the *local* half — a second request with the same key never even reaches the provider call once the row exists.

  Capacity itself is granted later, in `activate_add_on`, only once the resulting payment confirms `APPROVED` via the webhook — this method never flips `is_active` synchronously, matching `request_plan_change`'s "no capacity from the request path" rule.

payments/billing_views.py
```diff
@@ -0,0 +1,342 @@
...
+                data["billing_interval"],
+                payment_token=data.get("payment_token", ""),
+                idempotency_key=data["idempotency_key"],
+            )
+        except PaymentTokenRequiredError as error:
+            raise ValidationError({"payment_token": str(error)}) from error
+        except UnconfirmedPlanChangeError as error:
+            # A different plan change is already in flight and unconfirmed --
+            # 409 Conflict rather than a validation error on any one field.
+            return Response({"detail": str(error)}, status=status.HTTP_409_CONFLICT)
```
hugobessa, line 230: **`plan_change_pending_confirmation` surfaced as `409 Conflict`, not a field-level validation error — the fix for the second plan-divergence defect.**

Before `Subscription.plan_change_pending_confirmation` (migration `0013`) existed, a second upgrade requested before the first confirmed would overwrite `subscription.plan`, so the *first* charge's confirming webhook would grant whatever plan was requested most recently rather than the plan that charge actually paid for — a customer could pay for tier B and end up entitled to tier C's limits by requesting it a moment later. `request_plan_change` now raises `UnconfirmedPlanChangeError` for a *different* plan while one is in flight, caught here as `409` rather than `400`: this isn't a bad request, it's a legitimate request that has to wait for a prior one to settle. Re-requesting the *same* plan/interval never reaches this check — it's a no-op via the `already_settled` branch.

  Known limitation, not fixed in this phase: a first-upgrade whose charge *fails* leaves this flag set with no `APPROVED` webhook ever arriving to clear it, so the org is stuck unable to change plans until Phase 10's dunning path clears it on the failed-charge transition (recorded as a Phase 10 precondition in the tracking doc).

PR #201, opened 2026-07-21
feat(payments): grace period, dunning ladder, and restricted state (billing plans and limits, phase 10)

hugobessa:
Phase 10 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Built on Phase 9 (`plan/billing-plans-and-limits/phase-9`, [#200](https://github.com/vintasoftware/vinta-schedule-api/pull/200)), merged to `main` mid-run — this PR is rebased directly onto `origin/main` and targets it. `git log --oneline origin/main..plan/billing-plans-and-limits/phase-10` shows only this phase's own commits.

## What this ships

The lifecycle that answers "what happens when a subscription's recurring charge fails": a grace period, an escalating dunning ladder of retries and notifications, and the transition into `RESTRICTED`.

- `payments/services/billing_state_machine.py` — `LEGAL_BILLING_STATE_TRANSITIONS` (a `frozenset` of `(from_state, to_state)` pairs encoding the spec's lifecycle diagram plus the cancel edges the product needs) and `transition_billing_state`, the **single validated write path** for `Subscription.billing_state`. Every caller that changes `billing_state` — every webhook handler in `payments/views.py`, `SubscriptionService.confirm_plan_change` / `cancel_subscription`, and the `process_dunning` beat task via `DunningService` — routes through it now; nothing writes the field directly.
- `payments/services/dunning_service.py` — `DunningService`: `enter_grace` (`ACTIVE|FREE → GRACE` on a failed charge), `resolve_payment_success` (`GRACE|RESTRICTED → ACTIVE`), `expire_grace` (`GRACE → RESTRICTED`, or `GRACE → FREE` if usage now fits under the free plan), `check_free_fallback`, and `cancel`. Drives the notification ladder (in-app + email) at each rung.
- `BILLING_DEFAULT_GRACE_PERIOD_DAYS` setting (`vinta_schedule_api/settings/base.py`) — the fallback grace window when a plan doesn't set its own `grace_period_days`.
- `Subscription.last_dunning_attempt_at` (migration `0014`) — tracks the last retry so the throttle and the beat task's re-entrancy can agree on what already happened.
- Dunning-ladder email + in-app templates (`templates/payments/emails/dunning_*`, `templates/payments/in_app/entered_grace.body.txt`) and `payments/notification_contexts.py` context builders.
- `process_dunning` / `process_dunning_for_subscription` beat task (`payments/tasks.py`, scheduled in `vinta_schedule_api/celerybeat_schedule.py`) — hourly sweep over `GRACE`/`RESTRICTED` subscriptions that drives `DunningService.process_subscription` per row.

## Both inherited constraints held

Phase 9 left two hard constraints on this phase, both confirmed by the reviewer, not just asserted:

1. **`has_payment_method` stays `True` for a `GRACE` org with a card on file.** `DunningService` never touches `PaymentMethod` — a failed charge says nothing about whether the card is still attached, so postpaid accrual keeps working through the whole dunning window; the ladder, not the postpaid guard, is what escalates.
2. **`enter_grace` clears `plan_change_pending_confirmation`** in the same transaction that moves a subscription into `GRACE`, so a first-upgrade whose initial charge fails doesn't leave the organization stuck unable to request a different plan (Phase 9's flag has no other path that clears it on a failed charge, since the webhook never reaches the `APPROVED` branch).

## The BLOCKER — the substance of this PR

**Two clocks quantizing the same decision differently, found in review, not in the implementer's own spot-check.** The retry throttle (`MIN_DUNNING_RETRY_INTERVAL = 20h`, a rolling window) and the charge's idempotency key (originally `%Y-%m-%d`, a 24h calendar-day bucket) measured time on different clocks. A first retry landing 00:00–03:59 UTC would fire its *second* attempt 20 hours later — still the *same calendar day* — reuse the same `x-idempotency-key`, get silently deduplicated by the provider (**no real charge**), while `DunningService` still emailed the customer "we retried and it failed." Roughly 1 in 6 subscriptions (those whose first attempt landed in that four-hour window) would silently skip a genuine collection attempt — a direct breach of the phase's own "retry across the grace window" acceptance criterion. It was invisible because the schedule tests froze the clock at 09:00, never exercising the boundary.

**Fix: derive both decisions from one shared retry-bucket ordinal.** `_retry_attempt_ordinal(subscription, at)` computes `floor((at - grace_start) / MIN_DUNNING_RETRY_INTERVAL)` once; `_process_grace` uses it to decide *whether* to retry (a new attempt only fires when `now` falls in a later bucket than `last_dunning_attempt_at` did), and `_retry_charge_and_notify` uses the same number as the idempotency key (`dunning-retry-{subscription.pk}-{attempt_ordinal}`). Two views of one number can't disagree — a redelivery of the same logical attempt always lands in the same bucket and reuses its key; a genuinely new retry always gets a new one.

## Three SHOULD-FIX

- **`cancel_subscription` routed through the state machine.** It used to write `billing_state = CANCELLED` directly, the one caller that didn't go through `transition_billing_state`. Now it does, which required adding the `FREE|GRACE|RESTRICTED → CANCELLED` edges to the table beyond the diagram's single `ACTIVE → CANCELLED` — justified inline, since the product's cancel action is offered from any live state. Dunning bookkeeping (`grace_period_ends_at`, `last_dunning_attempt_at`) is cleared in the same write.
- **Free-fallback moved from every `GRACE` tick to grace expiry.** It used to run on every tick, so a payer whose usage happened to dip under the free plan's ceiling mid-window was dropped to `FREE` on the very next tick and never got the rest of the retry ladder — even though a card was on file and the org still owed money. `check_free_fallback` now only runs from `expire_grace`, once the window is unresolved; the full ladder runs first.
- **The downgrade→`GRACE` dead-edge gap, documented as a `TODO(phase-11)`.** `LEGAL_BILLING_STATE_TRANSITIONS` lists `(ACTIVE, GRACE)` and `(FREE, GRACE)` for two reasons each: payment failure (this phase's driver) and "downgrade leaves org over limit" (the diagram's other reason, no driver yet — `SubscriptionService._schedule_downgrade` stamps `grace_period_ends_at` but leaves `billing_state` unchanged, so `process_dunning`'s sweep, which only looks at `GRACE`/`RESTRICTED` rows, never sees it). Both edges carry an inline `TODO(phase-11)` marker; Phase 11 owns adding the driver or moving the enforcement onto a swept path.

## Gates

- Full suite **4451 passed** (base 4384; +67).
- `mypy` **305 errors / 57 files** — same ceiling as prior phases, zero new.
- `ruff` clean (511 files).
- `makemigrations --check` clean.
- Migration `0014` (`last_dunning_attempt_at`) reverses cleanly.

## Carried forward

- The downgrade-over-limit → `GRACE` driver gap (above) — Phase 11's to close, either by adding the driver or moving the over-limit-after-downgrade enforcement onto a path `process_dunning` sweeps.
- ⚠️ **`MercadoPagoPaymentAdapter.refund`'s `{refund.id}` idempotency-key bug** (a set literal, not a string — from Phase 9) still stands. Not touched by this phase's retry path, which uses the correct provider-idempotency plumbing (`change_subscription_plan(..., idempotency_key=...)`), just not copied into `refund`. Still worth a follow-up.
- **`_grace_period_days`'s ordinal anchor recomputes `grace_start` from the *current* plan's `grace_period_days`** rather than a value snapshotted when grace began. Editing a plan's grace window mid-episode would shift the retry-bucket anchor slightly for subscriptions already in `GRACE` on that plan. Low-risk edge (plan grace windows aren't expected to change often, and a shifted anchor only nudges retry timing, not correctness); deriving without adding a new column was the chosen schema-free fix.


payments/services/billing_state_machine.py
```diff
@@ -0,0 +1,115 @@
...
+        (BillingState.FREE, BillingState.CANCELLED),  # cancel a free-tier subscription
+        (BillingState.GRACE, BillingState.CANCELLED),  # give up on dunning and cancel
+        (BillingState.RESTRICTED, BillingState.CANCELLED),  # cancel instead of paying to recover
+        (BillingState.CANCELLED, BillingState.FREE),  # cycle ends (Phase 13 sweep)
+    }
+)
+
+
+def is_legal_billing_state_transition(from_state: str, to_state: str) -> bool:
+    """Whether ``from_state -> to_state`` is on the diagram (same-state included --
```
hugobessa, line 76: **The single transition table — every `billing_state` write in the codebase now goes through `transition_billing_state`, which validates against this set.**

Before this phase, `SubscriptionService.cancel_subscription` wrote `billing_state = CANCELLED` directly (fixed in the SHOULD-FIX round below), and there was no shared place enumerating which transitions the spec's lifecycle diagram actually allows. This `frozenset` of `(from_state, to_state)` pairs is that place: it encodes the diagram's edges plus the handful of cancellation edges the product's cancel action needs beyond the diagram's single `ACTIVE -> CANCELLED` (each justified inline, e.g. `RESTRICTED -> CANCELLED` for "cancel instead of paying to recover").

Two `TODO(phase-11)` markers sit on `(ACTIVE, GRACE)` and `(FREE, GRACE)` — both edges have a second reason on the diagram ("downgrade leaves org over limit") with no driver yet; only the payment-failure driver (`DunningService.enter_grace`) exists today. Phase 11 inherits that gap.

payments/services/dunning_service.py
```diff
@@ -0,0 +1,498 @@
...
+        the single shared quantity the throttle gate and the idempotency key
+        both read: two attempts in different buckets are two distinct charges
+        with two distinct keys; a redelivery of the same attempt lands in the
+        same bucket and reuses its key.
+        """
+        grace_ends_at = subscription.grace_period_ends_at
+        if grace_ends_at is None:
+            return 0
+        grace_start = grace_ends_at - datetime.timedelta(days=self._grace_period_days(subscription))
+        return (at - grace_start) // MIN_DUNNING_RETRY_INTERVAL
```
hugobessa, line 282: **The fix for this PR's BLOCKER — one retry-bucket ordinal feeding both the throttle decision and the charge's idempotency key.**

Before this fix, the retry throttle (`MIN_DUNNING_RETRY_INTERVAL = 20h`, a rolling window) and the idempotency key (a `%Y-%m-%d` calendar-day bucket) quantized time on two different clocks. A first retry landing 00:00-03:59 UTC would open the throttle gate again 20h later on the *same calendar day*, reuse the prior attempt's key, and get silently deduplicated by the provider — no charge, but the customer still got the "we retried and failed" email. About 1 in 6 subscriptions could hit that window.

`floor((at - grace_start) / MIN_DUNNING_RETRY_INTERVAL)` is now the one number both `_process_grace` (line ~256, gates on whether `current_ordinal` moved past the bucket `last_dunning_attempt_at` fell in) and `_retry_charge_and_notify` (line 303, keys the charge `dunning-retry-{subscription.pk}-{attempt_ordinal}`) read. Two views of one quantity can't disagree — a redelivery of the same logical attempt always lands in the same bucket; a genuinely new retry always gets a new one.

payments/services/dunning_service.py
```diff
@@ -0,0 +1,498 @@
...
+        not leave the organization stuck.
+
+        Does **not** touch ``PaymentMethod`` -- see module docstring.
+        """
+        if subscription.billing_state in (
+            BillingState.GRACE,
+            BillingState.RESTRICTED,
+            BillingState.CANCELLED,
+        ):
+            return subscription
```
hugobessa, line 162: **`enter_grace`'s idempotency is an early-return no-op, not a re-stamp.**

A failed-charge webhook can legitimately redeliver, or fire again while the subscription is already further along the ladder (`GRACE`, `RESTRICTED`) or done with it (`CANCELLED`) — none of those are a *source* state for this edge on the diagram. Rather than let `transition_billing_state` raise on a same-state or off-diagram call, this returns unchanged before the transition is even attempted. That matters beyond just avoiding an exception: it also means a second `enter_grace` call for an already-`GRACE` subscription does **not** re-stamp `grace_period_ends_at` further into the future — a naive redelivery-tolerant implementation that re-ran the whole body would silently extend the grace window on every duplicate webhook.

payments/services/dunning_service.py
```diff
@@ -0,0 +1,498 @@
...
+            transition_billing_state(subscription, BillingState.GRACE)
+            grace_days = self._grace_period_days(subscription)
+            subscription.grace_period_ends_at = timezone.now() + datetime.timedelta(days=grace_days)
+            subscription.last_dunning_attempt_at = None
+            update_fields = ["grace_period_ends_at", "last_dunning_attempt_at"]
+            if subscription.plan_change_pending_confirmation:
+                subscription.plan_change_pending_confirmation = False
+                update_fields.append("plan_change_pending_confirmation")
+            subscription.save(update_fields=update_fields)
+            transaction.on_commit(lambda: self._notify_entered_grace(subscription))
```
hugobessa, line 174: **Where Constraint 2 from Phase 9 actually gets discharged: `plan_change_pending_confirmation` clears in the same transaction that moves the subscription into `GRACE`.**

Phase 9's `request_plan_change` sets this flag on a first-upgrade attempt and only clears it from the `APPROVED` webhook branch (`confirm_plan_change`) — a charge that *fails* never reaches that branch, so without this, a declined first-upgrade would leave the flag stuck `True` forever, and the org could never request a different plan. `enter_grace` clears it here, conditionally (`if subscription.plan_change_pending_confirmation:`) so a subscription that was never in that state doesn't pick up a spurious extra field in `update_fields`.

payments/services/dunning_service.py
```diff
@@ -0,0 +1,498 @@
...
+
+        Free-fallback is deliberately **not** checked here: an org that entered
+        GRACE from a *payment failure* but happens to already sit under the
+        free-plan ceilings still owes money on a card that is on file, so it
+        gets the full retry ladder across the grace window; ``expire_grace``
+        decides FREE vs. RESTRICTED only once the window is unresolved (the
+        diagram's ``Grace -> Free`` edge is the downgrade-under-limit path, not
+        the payment-failure path -- a distinction a GRACE subscription carries
+        no reason field to make mid-window, so the only safe place to collapse
+        it is at expiry).
```
hugobessa, line 241: **Free-fallback deliberately does *not* run here — one of the three SHOULD-FIX findings.**

The original implementation checked `check_free_fallback` on every `GRACE` tick, so a payer whose usage happened to dip under the free plan's ceiling mid-window got dropped straight to `FREE` on the very next tick — even with a card on file and a genuine unpaid charge outstanding. That skipped the rest of the retry ladder for anyone whose usage pattern is naturally bursty. The fix moved the free-fallback check to `expire_grace` (this file, ~line 317) instead: it now only resolves the GRACE-vs-FREE-vs-RESTRICTED question once the full grace window has elapsed unresolved, so every payer gets the complete ladder of retries and escalating notifications first.

payments/services/dunning_service.py
```diff
@@ -0,0 +1,498 @@
...
+- **Never touches ``PaymentMethod``.** A failed charge says nothing about
+  whether the card is still attached -- ``EntitlementService.has_payment_method``
+  must keep reading ``True`` for a GRACE organization with a card on file, so it
+  keeps accruing postpaid usage; the dunning ladder, not the postpaid guard, is
+  what escalates it.
+- **Clears ``plan_change_pending_confirmation``** whenever it moves a
+  subscription into GRACE, so a first-upgrade whose initial charge fails does
+  not leave the organization stuck unable to request a different plan (Phase 9
+  set the flag; a failed charge never reaches the APPROVED webhook branch that
+  would otherwise clear it).
```
hugobessa, line 27: **`has_payment_method` is untouched by this phase — called out here since `entitlement_service.py` doesn't appear in this diff at all.**

Phase 9 made `has_payment_method` query the real `PaymentMethod` record instead of a `billing_state` allow-list, specifically so a `GRACE` organization with a card on file would correctly read `True` (a failed *charge* says nothing about whether the instrument is still attached). This phase's module docstring restates that as a hard constraint (`Never touches PaymentMethod`) precisely because it would have been easy to reach for `billing_state` again while building the dunning ladder — `DunningService` never writes or reads `PaymentMethod`, so the postpaid guard and the dunning escalation stay two independent signals, confirmed by the reviewer rather than just asserted.

PR #202, opened 2026-07-21
feat(payments): enforce RESTRICTED write block and calendar sync pause (billing plans and limits, phase 11)

hugobessa:
Phase 11 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Built on Phase 10 (`plan/billing-plans-and-limits/phase-10`, [#201](https://github.com/vintasoftware/vinta-schedule-api/pull/201)), **still unmerged** — this PR targets `plan/billing-plans-and-limits/phase-10`, not `main`. `git log --oneline plan/billing-plans-and-limits/phase-10..plan/billing-plans-and-limits/phase-11` shows only this phase's own commits.

## What this ships

Phase 10 built the lifecycle that gets a subscription into `RESTRICTED`. This phase is the payoff: `RESTRICTED` actually restricts something.

- `payments/services/entitlement_service.py` — `EntitlementService.is_billing_root_restricted`: the **one predicate**, resolved at the billing root, answering "must this organization's writes be blocked and its calendar sync paused?" `check_not_restricted` wraps it as a raise (`OverLimitError`, `remedy=resolve_billing`) for callers that don't already route through `check_limit` / `check_postpaid_allowance`, which fold the same test in directly as a hot-path short-circuit ahead of any capacity check.
- Write-block coverage: every explicit `check_not_restricted` call site on an update/delete path this phase adds — `GoogleCalendarServiceAccount` CRUD (`organizations/views.py`), service-account rotation, membership deactivate/role-change, `SystemUserTokenViewSet.deactivate` (`public_api/views.py`) — plus the two hot-path short-circuits already inlined into `check_limit`/`check_postpaid_allowance`.
- Sync-pause coverage: `calendar_integration/services/calendar_sync_service.py`'s `_check_not_restricted`, called from all three `request_*` methods (`request_calendar_sync`, `request_calendars_import`, `request_organization_calendar_resources_import`) before anything is enqueued, plus `calendar_integration/tasks/calendar_sync_tasks.py`'s `_restricted_or_skip` early return in every task body — defense in depth for work already queued when an organization becomes restricted mid-flight.
- Resync-on-recovery: `DunningService._trigger_resync_after_recovery`, called from both `resolve_payment_success` and `check_free_fallback` when the prior state was `RESTRICTED`. Fans `resync_organization_calendars_task` out per pooled organization on `transaction.on_commit`; the task reconciles a real drift window (30-day lookback / 365-day lookahead) rather than just resuming sync from "now."
- `payments/services/subscription_service.py`'s `_schedule_downgrade` now drives `billing_state` into `GRACE` through `transition_billing_state` — closing the Phase 10 dead edge (see below).
- Reads, exports, auth, and every `/billing/` write stay open by construction: `SubscriptionService` (the module backing `/billing/`) calls none of these guards, so a restricted organization can always pay its way out. `GRACE` is never blocked by any of this — only `RESTRICTED` blocks; a `GRACE` organization stays fully writable and its sync keeps running.
- Reseller cascade is automatic, not a separate code path: `is_billing_root_restricted` resolves at the billing root via `resolve_billing_root`, so a reseller child answers exactly the question its root would.

## The Phase 10 dead edge, closed

Phase 10 left `LEGAL_BILLING_STATE_TRANSITIONS`'s `(ACTIVE, GRACE)` / `(FREE, GRACE)` edges with two justifications each — payment failure (driven) and "downgrade leaves org over limit" (undriven, marked `TODO(phase-11)`). `_schedule_downgrade` stamped `grace_period_ends_at` but never touched `billing_state`, so `process_dunning`'s sweep (which only looks at `GRACE`/`RESTRICTED` rows) never saw a downgrade-induced grace window and it never expired.

This phase adds the driver: `_schedule_downgrade` now calls `transition_billing_state(subscription, BillingState.GRACE)` in the same transaction. `DunningService` tells a downgrade-originated episode apart from a payment-failure one via `_is_downgrade_grace` (`pending_plan_id is not None`) — skipping the charge-retry ladder (there's no charge to retry) and, at expiry, resolving against the just-applied lower-plan limits (`_fits_under_current_limits`) rather than the catalog `free` plan, restoring `ACTIVE` (not `FREE`) since the org remains a paying subscriber of its still-active plan. A `RESTRICTED`/`CANCELLED` subscription has no `GRACE` edge from those states; `_schedule_downgrade` catches `IllegalBillingStateTransitionError` and logs rather than aborting the downgrade — the limit/pending-plan stamps still apply, `billing_state` is left as-is.

## Review

Reviewer found **no BLOCKER**. SHOULD-FIX, all applied:

- `GoogleCalendarServiceAccount` CRUD was an `OrganizationModel` write the phase had left unguarded — now calls `check_not_restricted` at both the org-view upsert path and the dedicated `ServiceAccountViewSet`.
- A derivation test now binds the restricted-write probe set to `LimitedResource.values` plus the guarded update/delete methods, replacing a hand-maintained dict that was missing 2 members (a test-coverage gap, not a code hole).
- A reseller **sync-pause** cascade test now proves a restricted root pauses a child's sync — previously only the write-block cascade had that coverage.

NIT (applied): the group service now honors the bypass flag. NIT (left as comment): `check_limit`/`check_postpaid_allowance`'s inline `billing_state == RESTRICTED` copies are the same semantic predicate against already-resolved state, not an independently-derived one — docstrings corrected to say so explicitly rather than read as a second, possibly-diverging definition.

## Gates

Re-run after the fixer's SHOULD-FIX round, credentials cleared to match CI:

- Full suite **4564 passed / 0 failed** (implementer's own run: 4538; an independent run surfaced 9 failures that all passed in isolation at 2-3x runtime — contention phantoms, not a real regression, confirmed clean by the fixer's run).
- `mypy` **305 errors / 57 files** — same ceiling as Phase 10, zero new.
- `ruff` clean (514 files).
- `makemigrations --check` clean; `schema.yml` unchanged (no model change, no migration).

## Carried forward (needs owners)

Both judgment calls below are **inert today** — every organization is on `unlimited`, so no voluntary downgrade-over-limit and no real renewal charge happens yet. No code changed for either; flagging so they aren't lost before real plans ship (Phase 14).

- **Deferral A — `_is_downgrade_grace` conflates two grace reasons.** It infers "this grace episode is downgrade-originated" from `pending_plan_id is not None`. A genuine renewal-charge failure landing while a downgrade is already pending reads as downgrade-origin too, and its retry gets silently skipped — a real money-collection miss once real charges exist. The clean fix is a dedicated `Subscription` grace-reason field rather than inferring intent from `pending_plan_id`. Gate to Phase 14 or a dedicated follow-up.
- **Deferral B — voluntary downgrade-over-limit escalates to `RESTRICTED` with payment-failure copy.** The new downgrade→`GRACE` driver means a healthy, paying customer who voluntarily downgrades but still holds more resources than the new plan allows can ride the same path into `RESTRICTED` — fully write-blocked and sync-paused, with the `RESTRICTED`/dunning notification copy (`dunning_restricted.body.html` and the grace emails) reading as "your payment failed," which is wrong for this cohort. **Needs product sign-off** on whether voluntary downgrade-over-limit should reach `RESTRICTED` at all; if so, the notification copy needs to branch on grace reason (ties into Deferral A's reason field).


payments/services/entitlement_service.py
```diff
@@ -402,6 +402,66 @@ def lock_billing_root(self, organization: Organization) -> None:
...
+        A missing subscription reads as **not restricted** (``False``), never
+        restricted -- ``billing_state`` only exists on a real row, and an
+        organization with no billing set up at all (a broken invariant, not a
+        restricted one) must not be caught by this check; that would conflate "we
+        don't know" with "we know, and the answer is blocked", which the fail-open
+        convention the rest of this service follows forbids.
+        """
+        root = resolve_billing_root(organization)
+        subscription = self._get_subscription_for_root(root)
+        return subscription is not None and subscription.billing_state == BillingState.RESTRICTED
```
hugobessa, line 449: **The one predicate.** `is_billing_root_restricted` is the single semantic definition of "must this organization's writes be blocked and its calendar sync paused?" — both halves of this phase (the write guard's `check_not_restricted` call sites and every calendar-sync-pause site: `calendar_sync_tasks`, `CalendarSyncService`'s `request_*` methods, `CalendarWebhookService`) consult it, directly or via the deliberate hot-path copy in `check_limit`/`check_postpaid_allowance` below.

Resolved at the billing root (`resolve_billing_root`), so the reseller cascade is automatic — a child organization answers exactly the question its root would, with no separate cascade logic anywhere.

Two properties worth flagging explicitly: `GRACE` is **not** restricted (only `RESTRICTED` blocks; escalation is the dunning ladder, never a write/sync block — line 435), and a missing subscription reads as **not restricted** (fail-open, line 440) rather than caught by this check — a broken invariant (no billing set up) must not be conflated with "we know, and the answer is blocked."

payments/services/entitlement_service.py
```diff
@@ -402,6 +402,66 @@ def lock_billing_root(self, organization: Organization) -> None:
...
+
+        The entry point every guarded create/update/delete method that does not
+        already route through ``check_limit`` / ``check_postpaid_allowance``
+        (which fold ``is_billing_root_restricted`` in directly, see their
+        docstrings) calls before writing an ``OrganizationModel`` row on a guarded
+        resource. See ``is_billing_root_restricted`` for what "restricted" means
+        and why it is defined exactly once.
+        """
+        if self.is_billing_root_restricted(organization):
+            raise OverLimitError.from_restricted_organization()
```
hugobessa, line 463: `check_not_restricted` is the write-guard half's entry point — every guarded create/update/delete method that doesn't already route through `check_limit`/`check_postpaid_allowance` calls this before writing an `OrganizationModel` row. It's a thin raise wrapper over `is_billing_root_restricted` (`OverLimitError`, `remedy=resolve_billing`) — no separate logic, no second predicate.

Notably, `SubscriptionService` (the `/billing/` module) calls neither this nor `is_billing_root_restricted` anywhere — that's not an oversight, it's the exemption-by-construction that lets a restricted organization always pay its way out. There's no explicit "billing paths are exempt" branch to audit; the guard simply was never wired into that service.

payments/services/subscription_service.py
```diff
@@ -616,6 +642,17 @@ def _schedule_downgrade(
...
+                transition_billing_state(subscription, BillingState.GRACE)
+            except IllegalBillingStateTransitionError:
+                logger.warning(
+                    "_schedule_downgrade: Subscription %s requested a downgrade while "
+                    "billing_state=%s, which has no GRACE edge on the billing lifecycle "
+                    "diagram. grace_period_ends_at/pending_plan are stamped above "
+                    "regardless; billing_state is left unchanged.",
+                    subscription.pk,
+                    subscription.billing_state,
+                )
```
hugobessa, line 655: **The downgrade→GRACE driver — closes the Phase 10 dead edge.** Before this, `_schedule_downgrade` stamped `grace_period_ends_at` but left `billing_state` untouched, so `process_dunning`'s GRACE/RESTRICTED sweep never looked at the row and the stamped deadline never expired — a downgrade that left an org over its new limits could sit in an unenforced "grace window" forever.

Now it routes through `transition_billing_state(subscription, BillingState.GRACE)` in the same transaction as the limit/pending-plan stamp, putting it on the one path the sweep already watches. `DunningService._is_downgrade_grace` (keyed on `pending_plan_id is not None`) tells this apart from a payment-failure grace episode downstream — see the two deferrals in the description for the rough edges that inference leaves.

Line 645-655: `RESTRICTED`/`CANCELLED` have no `GRACE` edge on the diagram from those states. Rather than abort an otherwise-valid downgrade request, `IllegalBillingStateTransitionError` is caught and logged — the limits below still apply, `billing_state` is left as it was. Escalating *out of* `RESTRICTED` is deliberately not this method's job.

payments/services/dunning_service.py
```diff
@@ -190,13 +190,58 @@ def resolve_payment_success(self, subscription: Subscription) -> Subscription:
...
+
+        organization_ids = self.entitlement_service.get_pooled_organization_ids(
+            subscription.organization
+        )
+        transaction.on_commit(
+            lambda: [
+                resync_organization_calendars_task.delay(organization_id=organization_id)
+                for organization_id in organization_ids
+            ]
+        )
```
hugobessa, line 243: **Resync-on-recovery's fan-out.** Called only when the *prior* state was `RESTRICTED` (both call sites — `resolve_payment_success` line 199-200, `check_free_fallback` further down — guard on `was_restricted`); leaving `GRACE` has nothing to reconcile since sync was never paused for `GRACE`.

Fans out per **pooled organization** (`EntitlementService.get_pooled_organization_ids`) rather than per calendar directly — the same set every usage counter and the sync-pause guard resolve against — and wraps the `.delay()` calls in `transaction.on_commit` (line 238-243) so the fan-out cannot race the transaction that moved `billing_state` off `RESTRICTED`: queuing before commit could let a worker pick this up and find the subscription still `RESTRICTED` from an uncommitted read's point of view.

calendar_integration/tasks/calendar_sync_tasks.py
```diff
@@ -187,3 +239,92 @@ def import_organization_calendar_resources_task(
...
+
+    A missing organization or a still-restricted one (a redelivery racing a
+    further billing-state change) is a no-op, mirroring every other task in this
+    module.
+    """
+    organization = Organization.objects.filter(id=organization_id).first()
+    if not organization:
+        return
+    if _restricted_or_skip(entitlement_service, organization):
+        return
```
hugobessa, line 298: **`GRACE`-not-blocked and missing-subscription-fail-open, at the task layer.** `resync_organization_calendars_task` (and every other task in this module, via `_restricted_or_skip` above) consults `is_billing_root_restricted` — the same predicate defined once in `entitlement_service.py` — as defense in depth: the `request_*` methods already refuse to *enqueue* a restricted organization's sync work, so this early-return only matters for work already queued before an organization became restricted mid-flight, or the reverse case handled here (a redelivery racing a further billing-state change reads as still-restricted and no-ops, line 290-292).

The 30-day lookback / 365-day lookahead window (line 244-252) is a deliberate approximation — `Subscription` has no `restricted_at` timestamp, so there's no way to reconstruct the exact pause window; this is sized to cover a typical grace-plus-restricted episode and matches `OrganizationService.request_rooms_sync`'s existing forward default so a recovery resync isn't narrower than an ordinary manual one.

payments/services/entitlement_service.py
```diff
@@ -470,6 +530,30 @@ def check_limit(
...
+        # ``_resolve_remedy_for`` would otherwise have picked (below, unreached
+        # for a RESTRICTED subscription now that this short-circuit exists).
+        if subscription is not None and subscription.billing_state == BillingState.RESTRICTED:
+            return LimitCheckResult(
+                allowed=False,
+                resource_key=resource_key,
+                current_usage=0,
+                ceiling=0,
+                remedy=LimitRemedy.RESOLVE_BILLING,
+            )
```
hugobessa, line 555: Hot-path copy of `is_billing_root_restricted`, not a second predicate: `check_limit` inlines `subscription.billing_state == BillingState.RESTRICTED` against the `root`/`subscription` it has already resolved, rather than calling back into `is_billing_root_restricted` (which would re-walk the `parent` chain and re-fetch the subscription on the two hottest create paths in the product). Same test, same already-resolved data — reviewer confirmed this cannot diverge from the canonical definition, and the docstring above was corrected to say so explicitly rather than read as an independent re-derivation.

PR #203, opened 2026-07-21
feat(payments): add usage API and approaching-limit warnings (billing plans and limits, phase 12)

hugobessa:
Phase 12 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md). Built on Phase 11 (`plan/billing-plans-and-limits/phase-11`, [#202](https://github.com/vintasoftware/vinta-schedule-api/pull/202)), **still unmerged** — this PR targets `plan/billing-plans-and-limits/phase-11`, not `main`. `git log --oneline plan/billing-plans-and-limits/phase-11..plan/billing-plans-and-limits/phase-12` shows only this phase's own commits.

## What this ships

Phase 11 made `RESTRICTED` actually block writes and pause sync. This phase is spec Use-case 8, both halves: an organization can see where it stands before it gets there, and is warned before it is blocked rather than only by being blocked.

- **`GET /billing/usage/`** (`payments/billing_views.py`'s `BillingUsageViewSet`) — the pull half. For every `LimitedResource` member, reports `current_usage` against the resource's effective limit plus the billing root's `billing_state`. Resolved at the billing root (`resolve_billing_root`), same as every other read in this app, so a reseller child reports the pooled root's figures, not its own. No permission beyond `IsAuthenticated` — reads never block, including for a `RESTRICTED` organization (Phase 11 blocks writes and pauses sync, never reads; an organization must be able to see exactly what it needs to resolve before it can act on it).
- **The anti-drift property — this phase's whole point.** The usage number comes from exactly one place: `EntitlementService.get_effective_limit` / `get_current_usage` (and, for `event_occurrences`, the same `MeteredOccurrence` count the meter itself produces). There is no second, hand-rolled counting path anywhere in this phase. `payments/tests/views/test_usage_view.py` makes this machine-checked rather than a design intent that could silently rot: its resource set is derived from `LimitedResource.values` itself (not a hand-typed subset, so a newly added `LimitedResource` member fails the test until a seeder is registered for it), and for every member it compares the API's reported figure against the actual enforcement primitive — `EntitlementService.check_limit` for prepaid resources, `check_postpaid_allowance` for `event_occurrences` — never against a second independent count. `payments/tests/test_usage_serialization.py` pins the two easy-to-get-wrong serialization rules: an unlimited resource serializes as `null` (never `0`, which would read as "fully consumed" — the opposite of unlimited), and a reseller child's response carries the root's pooled totals.
- **`check_approaching_limits` / `check_approaching_limits_for_subscription`** (`payments/tasks.py`) — the push half, a new Celery beat entry point (`vinta_schedule_api/celerybeat_schedule.py`) fanning out one tick per subscription (excluding `RESTRICTED` — already blocked, a further warning adds nothing — and `CANCELLED` — running out the clock to `FREE`, not accruing toward a block) to `payments/services/usage_warning_service.py`'s `UsageWarningService.check_subscription`.
- `UsageWarningService` warns at 80% (`APPROACHING_LIMIT_THRESHOLD`) of the identical effective limit `check_limit`/`check_postpaid_allowance` enforce against, and again at 100% (`LimitWarningLevel.REACHED`) — "approaching" and "reached" are two independently debounced markers, so both crossings in the same cycle each notify exactly once rather than the second being swallowed by the first.
- **`LimitWarningNotification`** (`payments/models.py`, migration `0015`) — the durable debounce ledger, unique on `(subscription, resource_key, billing_period_start, level)`. `billing_period_start` is `current_billing_period_start` — the same anchor the `event_occurrences` usage counter and the meter both use — so the debounce resets on cycle rollover rather than being permanently silenced by a marker from a prior cycle. A durable DB row, not an in-memory flag, is what survives a beat task being retried on a different worker.
- In-app notification plumbing: `payments/notification_contexts.py` (`approaching_limit_context` / `limit_reached_context`), two new templates under `templates/payments/in_app/`, and the `di_core/containers.py` wiring for `UsageWarningService`.

## The fixer's round (`a8fb1c5`)

Reviewer found no BLOCKER; 3 SHOULD-FIX + 1 NIT, all applied:

- **Debounce claim reordered to claim-and-send-in-one-transaction.** The original order marked the debounce row, then sent the notification outside that transaction — a transient send failure (network blip, provider outage) still left the marker claimed, silently debouncing that resource's warning for the rest of the billing cycle even though nothing was ever delivered. Now `_check_resource` claims the marker and calls `_notify` inside the same `transaction.atomic()` block; a `_notify` exception rolls the transaction back and un-claims the marker, so the next beat tick within the same cycle retries. The unique constraint still closes out the same race between two concurrent beat ticks — a concurrent `mark_if_new` blocks on the pending row until this transaction commits or rolls back, and only the tick whose transaction actually commits sends.
- The `RESTRICTED`-organization usage test in `test_usage_serialization.py` was hardened to seed and assert real, non-zero usage/limit values, rather than accepting whatever `check_limit`/`check_postpaid_allowance`'s `0/0` block-decision sentinel would have produced — pinning that the usage view reports the organization's true figures, never that sentinel.
- The in-app templates were rendering the raw `resource_key` (e.g. `resource_calendars`); `notification_contexts.py` now resolves `LimitedResource(resource_key).label` for a human-readable name in both templates.

## Gates

Re-run after the fixer's round:

- Full suite **4595 passed / 0 failed**.
- `mypy` **305 errors / 57 files** — same ceiling as Phase 11, zero new.
- `ruff` clean (519 files).
- `makemigrations --check` clean; migration `0015` reverses cleanly; `schema.yml` regenerated.


payments/tests/views/test_usage_view.py
```diff
@@ -0,0 +1,252 @@
...
+        # Every LimitedResource member appears exactly once.
+        assert resource_key in rows
+        row = rows[resource_key]
+
+        assert row["current_usage"] == enforcement_result.current_usage
+        assert row["limit_value"] == enforcement_result.ceiling
+        # And, independently, real usage was actually seeded -- a test that
+        # only proved "0 == 0" would not have caught the API and the guard
+        # computing usage two different ways.
+        assert enforcement_result.current_usage is not None and enforcement_result.current_usage > 0
```
hugobessa, line 242: **The anti-drift test's independence is what makes this phase's claim real, not just documented.** The parametrized `test_usage_view_matches_the_enforcement_primitive` (line 218) is driven by `LimitedResource.values` itself, not a hand-typed subset, and `test_every_seeder_is_registered_for_every_limited_resource` (line 207) fails loudly if a newly added `LimitedResource` member has no seeder wired into `SEEDERS` — so a new resource can't silently ship uncovered by this test.

For each member it seeds real, non-zero usage (a test that only proved `0 == 0` wouldn't catch the API and the guard computing usage two different ways) and compares the API's reported figure against `EntitlementService.check_limit` / `check_postpaid_allowance` directly — the actual guard entry points, never a second independent count.

payments/models.py
```diff
@@ -622,3 +627,59 @@ class Meta(BaseModel.Meta):
...
+    class Meta(BaseModel.Meta):
+        constraints: ClassVar = [
+            UniqueConstraint(
+                fields=["subscription", "resource_key", "billing_period_start", "level"],
+                name="uniq_limit_warning_notification",
+            )
+        ]
+
+    def __str__(self):
+        return f"{self.subscription} - {self.resource_key} - {self.level} @ {self.billing_period_start.isoformat()}"
```
hugobessa, line 685: **`LimitWarningNotification` — durable debounce, same pattern as `MeteredOccurrence`/`ProviderWebhookEvent`.** The unique constraint on `(subscription, resource_key, billing_period_start, level)` is the correctness mechanism, not the code path that claims it: `check_approaching_limits` re-checks every subscription on every beat tick, so without a durable row it would re-send the same warning every tick for as long as usage stays above the threshold. A DB row, not an in-memory flag, is what survives a beat task being retried on a different worker.

`billing_period_start` (line 655) is anchored on `current_billing_period_start` — the same function the `event_occurrences` counter and the meter both use — so the debounce resets on cycle rollover rather than staying silenced by a stale marker. `level` (line 672) keeps "approaching" and "reached" independently debounced, so both crossings in one cycle each notify once.

payments/services/usage_warning_service.py
```diff
@@ -0,0 +1,221 @@
...
+                subscription_id=subscription.pk,
+                resource_key=resource_key,
+                billing_period_start=billing_period_start,
+                level=level,
+            )
+            if not is_new:
+                return
+            self._notify(subscription, resource_key, level, current_usage, limit_value)
+
+    @staticmethod
```
hugobessa, line 150: **Claim-and-send-in-one-transaction — the fixer's change.** The original order marked the debounce row, then sent outside that transaction; a transient send failure still left the marker claimed, silently dropping that resource's warning for the rest of the billing cycle. Now `mark_if_new` (line 139-146) and `_notify` (line 148) run inside the same `transaction.atomic()`: if `_notify` raises, the rollback un-claims the marker, so the next beat tick within the same cycle retries instead of the warning being permanently swallowed.

The unique constraint still resolves the race between two *concurrent* beat ticks claiming the same marker — a concurrent `mark_if_new` blocks on the pending row until this transaction commits (success, no double-send) or rolls back (failure, and the other tick's claim then goes through and sends instead).

payments/tests/services/test_usage_warning_service.py
```diff
@@ -0,0 +1,378 @@
...
+            ).count()
+            == 1
+        )
+
+        # A further tick in the same cycle does not re-notify: the
+        # now-committed marker debounces normally.
+        with freeze_time(FREEZE_START + datetime.timedelta(hours=2)):
+            service.check_subscription(subscription)
+
+        assert flaky_notification_service.create_notification.call_count == 2
```
hugobessa, line 378: `test_notify_failure_does_not_mark_and_is_retried_on_the_next_tick` is the test the fixer's round added to pin the claim-and-send reordering: a failing send leaves no `LimitWarningNotification` row at all (not a "sent" marker with no notification behind it), a second tick within the same cycle retries and succeeds, and a third tick does not re-notify — the now-committed marker debounces normally from there on.

payments/tests/test_usage_serialization.py
```diff
@@ -0,0 +1,215 @@
...
+            )
+
+        response = auth_client.get(usage_url())
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data["billing_state"] == BillingState.RESTRICTED
+        rows = {row["resource_key"]: row for row in response.data["limits"]}
+        row = rows[LimitedResource.RESOURCE_CALENDARS]
+        assert row["limit_value"] == 5
+        assert row["current_usage"] == 3
```
hugobessa, line 215: `TestRestrictedOrganizationCanStillReadUsage` — reads stay open under `RESTRICTED`, and (the fixer's hardening) it now seeds and asserts real, non-zero usage/limit values rather than accepting whatever `check_limit`/`check_postpaid_allowance`'s `0/0` block-decision sentinel would have produced. This pins that `GET /billing/usage/` reports the organization's true usage, never routed through the sentinel those guards deliberately return for a `RESTRICTED` subscription's own block decision.

payments/billing_views.py
```diff
@@ -83,7 +83,22 @@ def list(self, request, *args, **kwargs):
...
+    ``payments.services.usage_warning_service.UsageWarningService`` (the
+    "push" half -- proactive approaching-limit notifications) also reads its
+    ceiling from -- so this endpoint, the enforcement guards, and the beat
+    warning can never disagree about a number.
+
+    No permission beyond ``IsAuthenticated``, deliberately -- a read never
+    blocks, including for a ``RESTRICTED`` organization (Phase 11 blocks
+    writes and pauses sync, never reads; an organization must be able to see
+    exactly what it needs to resolve before it can act on it).
+    """
```
hugobessa, line 101: **The usage view reuses the enforcement counters — it does not recompute them.** `BillingUsageViewSet.retrieve_usage` calls the identical `EntitlementService.get_effective_limit` / `get_current_usage` methods `check_limit`/`check_postpaid_allowance` count against, and `UsageWarningService` (the beat-side "push" half) also reads its ceiling from. There is deliberately no second "what is this org using" computation anywhere in this phase — see the anti-drift test comment below for how that's made machine-checked rather than just a docstring claim.

Line 97-100: reads stay open under `RESTRICTED` by construction — this viewset carries no permission beyond `IsAuthenticated`. Phase 11 blocks writes and pauses sync, never reads; an organization already blocked needs this endpoint working to see what it needs to resolve before it can act.

PR #204, opened 2026-07-21
feat(payments): close billing cycles with overage charge and reconciliation (billing plans and limits, phase 13)

hugobessa:
Phase 13 of the [billing plans and limits plan](../../../ai-plans/2026-07-18-BILLING_PLANS_AND_LIMITS_IMPLEMENTATION_PLAN.md) — the **final executable phase**; the plan is code-complete through this phase (Phase 14, rolling organizations onto real plans, is deferred by user decision, gated on product's real limit values). Built on Phase 12 (`plan/billing-plans-and-limits/phase-12`, [#203](https://github.com/vintasoftware/vinta-schedule-api/pull/203)), **still unmerged** — this PR targets `plan/billing-plans-and-limits/phase-12`, not `main`. `git log --oneline plan/billing-plans-and-limits/phase-12..plan/billing-plans-and-limits/phase-13` shows only this phase's own commits.

## What this ships

Every earlier phase metered usage, gated writes, and warned before a limit — nothing ever settled a period or moved money. This phase closes that loop: `close_billing_periods` (`payments/tasks.py`), a new Celery beat entry point, fans out one `close_subscription_billing_period` task per billing-root subscription whose current period has ended, and `CycleCloseService.close_subscription` (`payments/services/cycle_close_service.py`) does the settling — **reconcile → charge overage → roll the period → counter reset (via the roll)** — for every elapsed period, one at a time, up to `MAX_CLOSE_PERIODS_PER_RUN` (24) periods per run.

- **Exactly-once by construction.** The whole reconcile → charge → roll for one period runs inside a single `transaction.atomic()` under `SELECT ... FOR UPDATE` on the subscription row (`close_subscription`, `cycle_close_service.py:148-178`). The overage charge's idempotency key is `overage-{subscription.pk}-{period_start.isoformat()}` (`overage_idempotency_key`, `cycle_close_service.py:82-96`), derived from the **pre-roll** `current_period_start` so a retry of the same period reproduces the identical key. The durable "this period is closed" marker is the rolled `current_period_start`/`current_period_end` itself — no new model, no migration — so a re-run's `current_period_end <= now` guard is a no-op once rolled. Two concurrent sweeps of the same subscription serialize on the row lock (one thread closes, the other re-reads the rolled period and does nothing); a crash between the charge and the roll unwinds the transaction and retries with the same key, and the provider dedupes the resend. Best-effort across subscriptions: one subscription's close failing is caught and logged in `close_subscription_billing_period` (`tasks.py:199-236`) without aborting the rest of the sweep.
- **Same-derivation charge and audit.** `overage_total()` (`payments/querysets.py:67-80`, new) sums the stamped `unit_price` of every `MeteredOccurrence` row where `is_within_allowance=False` for a `for_billing_period(subscription_id, period_start)` queryset — the same queryset `MeteringService.reconcile_period` reads its identity set from. What gets charged and what reconciliation audits can never be two different numbers. The boundary is the period the meter stamped, not recomputed at close time — `is_within_allowance` was frozen at meter time (Phase 7) so a later limit change cannot retroactively reprice a closed period.
- **Inert real-money gate.** `_charge_overage` (`cycle_close_service.py:215-247`) short-circuits and charges nothing when the organization's effective `event_occurrences` limit is NULL (unlimited) — every organization is unlimited for the whole rollout, so this branch is taken today: the machinery runs and reconciles, but no money moves. Real-money activation is gated to Phase 14 alongside the recurrence pk-aliasing fix (an open-ended recurring series can duplicate indefinitely and inflate the metered count — turning that into a charge before the fix would bill for occurrences that never happened).
- **The two deferred period-boundary actions land here.** `_apply_pending_plan_change_if_due` (`cycle_close_service.py:301-353`) applies Phase 9's scheduled downgrade flip once its effective moment has passed. `_apply_cancelled_to_free_if_due` (`cycle_close_service.py:355-381`) moves a `CANCELLED` subscription that has run out its paid cycle to `FREE`, routed through `transition_billing_state` like every other `billing_state` write. Both only fire once at least one period has actually closed in the pass.
- **Finance reconciliation command.** `payments/management/commands/reconcile_billing_period.py` re-runs `MeteringService.reconcile_period` for a named subscription and period, read-only, for finance to point at any past period. No new write path.
- **No migration.** The durable "period closed" marker is the rolled `current_period_start`, not a new model.

## The fixer's round (`bfc2bd2`)

Reviewer found NO BLOCKER. One correctness fix and three test-gap closures, all applied:

- **`resolve_settlement_period` monthly-reconstruction bug.** The finance command was reconstructing an *annually*-billed subscription's past period with a 12-month stride when overage actually settles — and cycle close actually rolls — monthly regardless of `billing_interval` (`overage_settlement_step()`). A new shared helper, `resolve_settlement_period` (`payments/services/subscription_service.py:198-214`), reconstructs history with the monthly stride and is now used by both `reconcile_billing_period.py` and `MeteringService.reconcile_period`'s internal bounds — before the fix the command's displayed bounds and the report's own bounds could disagree for an annual plan.
- **Catch-up test now proves N distinct keys.** A subscription whose sweep missed several cycles closes them one at a time in the same run; the added assertion checks each closed period produced its own distinct idempotency key and its own charge, not one aggregated charge across periods.
- **CANCELLED final-period overage before the FREE flip.** A new test proves the last period's overage is charged (through the normal `_charge_overage` path) *before* `_apply_cancelled_to_free_if_due` flips `billing_state` to `FREE` — the flip must not skip settling money owed on the way out.
- **Real concurrency test.** `payments/tests/test_cycle_close_concurrency.py` (new), `@pytest.mark.django_db(transaction=True)`, two OS threads racing `close_subscription` on the same subscription with a `DedupingPaymentService` that models the provider's idempotency dedup. Proves the row lock, not the provider dedup, is what makes a double-close resolve to exactly one charge — 6/6 green in isolation.

## Gates

Re-run after the fixer's round:

- Full suite **4622 passed / 0 failed**.
- `mypy` **305 errors / 57 files** — same ceiling as Phase 12, zero new.
- `ruff` clean (530 files).
- `makemigrations --check` clean — no migration this phase.

## Carried forward (Phase 14 activation preconditions)

Two structural findings recorded, not fixed — both inert today (every organization is `unlimited`) and both are money-path restructurings that belong with Phase 14's activation work, alongside the tests that exercise a real charge:

- **C — the overage gate keys on the current effective limit, not the stamped rows.** `_charge_overage` short-circuits on the *current* effective `event_occurrences` limit being NULL. A mid-period move to `unlimited` between metering and close would drop legitimately-stamped overage silently (undercharge, never overcharge — direction is safe, but still wrong). **Phase 14 must re-key this gate on `overage_total() > 0` over the stamped rows**, honoring the same "stamped at meter time, never repriced later" principle the allowance boundary already follows.
- **D — catch-up settles up to 24 periods in one transaction holding the row lock across up to 24 provider round-trips.** `close_subscription`'s `while` loop wraps the entire catch-up run in one `transaction.atomic()` + `select_for_update`. If period N's charge fails, periods 1..N-1's rolls unwind too (provider dedup saves the money via the stable key, but re-reconciles/re-dedupes N-1 periods, and the row lock is held across two dozen network calls in the meantime). **Phase 14 should give each period its own `transaction.atomic()`.**
- Plus the recurrence pk-aliasing fix (Phase 7 finding) that gates real-money activation generally.

A NIT was also recorded and left inert: `_apply_cancelled_to_free_if_due` re-logs a warning every beat tick for a subscription stuck CANCELLED-adjacent but not on the `CANCELLED → FREE` edge — worth a one-shot marker or rate-limit at activation, not before.


payments/services/cycle_close_service.py
```diff
@@ -0,0 +1,421 @@
...
+        with transaction.atomic():
+            locked = Subscription.objects.select_for_update().get(pk=subscription.pk)
+            steps = 0
+            while locked.current_period_end <= now and steps < MAX_CLOSE_PERIODS_PER_RUN:
+                closed.append(self._close_one_period(locked))
+                steps += 1
+            if closed:
+                self._apply_pending_plan_change_if_due(locked, now)
+                self._apply_cancelled_to_free_if_due(locked)
+        return closed
```
hugobessa, line 178: **Exactly-once by construction — the whole close is one lock, one transaction.** `close_subscription` wraps the entire reconcile → charge → roll sequence for every elapsed period in a single `transaction.atomic()` under `SELECT ... FOR UPDATE` on the subscription row (line 170). A concurrent sweep of the *same* subscription blocks on that lock until this transaction commits, then re-reads the rolled period and finds nothing left to close — so a double-charge under a race is prevented by the lock, not by the provider's idempotency dedup (that's the separate crash-safety net for a redelivery/retry, not for a genuine race).

The `while` guard (line 172) is what makes a re-run idempotent: once a period rolls, `current_period_end` moves into the future, so the guard exits immediately on a subsequent call — no separate "period closed" record is needed. See `payments/tests/test_cycle_close_concurrency.py` for the test that proves the lock (not the dedup) is load-bearing.

payments/services/cycle_close_service.py
```diff
@@ -0,0 +1,421 @@
...
+    provider's own idempotency header (Phase 9 plumbing, via
+    ``PaymentService.create_payment`` -> ``BasePaymentAdapter.process``), so two
+    attempts to close the *same* period — a Celery redelivery, or a retry after a
+    crash between the charge and the period-roll — resolve to **one** charge at the
+    provider even when the local ``Payment`` row from the first attempt was rolled
+    back. Derived from ``period_start`` (not "now" and not a fresh uuid) precisely
+    so a re-run of the same closing period produces a byte-identical key; a per-run
+    key would double-charge on every retry.
+    """
+    return f"overage-{subscription.pk}-{period_start.isoformat()}"
```
hugobessa, line 96: **`overage_idempotency_key` — derived from the pre-roll period, not "now".** The key is `overage-{subscription.pk}-{period_start.isoformat()}`, forwarded to the provider's own idempotency header (Phase 9 plumbing). Deriving it from `period_start` rather than a fresh uuid or timestamp is what makes a retry of the *same* closing period reproduce a byte-identical key — a crash between the charge and the period roll (line 203 in `_close_one_period`) unwinds the transaction, and the next attempt calls `_charge_overage` with the same still-unrolled `period_start`, so the provider itself refuses the second charge even though the local `Payment` row from the first attempt was rolled back.

payments/services/cycle_close_service.py
```diff
@@ -0,0 +1,421 @@
...
+            subscription.organization, LimitedResource.EVENT_OCCURRENCES
+        )
+        if effective_limit.limit_value is None:
+            logger.debug(
+                "Cycle close: subscription %s has an unlimited event_occurrences allowance; "
+                "rolling the period and reconciling but charging no overage (real-money overage "
+                "is gated to Phase 14).",
+                subscription.pk,
+            )
+            return Decimal("0"), None
```
hugobessa, line 247: **The real-money gate.** `effective_limit.limit_value is None` (line 240) short-circuits before any charge is attempted — every organization is `unlimited` for the whole rollout, so this branch is what runs today: the period still reconciles and rolls, but nothing is charged. The docstring is explicit about why this isn't removed yet — the recurrence pk-aliasing defect (Phase 7 finding) can inflate the metered count, and turning an inflated count into a charge before that's fixed would bill for occurrences that never happened.

Carried forward to Phase 14 (not fixed here, since it's inert today and a money-path change): this gate keys on the *current* effective limit, not on whether the stamped rows actually carry overage. A mid-period move to unlimited between metering and close would silently drop legitimately-stamped overage — safe direction (undercharge), but Phase 14's activation should re-key this on `overage_total() > 0` over the stamped rows instead, matching the "stamped at meter time, never repriced" principle the allowance boundary itself already follows.

payments/querysets.py
```diff
@@ -52,3 +53,28 @@ def for_organizations(self, organization_ids: Sequence[int]) -> MeteredOccurrenc
...
+        **The single derivation of "how much overage does this period owe".** Cycle
+        close charges exactly this (``CycleCloseService``), and it is a sum over the
+        very same ``for_billing_period`` rows ``reconcile_period`` recomputes its
+        identity set from — so what gets charged and what reconciliation audits can
+        never be two different numbers. The total is ``sum(unit_price) where not
+        is_within_allowance``, never ``count * current_price``: the price is the one
+        stamped when each occurrence was metered, so re-pricing the plan after the
+        fact cannot change a closed period's bill.
+        """
+        return self.overage().aggregate(total=Sum("unit_price"))["total"] or Decimal("0")
```
hugobessa, line 80: **`overage_total()` — the single derivation of "how much overage does this period owe", summing stamped prices, never repricing.** `overage()` filters to `is_within_allowance=False` (stamped at meter time, Phase 7), and `overage_total()` sums those rows' `unit_price` — not `count * current_price`. The price stored on each row is the one in force when it was metered, so a plan repricing after the fact cannot change what a closed period owes.

This is also the exact queryset `reconcile_period` recomputes its identity set from (both call `for_billing_period(subscription_id, billing_period_start)`), so what cycle close charges and what reconciliation audits are structurally the same numbers, not two hand-rolled counts that are supposed to agree.

payments/services/subscription_service.py
```diff
@@ -163,6 +195,25 @@ def resolve_billing_period(
...
+    for *every* plan, and a subscription's stored period is created one month long
+    (``create_subscription_for_organization``) and rolled one month forward at close
+    (``CycleCloseService._roll_period``) regardless of ``billing_interval`` — so an
+    annually-billed subscription's past periods must be walked back monthly, not by
+    twelve-month strides. Reconstructing an annual plan's history with the plan-cycle
+    stride lands on the wrong bounds (or overshoots into
+    ``BillingPeriodResolutionError``); this is the resolution that matches how close
+    actually rolls and how the meter stamped those historical rows.
+    """
+    return resolve_billing_period(subscription, moment, step=overage_settlement_step())
```
hugobessa, line 214: **The fixer's correctness fix — `resolve_settlement_period` walks history with the monthly stride, not the plan's billing interval.** Overage settles monthly for every plan (`overage_settlement_step()`), including annually-billed ones — cycle close rolls `current_period_start` one month at a time regardless of `billing_interval`. Before this fix, the finance command reconstructed an annual subscription's past period with a 12-month stride, landing on the wrong bounds (or overshooting into `BillingPeriodResolutionError`) and disagreeing with `reconcile_period`'s own internal bounds. Both callers now share this one function, so the command's displayed period and the report's computed period can't drift apart.

payments/tasks.py
```diff
@@ -170,6 +171,71 @@ def process_dunning_for_subscription(
...
+    ``(subscription, period_start)`` idempotency key), so a
+    ``CELERY_TASK_ACKS_LATE`` redelivery is harmless.
+
+    Only billing-root subscriptions with an elapsed period are selected
+    (``CycleCloseService.subscriptions_to_close`` reuses
+    ``MeteringService.subscriptions_to_sweep`` so "which subscription owns this
+    usage" has a single definition).
+    """
+    for subscription_id in CycleCloseService.subscriptions_to_close():
+        close_subscription_billing_period.delay(subscription_id)
```
hugobessa, line 194: **Beat entry point — the window is decided once, each subscription closes independently.** `close_billing_periods` resolves "which subscriptions are due" a single time from `CycleCloseService.subscriptions_to_close()` (itself built on `MeteringService.subscriptions_to_sweep()`, so "which subscription owns this usage" has one definition shared with the meter), then fans out one `close_subscription_billing_period.delay(...)` per subscription. Each task is independently idempotent (the rolled period + the overage idempotency key), so one subscription's declined charge or provider error — caught inside `close_subscription_billing_period`, not here — never aborts the rest of the sweep, and a `CELERY_TASK_ACKS_LATE` redelivery of this task itself is harmless (re-dispatching an already-closed subscription is a no-op).

payments/tests/test_cycle_close_concurrency.py
```diff
@@ -0,0 +1,192 @@
...
+    # failed and both closed the same period.
+    assert sorted(closed_counts) == [0, 1], f"expected one closer, got {closed_counts}"
+    # The provider was asked to charge exactly once — not deduped-down from two
+    # attempts, but a single call — with one distinct key.
+    assert len(payment_service.calls) == 1
+    assert payment_service.settled_keys == {overage_idempotency_key(subscription, PERIOD_START)}
+    # The period rolled exactly one month, not two.
+    subscription.refresh_from_db()
+    assert subscription.current_period_start == PERIOD_END
+    assert subscription.current_period_end == PERIOD_END + relativedelta(months=1)
```
hugobessa, line 192: **The test that proves the lock, not the provider dedup, is what prevents a double charge.** Runs against a real database (`transaction=True`) with two OS threads on separate connections racing `close_subscription` for the same subscription; `DedupingPaymentService.create_payment` sleeps mid-call so the lock winner holds the row across the race window, forcing the two transactions to genuinely overlap rather than happening to interleave.

Self-validating: under a correct lock, exactly one thread finds a period to close (`sorted(closed_counts) == [0, 1]`, line 184) and the provider sees exactly one `create_payment` call with one distinct key (lines 186-188) — a broken lock would let both threads read the un-rolled period and both attempt the charge (`[1, 1]`, two calls), which this assertion fails on loudly rather than being silently saved by the provider's own dedup (a different failure mode this test deliberately isolates from).

PR #206, opened 2026-08-04
feat: self-serve organization slug (Phase 1 — auth branding)

hugobessa:
## Phase 1 — Self-serve organization slug

First phase of the [Organization Auth-Area Branding plan](../blob/main/ai-plans/2026-08-04-ORGANIZATION_AUTH_BRANDING_IMPLEMENTATION_PLAN.md). Lets an organization admin pick a public slug for their organization, subject to uniqueness, format, reserved-word, and confusable-character rules. The slug becomes the key for the branded login URL and `brandingForTenant` in later phases, and a precondition for configuring branding.

### What's in it
- **`Organization.slug`** — `SlugField`, unique, nullable, indexed. `default=None` so multiple unset organizations coexist as `NULL` under the unique index (rather than colliding on `""`). Migration `0018_organization_slug` (single `AddField`, no backfill).
- **`organizations/slug_validation.py`** — one shared callable `validate_organization_slug()` enforcing three rules, so the REST serializer, the admin form, and the later GraphQL input share one implementation:
  - **Reserved words** — our own live route segments (`super`, `schema`, `s3direct`, `auth`, `api`, `login`, `admin`, `app`, `dashboard`, `static`, `media`, …) and names implying the vendor (`vinta`, `vinta-schedule`, variants). Kept as data, not scattered conditionals.
  - **Format & length** — lowercase alphanumeric + internal hyphens, bounded 3–63, no leading/trailing hyphen, not purely numeric.
  - **Confusables** — rejects non-ASCII / mixed-script lookalikes (the phishing defense). No confusables library is present in the dependency tree, so this uses stdlib `unicodedata` rather than adding a license-gated dependency.
- **REST** — `slug` writable on `OrganizationSerializer` (declared as `CharField` so the shared validator is the sole authority and its rule-specific messages surface; a collision returns **400 naming the conflict**, not a 500); read-only on `OrganizationBriefSerializer`.
- **Admin** — `slug` exposed and searchable, with `slug` declared as a plain `forms.CharField` on the admin form so the same validation (and its messages) apply to staff too.
- `schema.yml` regenerated.

### Tests
Table-driven validation unit tests (every reserved word, each format sub-rule, confusable lookalikes, valid acceptance); model tests (uniqueness, multiple NULLs coexist, optional-on-create, mutable); REST endpoint tests (admin PATCH sets slug, non-admin refused, duplicate → 400 naming the collision, reserved/confusable rejected through the endpoint); admin-form tests (same rules incl. confusable + reserved through the form). Full suite: **4707 passed**.

### Review notes
Adversarial review caught that `SlugField` (DRF + admin `ModelForm`) auto-attaches an ASCII-only regex that ran *before* the custom validator — making the confusables rule dead code on both write surfaces and returning generic messages. Fixed by declaring `slug` as `CharField`/`forms.CharField` on both surfaces (model field unchanged, no extra migration), with new tests exercising a Cyrillic-homoglyph slug through both PATCH and the admin form.

### Notes for reviewers
- Slug length bounds (3–63) and the "not purely numeric" interpretation (strip hyphens, reject all-digit remainder) are judgment calls not pinned by the plan.
- The reserved-word list is a product asset (plan Open Question 3) — seeded here with live routes + vendor variants; expected to grow.

https://claude.ai/code/session_01PhB71XXpM9repoyHaWMxe2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PhB71XXpM9repoyHaWMxe2)_


organizations/slug_validation.py
```diff
@@ -0,0 +1,182 @@
...
+        "graphql",
+        "webhooks",
+        "webhook",
+        "help",
+        "support",
+        "docs",
+        "status",
+        "www",
+    }
+)
```
arthurzeras, line 99: If creating new routes in future, we need to include these new routes here, right? How can we enforce the agents to include them? Could a simple skill do this?

The vice-versa is also a possible scenario: A slug exists and we need a new route with the same name, should we need to check the current slugs before create a new route?

hugobessa: Good call — enforcement now lives in the test suite instead of a reviewer's memory. Added `organizations/tests/test_reserved_slug_routes.py`: it derives the top-level route segments from the **live** root URLconf at test time and asserts every one is in `RESERVED_ORGANIZATION_SLUGS`, naming any offender. So a future `path("reports/", …)` added without reserving `reports` fails CI automatically — no hardcoded copy of the route list to drift.

Two deliberate scope choices, documented in the test's docstring:

- **Only top-level `urls.py` mounts** (`auth/`, `api/`, `s3direct/`, `super/`, `schema/`, `graphql/`, …) are enforced — not the DRF-router resources mounted at the empty-prefix root (`organizations/`, `branding/`, `calendar-events/`, …). Those are API resource paths rather than vendor/infrastructure names, and a branded slug is served under `/login/<slug>/`, so there's no actual routing collision; reserving all ~20 of them would also over-block legitimate org slugs like `profile`. If we later decide a specific resource name shouldn't be claimable, we add it to the reserved list and this same test keeps guarding it.
- **The reverse direction** (a slug already exists, then someone adds a route of the same name) is intentionally *not* guarded here: routes are ours to name, so the reserved list is the one-way guard. The realistic remedy if it ever happens is renaming the new route or reserving+migrating, which is a conscious decision, not something to fail silently on at slug-creation time.

---
_Generated by [Claude Code](https://claude.ai/code)_

arthurzeras (approved):

PR #210, opened 2026-08-04
feat: audit branding writes and expose can_manage_branding (Phase 4 — auth branding)

hugobessa:
## Phase 4 — Audit branding writes and expose the capability field

Fifth phase of the [Organization Auth-Area Branding plan](../blob/main/ai-plans/2026-08-04-ORGANIZATION_AUTH_BRANDING_IMPLEMENTATION_PLAN.md). Every branding write is recorded with a diff, and the dashboard can tell whether to show the branding page without probing.

> **Stacked on [#209](https://github.com/vintasoftware/vinta-schedule-api/pull/209) (Phase 3).** Review this diff against that branch.

### What's in it
- **Audit** — `OrganizationBrandingView` (PUT/PATCH) and GraphQL `updateBranding` record branding **create** (no diff) and **update** (diff of changed fields only) through `AuditService.record`, actor derived per surface (`actor_from_membership` for REST, `actor_from_system_user` for the partner API), subject = the branding row. A refused write (gate or validation failure) records **nothing** — the audit call is unreachable when the write doesn't happen, and the GraphQL path defers via `transaction.on_commit` so a rolled-back atomic block never enqueues.
- **`can_manage_branding`** — read-only field on `CurrentMembershipSerializer` and `MyMembershipSerializer`, computed as **parentless-and-entitled** via the shared `is_branding_eligible_organization` helper. Deliberately **excludes the slug condition** so it exactly equals Phase 3's GET read-gate reachability: a slug-less parentless entitled org gets `true`, so the SPA renders the branding page and shows the "pick a slug first" prompt rather than hiding a page from the admins one step away.
- Shared `branding_diff_state()` snapshot helper (all six mutable fields; `logo` normalized to its stored key). No migration (audit uses the existing model; the field is computed).

### Tests
`can_manage_branding` true only for parentless+entitled (incl. slug-less), false for parented/free; audit create + update-diff (PUT and PATCH) name only changed fields; refused writes record nothing on both surfaces; membership payload carries the field across all gate cases. Full suite: **4800 passed**.

### Review notes (Tier 3 / sonnet)
No BLOCKER. Verified holding: audit diff ordering (PATCH before-state captured strictly before mutation), refused-write-records-nothing on both surfaces, `can_manage_branding` == the two-condition helper == GET-reachability, correct actor + org on each record. Two SHOULD-FIXes fixed:
- **N+1 on `GET /organizations/mine/`** — the per-row entitlement lookup is now batched via a new `EntitlementService.has_entitlement_for_organizations` bulk helper + a `ListSerializer` that resolves eligibility once for the whole list; a query-count test asserts entitlement queries don't scale with membership count (parented orgs short-circuit before any entitlement query).
- **Late imports** in `mutations.py` hoisted to module top (confirmed no import cycle).

https://claude.ai/code/session_01PhB71XXpM9repoyHaWMxe2

---
_Generated by [Claude Code](https://claude.ai/code/session_01PhB71XXpM9repoyHaWMxe2)_

### DIFF
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -1,4 +1,5 @@
 import enum
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Annotated
 
 from dependency_injector.wiring import Provide, inject
@@ -49,6 +50,28 @@ def _organization_holds_white_label_branding(
     )
 
 
+@inject
+def _organizations_hold_white_label_branding(
+    organizations: Sequence[Organization],
+    entitlement_service: Annotated[EntitlementService, Provide["entitlement_service"]] = None,  # type: ignore[assignment]
+) -> dict[int, bool]:
+    """Bulk sibling of ``_organization_holds_white_label_branding``, backing
+    ``is_branding_eligible_organizations``. Same ``@inject``/``Provide``
+    fail-closed pattern: an unresolvable entitlement service denies every
+    organization in the batch rather than admitting any of it.
+
+    Deliberately does not go through ``has_entitlement_cached`` — the point of
+    this function is to answer for the whole batch in two queries, which
+    already beats what the per-organization request memo achieves for a batch
+    of distinct organizations.
+    """
+    if entitlement_service is None or not organizations:
+        return {}
+    return entitlement_service.has_entitlement_for_organizations(
+        organizations, Entitlement.WHITE_LABEL_BRANDING
+    )
+
+
 def is_branding_eligible_organization(organization: Organization | None) -> bool:
     """Shared branding-eligibility gate: ``organization`` has no parent AND holds
     the ``white_label_branding`` entitlement.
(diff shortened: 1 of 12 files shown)


organizations/permissions.py
```diff
@@ -40,6 +41,28 @@ def _organization_holds_white_label_branding(organization: Organization) -> bool
     )
 
 
+def _organizations_hold_white_label_branding(
```
hugobessa, line 54: This function isn't using the dependency injection framework correctly. It must use the `Provide` type with `Annotated` and the `inject` decorator.

hugobessa: Fixed — converted the branding entitlement helpers from the `di_core.containers.container` service-locator to proper framework injection, matching the established pattern (`audit/services.py`, `accounts/account_adapters.py`):

```python
@inject
def _organization_holds_white_label_branding(
    organization: Organization,
    entitlement_service: Annotated[EntitlementService, Provide["entitlement_service"]] = None,  # type: ignore[assignment]
) -> bool:
    if entitlement_service is None:
        return False
    return has_entitlement_cached(entitlement_service, organization, Entitlement.WHITE_LABEL_BRANDING)
```

The `= None` default keeps it fail-closed (denies if wiring hasn't run); the `organizations` package is already wired via `container.wire(packages=INTERNAL_INSTALLED_APPS)`, so callers never pass the service and the public gates (`is_branding_eligible_organization` / `is_branding_eligible_organizations`) just delegate. The bulk sibling `_organizations_hold_white_label_branding` got the same treatment.

Applied at the **origin** of the pattern so every PR in the stack is internally correct, not just this one:
- **#208 (Phase 2b)** — `is_branding_eligible_organization`
- **#209 (Phase 3)** — `_organization_holds_white_label_branding` (the factored helper)
- **#210 (Phase 4)** — `_organizations_hold_white_label_branding` (the bulk helper)

Those three branches were amended and phases 2a/5/6/7/8/9 rebased on top, so the whole stack was force-pushed. The end-to-end tests (the S3Direct destination `auth` callable and the GraphQL logo-signing gate, which call these helpers) pass, confirming injection resolves at runtime.

---
_Generated by [Claude Code](https://claude.ai/code)_

arthurzeras (approved):

PR #218, opened 2026-08-05
feat(calendar): add group-slot scoping foundation to availability tables

hugobessa:
## Summary

Phase 0 of the Calendar Group-Scoped Availability plan. It lays the schema and default-exclude scoping foundation that every later phase consumes, with **zero behavior change anywhere** — no code writes the new column yet.

Adds a nullable `group_slot` reference (`OrganizationForeignKey`, `on_delete=CASCADE`) to the existing `AvailableTime` and `BlockedTime` tables. `NULL` means a base row (today's behavior, visible on every read path); a non-null value scopes the row to exactly one `CalendarGroupSlot`. The default managers (`objects`) now exclude group-scoped rows; the rows are reachable only through the explicit `for_group_slot(id)` / `unscoped()` accessors.

## Plan reference

`ai-plans/2026-08-05-CALENDAR_GROUP_SCOPED_AVAILABILITY_IMPLEMENTATION_PLAN.md` — Phase 0.

## Key decisions (not obvious from the diff)

- **Storage shape** (Guiding Decisions → "Storage shape"): reuse the existing recurring `AvailableTime`/`BlockedTime` models rather than new models, so the four versioned occurrence SQL functions, exception model, bulk-modification model, and queryset are reused unchanged.
- **Row scoping** (Guiding Decisions → "Row scoping"): the *default* manager excludes group-scoped rows. Chosen over per-call-site filtering because a missed call site is a silent correctness failure. Default-exclude means every existing call site is correct with zero edits — confirmed by the full existing suite passing untouched.
- **Cascade correctness**: the FK constraints are created `DEFERRABLE INITIALLY DEFERRED` (matching Django's native Postgres FK behavior). Django's deletion collector skips ordering deletes for nullable CASCADE relations, assuming the FK check is deferred to COMMIT; a non-deferrable constraint would make `CalendarGroupSlot` deletion intermittently raise a foreign-key violation instead of cascading. Cascade runs through `Model._base_manager` (unfiltered), so it reaches group-scoped rows even though `objects` hides them.
- **`only_user_authored` stays scope-agnostic** at the queryset level, so the billing counter can count group-scoped windows (metered too) via `unscoped()` in later phases.
- **SQL-function assumption verified** (Open Question 1): the occurrence functions take a row id as input and select no rows themselves, so scoping is caller-side and **no function version bumps were needed**.

## Migration safety

`0042` is lock-aware on these hot tables: metadata-only `ADD COLUMN ... NULL` (no rewrite), FK added `NOT VALID` then `VALIDATE` in a separate weak-lock step, partial indexes built `CONCURRENTLY` (`atomic = False`). Reverse path drops indexes → constraints → columns cleanly. No backfill.

## Feature flag

None — the feature is self-gating. An unconfigured calendar takes the fall-through path and behaves byte-for-byte as today. This phase is unreachable from any existing flow because no code writes the new column yet.

## Test plan

```bash
docker compose run --rm api uv run python manage.py makemigrations --check
docker compose run --rm api uv run pytest calendar_integration/tests/test_models.py calendar_integration/tests/services/test_group_slot_scoping_phase0.py -q
```

### DIFF
--- a/calendar_integration/models.py
+++ b/calendar_integration/models.py
@@ -1348,8 +1348,36 @@ class BlockedTime(RecurringMixin):
         help_text="If this is a continuation of a split series",
     )
 
+    # Group-scoped availability (Phase 0 of CALENDAR_GROUP_SCOPED_AVAILABILITY):
+    # NULL means a base row — today's behavior, visible on every read path. A
+    # non-null value scopes the row to that one CalendarGroupSlot; it is invisible
+    # to the default manager (`objects`) and only reachable through the explicit
+    # `for_group_slot` / `unscoped` accessors. `on_delete=CASCADE` so removing a
+    # calendar from a slot (or deleting the slot/group) deletes its group-scoped
+    # blocked time with it, matching the spec's cascade rule.
+    group_slot = OrganizationForeignKey(
+        CalendarGroupSlot,
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="group_scoped_blocked_times",
+        help_text=(
+            "If set, this blocked time applies only when the calendar is "
+            "evaluated inside this group slot, and nowhere else. Null (the "
+            "default) means a base row that blocks time everywhere the "
+            "calendar is evaluated."
+        ),
+    )
+
     class Meta:
         unique_together = (("calendar_fk_id", "external_id"),)
+        indexes = (
+            models.Index(
+                fields=["organization", "group_slot_fk"],
+                condition=models.Q(group_slot_fk__isnull=False),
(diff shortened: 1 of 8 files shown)


calendar_integration/models.py
```diff
@@ -1422,6 +1450,36 @@ class AvailableTime(RecurringMixin):
         help_text="If this is a continuation of a split series",
     )
 
+    # Group-scoped availability (Phase 0 of CALENDAR_GROUP_SCOPED_AVAILABILITY):
```
hugobessa, line 1453: Do not make references to ai plans from comments in the codebase

PR #220, opened 2026-08-05
feat(calendar): add group-scoped availability window write service

hugobessa:
## Summary

Phase 1a of the Calendar Group-Scoped Availability plan. Adds the service-layer create/update/delete for group-scoped availability windows — audited, permission-gated, and reporting the bookings a narrowing orphans. Reachable only through new service methods; no existing caller enters this code.

## Plan reference

`ai-plans/2026-08-05-CALENDAR_GROUP_SCOPED_AVAILABILITY_IMPLEMENTATION_PLAN.md` — Phase 1a. Spec use-cases UC-1 (admin narrows a surgeon to operating days) and UC-6 (admin tightens a window that orphans bookings).

## Key decisions

- **A group-scoped window is an `AvailableTime` row with `group_slot` set** (Guiding Decisions → "Storage shape"), reusing the existing recurrence machinery (recurrence rules, per-window timezone, occurrence expansion). All writes/reads go through the explicit `AvailableTime.objects.for_group_slot(...)` / `.unscoped()` accessors from Phase 0 — never the default `objects` manager, which excludes group-scoped rows.
- **Occurrence-expansion safety**: `_group_scoped_available_times_expanded` annotates occurrences (`annotate_recurring_occurrences_on_date_range`) BEFORE calling `get_occurrences_in_range()`, so the instance's cached `recurring_occurrences` short-circuits `RecurringMixin`'s internal re-fetch through the default (base-rows-only) manager — which would otherwise silently return nothing for a group-scoped master (the Phase 0 carry-forward note).
- **Non-disclosure permissions** (spec Permissions decision): a stranger, an owner of a *different* calendar in the same group, and a genuinely-missing membership all raise the identical `CalendarGroupSlotConfigNotFoundError` (same type + message), so no error shape reveals a group the caller cannot see. Org admins may manage anyone's; resource calendars with no owner are admin-only.
- **Audit** (Objective 4): create/update/delete each emit an audit record; update captures a before/after diff over times, timezone, and the rrule string.
- **Orphaned-booking detection** (UC-6): a narrowing update — and creating the *first* window for a (calendar, slot), which flips the slot from fall-through to narrowed evaluation — collects confirmed future bookings now outside the configuration and returns them in `GroupScopedAvailabilityWriteResult`. Nothing is cancelled or modified. Adding a later window only widens the union and cannot orphan.
- **Slot-membership removal cleanup**: removing a calendar from a slot while the slot survives deletes only the membership row; the `group_slot` FK cascades on *slot* deletion, not membership removal. So `_reconcile_slot` now explicitly deletes that calendar's group-scoped windows on membership removal — audited per window — satisfying the spec's "calendar removed from a slot → windows deleted" edge case. (Phases 2a/3a must extend the same cleanup for blocks and quota rules — noted in a code comment.)

## Feature flag

None — self-gating; reachable only through new service methods, no existing caller enters this code, no migration.

## Test plan

```bash
docker compose run --rm api uv run pytest calendar_integration/tests/services/test_calendar_group_service_availability_windows.py -vs
```

Covers: CRUD; recurrence + per-window timezone round-trip; audit records with diffs; both permission-denial cases without disclosing the group; orphan detection on update and first-window create (returns them, mutates nothing); window deletion on both whole-slot deletion and calendar-removed-from-slot (with audit).

### DIFF
--- a/calendar_integration/services/calendar_group_service.py
+++ b/calendar_integration/services/calendar_group_service.py
@@ -436,6 +436,21 @@ class CalendarGroupService:
 
         if to_remove:
             self._ensure_no_future_selections(slot=slot, calendar_ids=to_remove)
+
+            # Delete group-scoped windows for the removed calendars.
+            # The FK on AvailableTime.group_slot → CalendarGroupSlot cascades on
+            # SLOT deletion only, not on membership removal, so we must explicitly
+            # clean up the orphaned rows here. These deletions are cascading from
+            # membership removal (already audited as part of the group update), so
+            # we do not audit them individually.
+            # TODO(Phase 2a, 3a): BlockedTime.for_group_slot() and quota rules
+            # must extend this cleanup when those phases add their group-scoped rows,
+            # using the same pattern: delete rows for removed calendars in to_remove.
+            org_id = cast(Organization, self.organization).id
+            AvailableTime.objects.unscoped().filter_by_organization(org_id).filter(
+                group_slot_fk=slot, calendar_fk_id__in=to_remove
+            ).delete()
+
             CalendarGroupSlotMembership.objects.filter_by_organization(self.organization.id).filter(
                 slot_fk=slot, calendar_fk_id__in=to_remove
             ).delete()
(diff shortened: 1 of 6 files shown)


calendar_integration/services/calendar_group_service.py
```diff
@@ -64,6 +69,36 @@
     from payments.services.entitlement_service import EntitlementService
 
 
+def _time_range_fully_covered(
```
hugobessa, line 72: I'm a bit concerned with the performance here. Shouldn't this be a database check instead?

PR #221, opened 2026-08-05
feat(calendar): intersect group-scoped windows into discovery and booking validation

hugobessa:
## Summary

Phase 1b of the Calendar Group-Scoped Availability plan — the phase that proves Goal 3 (zero change for unconfigured groups). A calendar with group-scoped windows is now offered only inside them, in that group only, and a booking or reschedule that names it outside them is rejected.

## Plan reference

`ai-plans/2026-08-05-CALENDAR_GROUP_SCOPED_AVAILABILITY_IMPLEMENTATION_PLAN.md` — Phase 1b. Spec use-cases UC-1, UC-4, and the discovery half of UC-6; Acceptance scenarios 1–4.

## Key decisions

- **Zero-change early-out (Goal 3)**: the "does any calendar in this group have a group-scoped window" check is an `Exists()` annotation folded into the `CalendarGroupSlotMembership` query discovery ALREADY issues — no new round trip. When empty, `fetch_group_scoped_available_spans` is skipped entirely. Verified: `find_bookable_slots` 6→6 queries, `check_group_availability` 5→5, byte-identical output — asserted permanently in `test_unconfigured_group_discovery_is_byte_for_byte_unchanged` (the flag-off substitute).
- **Intersect-only, never widen** (Guiding Decisions): every path computes `base_free` first and short-circuits `False` before consulting any group-scoped span, so a window can only narrow an already-free result. A window outside base availability yields no offered time (Acceptance 3).
- **Configured-but-no-overlap**: a calendar WITH windows but none covering time T is not bookable at T (empty overlap ≠ fall-through). Distinct from a calendar with NO windows, which falls through to base availability.
- **Booking + reschedule rejection**: `_assert_calendars_within_group_scoped_windows` runs in `create_grouped_event` and `reschedule_grouped_event`. Reschedule enforcement was extended to ALL selected calendars (not just primary) so every enforcement surface agrees (spec Objective 3). Error `CalendarGroupScopedRuleViolationError(calendar_id, rule_type)` names the calendar and rule type only — never configured window times (spec Errors decision).
- **Single-calendar path untouched**: `bookable_slots_service.py` is zero-diff (Non-goal).
- **Recurrence-exception safety (permanently closed)**: `RecurringMixin._get_occurrences_in_range`'s exception-instance lookup now uses `_base_manager` when the master is group-scoped (guarded — all other recurring models unchanged), so a group-scoped recurrence exception can never be silently missed. This closes a trap carried forward from Phase 0/1a.

## Feature flag

None — self-gating via the early-out; no migration.

## Test plan

```bash
docker compose run --rm api uv run pytest calendar_integration/tests/services/test_group_scoped_availability_windows_discovery.py -vs
```

Covers: surgeon scenario end to end (offered only Tue/Thu in Surgery, full base availability in a second group); window outside base availability offers nothing; explicit booking + reschedule rejection (primary and non-primary calendars) with the leak-free error shape; positive reschedule into a valid window; unconfigured-group query-count + output identity; a group-scoped recurring master's exception is honored.

### DIFF
--- a/calendar_integration/services/slot_engine.py
+++ b/calendar_integration/services/slot_engine.py
@@ -331,18 +331,12 @@ def _iter_group_scoped_available_time_occurrences(
     configured.
 
     Reads through ``AvailableTime.objects.unscoped()`` -- group-scoped rows are
-    invisible to the default manager. Mirrors
-    ``CalendarGroupService._group_scoped_available_times_expanded``'s
-    annotate-first strategy (that method now delegates here): masters are
-    fetched WITH ``annotate_recurring_occurrences_on_date_range`` already
-    attached, so ``get_occurrences_in_range()`` finds ``recurring_occurrences``
-    cached on the instance and never falls through to ``RecurringMixin``'s
-    internal exception-instance re-fetch, which goes through the DEFAULT
-    (base-rows-only) manager and would otherwise silently drop a group-scoped
-    master's exceptions (see the Phase 0 carry-forward note). No group-scoped
-    window can carry an exception yet as of Phase 1b -- there is no write path
-    that creates one -- so that re-fetch is never actually exercised, but the
-    annotate-first shape is what keeps this safe if/when one becomes reachable.
+    invisible to the default manager. Occurrence expansion for group-scoped
+    masters is safe because (a) no write path creates a group-scoped recurrence
+    exception yet, and (b) ``RecurringMixin._get_occurrences_in_range`` now
+    routes the exception-instance lookup through ``_base_manager`` when the
+    master is group-scoped, ensuring group-scoped exception rows are found if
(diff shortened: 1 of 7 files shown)


calendar_integration/services/slot_engine.py
```diff
@@ -262,3 +309,132 @@ def apply_policy_filter(
...
+        .select_related("recurrence_rule")
+    )
+
+    non_recurring_times = base_qs.filter(
+        start_time__lt=end_date,
+        end_time__gt=start_date,
+        recurrence_rule__isnull=True,
+        is_recurring_exception=False,
+    )
+    for at in non_recurring_times:
```
hugobessa, line 371: shouldn't we use `non_recurring_times.iterator()` here?

PR #224, opened 2026-08-05
feat(calendar): add group-scoped blocked time writes and enforcement

hugobessa:
## Summary

Phase 2a of the Calendar Group-Scoped Availability plan. A member can now block time within one group without affecting any other group, and the block wins over any window. Blocks are the negative analog of windows — `BlockedTime` rows with `group_slot` set — reusing the same recurrence machinery, audit, permissions, and early-out discipline.

## Plan reference

`ai-plans/2026-08-05-CALENDAR_GROUP_SCOPED_AVAILABILITY_IMPLEMENTATION_PLAN.md` — Phase 2a. Spec use-case UC-3.

## Key decisions

- **Blocks beat everything** (spec resolution order): on all three enforcement paths (`slot_engine.calendar_free_for_window`, `check_group_availability`'s per-range loop, `_assert_calendars_within_group_scoped_windows`) a group-scoped block that overlaps the requested time excludes/rejects the calendar BEFORE the window-coverage check runs. A calendar with a full-day window but a mid-day block is not offered mid-day (verified in discovery and explicit booking). Rejection uses `GroupScopedRuleType.INSIDE_BLOCK`.
- **Every block create/update runs orphan detection** — unlike windows (where only the first-window create narrows, since windows OR together), each block independently subtracts time, so any block create or update can orphan a confirmed future booking. Delete never computes orphans (removing a block only widens). Covered including recurring blocks and partial-overlap bookings.
- **Zero-change early-out**: a second `Exists()` (`has_group_scoped_block`) folds into the same per-slot membership query already used for windows; the block-span fetch is skipped when no calendar has a block. Query counts unchanged (`find_bookable_slots` 6, `check_group_availability` 5) with byte-identical output on the unconfigured path — asserted.
- **Membership-removal cleanup**: `_reconcile_slot` now deletes and audits group-scoped blocks (as well as windows) when a calendar is removed from a slot, via a shared `_delete_group_scoped_rows_for_removed_calendars` helper.
- **Occurrence-exception safety** reuses the Phase 1b `_base_manager` guard in `RecurringMixin`; block occurrence expansion goes through a `for_group_slot`-based annotate-first path.
- `external_id` for group-scoped blocks is `group-scoped-block-{uuid4}` to satisfy `BlockedTime`'s `unique_together(calendar, external_id)`.

## Feature flag

None — self-gating (early-out); no migration. REST/public API surfaces are Phase 2b; metering blocked time is Phase 2c.

## Test plan

```bash
docker compose run --rm api uv run pytest calendar_integration/tests/services/test_calendar_group_service_blocked_time.py calendar_integration/tests/services/test_group_scoped_blocked_time_discovery.py -vs
```

Covers: write lifecycle / audit / permissions / non-disclosure; block hides the calendar in that group only; block beats an overlapping window (discovery + explicit booking); recurring-block and partial-overlap orphan detection; membership-removal cascade (block deleted + audited); unconfigured-path query count + output identity.

### DIFF
--- a/calendar_integration/services/slot_engine.py
+++ b/calendar_integration/services/slot_engine.py
@@ -14,6 +14,14 @@ single-calendar / bundle walker both depend on:
   apply at each candidate window.
 - :func:`apply_policy_filter` — drop candidate proposals that violate a resolved
   :class:`EffectivePolicy` (lead-time, max-horizon, buffer envelope).
+- :func:`fetch_group_scoped_available_spans` / :func:`expand_group_scoped_available_times`
+  — batched group-scoped ``AvailableTime`` windows (``CALENDAR_GROUP_SCOPED_AVAILABILITY``
+  Phase 1b).
+- :func:`fetch_group_scoped_blocking_spans` / :func:`expand_group_scoped_blocked_times`
+  — the block analog (``CALENDAR_GROUP_SCOPED_AVAILABILITY`` Phase 2a): applied
+  in :func:`calendar_free_for_window` AFTER base availability and BEFORE the
+  window intersection, since a group-scoped block wins regardless of what any
+  window says.
 
 Everything here is **stateless and org-scoped through the passed organization
 id**.  The functions are factored out of ``CalendarGroupService`` verbatim so the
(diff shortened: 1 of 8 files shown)


calendar_integration/services/slot_engine.py
```diff
@@ -438,3 +466,136 @@ def fetch_group_scoped_available_spans(
...
+        is_recurring_exception=False,
+    )
+    for bt in non_recurring_times:
+        yield bt.group_slot_fk_id, bt.calendar_fk_id, bt  # type: ignore[misc]
+
+    recurring_times = base_qs.filter(recurrence_rule__isnull=False).filter(
+        Q(recurrence_rule__until__isnull=True) | Q(recurrence_rule__until__gte=start_date),
+        start_time__lte=end_date,
+    )
+    for master_time in recurring_times:
```
hugobessa, line 537: shouldn't we use `recurring_times.iterator()` instaed?

PR #233, opened 2026-08-06
fix(organizations): thread branding root's slug into the invitation accept URL

arthurzeras:
## Summary
- Invitation emails and the `createInvitation` GraphQL mutation branded the email's *content* (app name, logo, colors) but the accept-invite *link itself* always pointed at the same slug-less `/auth/accept-invite/?token=...` URL — so a branded organization's accept page had no way to know which organization's branding to render before the invitee authenticates.
- Adds `organizations/invitation_urls.py` to key the accept link on the resolved **branding root's** slug (so a reseller's child organization's invite carries the reseller's slug, not its own unset one), via a new `account_accept_invitation_branded` settings template. Falls back to the original URL, byte-for-byte, whenever no branding root resolves.
- Amends the Organization Auth-Area Branding plan's Phase 5 (`ai-plans/2026-08-04-ORGANIZATION_AUTH_BRANDING_IMPLEMENTATION_PLAN.md`) with the gap found, the fix, and updated touch list.
- Client handoff doc drafted separately at `.vinta-ai-workflows/client-handoffs/2026-08-06-branded-invitation-accept-url.md` (gitignored — share with the SPA team directly) — flags that the SPA needs a new route for the `/o/{slug}/auth/accept-invite/` shape before/alongside this deploying.

## Test plan
- [x] New unit tests: `organizations/tests/test_invitation_urls.py`
- [x] New integration tests: `organizations/tests/test_services.py`, `public_api/tests/test_provisioning_flow.py` (covers the reseller-child-inherits-slug case)
- [x] Full `organizations/` + `public_api/` suite passes on this branch (1493 passed)
- [x] Ruff and mypy clean on touched files
- [x] Manually verified real GraphQL responses (branded and unbranded) via a local test run

### DIFF
--- a/vinta_schedule_api/settings/test.py
+++ b/vinta_schedule_api/settings/test.py
@@ -7,6 +7,11 @@ DEBUG = True
 
 # Add the invitation-accept URLs so sendEmail=false tests receive a non-null inviteUrl.
 # The base settings omit these keys; staging/production set them against FRONTEND_BASE_URL.
+# Unlike staging/production, these values are plain strings, not f-strings -- the host
+# here is a hardcoded literal, so {token}/{org_slug} don't need the {{ }} escaping those
+# other settings use to survive f-string interpolation of FRONTEND_BASE_URL. Single braces
+# is correct here: build_invitation_accept_url calls .format(token=..., org_slug=...) on
+# these templates directly.
 HEADLESS_FRONTEND_URLS = {
     **HEADLESS_FRONTEND_URLS,
     "account_accept_invitation": "http://localhost:3000/auth/accept-invite/?token={token}",
--- a/vinta_schedule_api/settings/test.py
+++ b/vinta_schedule_api/settings/test.py
@@ -5,11 +5,14 @@ from .base import *
 # and N+1 regressions fail the suite instead of only surfacing on the dev runtime.
 DEBUG = True
 
-# Add the invitation-accept URL so sendEmail=false tests receive a non-null inviteUrl.
-# The base settings omit this key; staging/production set it against FRONTEND_BASE_URL.
+# Add the invitation-accept URLs so sendEmail=false tests receive a non-null inviteUrl.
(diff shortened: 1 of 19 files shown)


vinta_schedule_api/settings/test.py
```diff
@@ -5,11 +5,14 @@
...
 
-# Add the invitation-accept URL so sendEmail=false tests receive a non-null inviteUrl.
-# The base settings omit this key; staging/production set it against FRONTEND_BASE_URL.
+# Add the invitation-accept URLs so sendEmail=false tests receive a non-null inviteUrl.
+# The base settings omit these keys; staging/production set them against FRONTEND_BASE_URL.
 HEADLESS_FRONTEND_URLS = {
     **HEADLESS_FRONTEND_URLS,
     "account_accept_invitation": "http://localhost:3000/auth/accept-invite/?token={token}",
+    "account_accept_invitation_branded": (
+        "http://localhost:3000/o/{org_slug}/auth/accept-invite/?token={token}"
```
hugobessa, line 19: The other settings are using {{ }} instead of { }, please check which one is correct

hugobessa (changes requested):

vinta_schedule_api/settings/test.py
arthurzeras: Both are correct. Staging/production build the templates as f-strings to interpolate `FRONTEND_BASE_URL`, so they must escape the placeholders; `test.py` uses plain strings (hardcoded host, no interpolation needed), so single braces are already the literal placeholders `build_invitation_accept_url` formats. Rather than change working code, I added a comment in [test.py](https://github.com/vintasoftware/vinta-schedule-api/pull/233/changes/af08016e32fb38c620eebdb25088c09db5c445c2#diff-843a9bd3b5e17347ad27fd1a91fa71f985fca3f068e67bbbc925096fa3648a5eR10-R14) explaining the distinction...

